PR #130, opened 2022-07-31
Debugging/channel settings

glitch003:
You probably don't want to merge this directly because I added some extra debug logging and had to change the bundle ID to get provisioning to work for a physical device, but I figured seeing this in a PR would make it easier to tell what I actually changed.


Meshtastic/Helpers/BLEManager.swift
```diff
@@ -1169,6 +1168,55 @@ class BLEManager: NSObject, ObservableObject, CBCentralManagerDelegate, CBPeriph
 		return 0
 	}
 	
+	public func getChannel(channelIndex: UInt32, wantResponse: Bool) -> Bool {
```
glitch003, line 1283: This is the main function I added

Meshtastic/Helpers/MeshPackets.swift
```diff
@@ -1133,6 +1133,9 @@ func adminAppPacket (packet: MeshPacket, meshLogging: Bool, context: NSManagedOb
 		print(try! powerConfig.jsonUTF8Data())
 		print(powerConfig.meshSdsTimeoutSecs)
 		
+	} else if let channel = try? Channel(serializedData: packet.decoded.payload) {
```
glitch003, line 1358: Attempt at parsing the incoming channel.  Doesn't seem to work.

Meshtastic/Views/Settings/ShareChannel.swift
```diff
@@ -86,6 +86,13 @@ struct ShareChannel: View {
 			.onAppear {
 
 				self.bleManager.context = context
+				let i: UInt32 = 1;
```
glitch003, line 90: This is where we actually get the channel, when the user views the "share channel" screen

PR #278, opened 2022-12-30
2.0.10 working changes

garthvh:
* Real Acks for DM's
* Fix for inconsistent delete all messages functionality
* Channel editor
* Don't show distance if the current location is Apple Park
* Save canned messages messages string to local database
* Additional translation strings
* Extend mesh map all the way to the top of the device screen
* Update Protobufs
* Use ack errors for admin message log
* Localize formatting of date strings
* Localize the mesh log
* Respect use 24hour date setting
* Filter out default location from node list distance measurement


garthvh:
@formtapez There are a lot of new strings available for translation, I added the English version for all of them to the DE file.

PR #281, opened 2023-01-03
german translation

formtapez:
i left some strings untouched, because im missing the context for them.

it would be nice if the strings with printf-formating could have an alternative order of parameters, so we could make more sense of it. (translation just sounds wrong, if you cant switch the order of some parameters)


caveman99 (approved):
Thanks, i can fill in a few blanks after the merge ...

garthvh:
Not sure if there was a way to move the variable order around, there are not that many with more than one argument. When there is one variable you can put it anywhere in the formatted string

PR #284, opened 2023-01-06
Add Chinese translation

BG6TNB:
Only Simplified Chinese was added to Localization, the text in the rest of the UI was not handled.


garthvh:
Thanks, I need to get the current release merged and then will create a branch for this and add all the newly created strings for you to take a look at

garthvh:
I merged a whole bunch of new strings

PR #287, opened 2023-01-12
finished the Chinese translation for Localizable.strings

BG6TNB:
I have finished the Chinese translation of the main UI, but I found that there is still some text in the settings that has not been added to the Localizable.strings.


garthvh:
Yeah there are probably a few strings remaining

PR #295, opened 2023-01-21
Fix protoc

sachaw:
A few changes/considerations

1. Generated files are now in `Meshtastic/Protobufs/meshtastic`
    - The old generated swift files will need to be removed from the old locationwhen new ones are generated.
    - The generation step can be offloaded to CI like we do with some of the other repo's


gen_protos.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # simple sanity checking for repo
-if [ ! -d "../Meshtastic-protobufs" ]; then
+if [ ! -d "../protobufs" ]; then
   echo "Please check out the https://github.com/meshtastic/Meshtastic-protobufs the parent directory."
   exit
```
caveman99, line 6: The path is still pointing to the old URL, even if it is just a hint...

sachaw: That should be correct, as that's the folder that the git submodule clones to, but please let me know if I'm missing something

caveman99: I was targetting line 5 ... the echo :-)

sachaw: Lol, my bad, mobile app doesn't show the exact line 🤦

caveman99: Sorry to be so nitpicky, but a new developer onboarding will stumble across this. When i started with the firmware, i was searching 2 days for the nRF includes till i found out i had to check out submodules...

sachaw: Agreed, thanks

sachaw:
@garthvh not sure where you update the include path for the change in folder, but should be all good apart from that.

garthvh:
Merged with https://github.com/meshtastic/Meshtastic-Apple/pull/306

PR #310, opened 2023-02-06
Node map annotations

garthvh:


garthvh:
Add speed and heading to node map annotations

PR #386, opened 2023-08-27
Polish translation

kokroo:
This is the localization file for the Polish language.


garthvh:
Thanks!  merged.

I did add 5 new strings recently that I added to the polish file in the working copy branch, I will push out a TestFlight version.

PR #403, opened 2023-09-30
Add Chinese translation

BG6TNB:


BG6TNB:
Add Chinese translation

PR #464, opened 2024-01-19
Allow users to enter an existing key when adding a channel

edhaber:
Allow the key field to be edited by a user and validate that the user supplied key is valid base64 encoded and the correct key length.


garthvh:
This is not working properly on load, and needs a bigger control for keys.

I have reverted this for now.

PR #475, opened 2024-02-01
fix: offline map type option getting out of sync

Austinpayne:
Fixes a bug where the offline map picker option was getting inverted until another action was taken, e.g. close options, change map type, etc.

Before:

https://github.com/meshtastic/Meshtastic-Apple/assets/7528924/1171a6e6-d696-420d-b382-a48ace0fe9d0

After:

https://github.com/meshtastic/Meshtastic-Apple/assets/7528924/d0f7fff8-c8da-4396-9fca-d02ec97ec5af


garthvh:
Merged, thanks!

PR #484, opened 2024-02-12
fix: slow typing speed when lots of messages

Austinpayne:
Refactors both the channel and user message views to isolate typing state which prevents excessive re-rendering of large message lists on every new character typed. Also consolidates typing view code of both lists into the new `TextMessageField` and related sub views.

Note: This is probably not much of an issue on newer phones. It's most pronounced on my old iPhone 8 (starts feeling slow around 50 messages) but on an iPhone 13 mini I did not notice it until I had hundreds of messages. Still, this feels like a very positive improvement, typing is very snappy after isolating the state to a separate view.

Before:

https://github.com/meshtastic/Meshtastic-Apple/assets/7528924/a3863f38-4ced-4630-9130-6dc4f4a58f20

After:

https://github.com/meshtastic/Meshtastic-Apple/assets/7528924/0fcc26de-00ea-4658-addb-bd1f5e61b67b


garthvh (approved):
Awesome, looks great

garthvh:
This will also make it easier to enable the anchor on scroll position if available, which also seems to perform better.

https://developer.apple.com/documentation/swiftui/view/scrollposition(id:anchor:)

Issue #456, opened 2024-01-07
Resend failed message button


dmthepm:
### OS

iOS

### Description

When a message fails to send due to max transmission reached, I would love to be able to long press and hit resend. Maybe you can leave the record of the failed send there but just populate and send a new message. 

### Participation

- [ ] I am willing to submit a pull request for this issue.

garthvh:
Both phone apps currently allow you to copy the message and resend it yourself.

This has been on my list for a while, but it is hard to know how to handle it in a mesh friendly way, acks can take quite a while.
![longack](https://github.com/meshtastic/Meshtastic-Apple/assets/1795163/f651d042-8cf5-4e61-9edf-96ae2c1ec3e3)

drewzh:
I would find this incredibly handy. I'm sometimes on the move and need to send out 'ping' messages to test propagation. Having to retype, even 'ping' each time I need to send a test, is quite tedious. One other potential fix for this (in my use case at least), would be to provide an ability to send a canned message, which requires only 2 clicks from the user.

garthvh:
With messages taking up to 10 minutes the ability to copy and paste a message will remain the solution.

drewzh:
> With messages taking up to 10 minutes the ability to copy and paste a message will remain the solution.

@garthvh I understand the desire to not allow re-transmission of a message that may already be in play (and why would you, the system does that for you), but when the max-transmission limit has been reached, would it not be nice to give the user a quick retry option, perhaps with a small link that's easy to click - or still really useful, an option on the context menu to re-send?

For me, copying and pasting whilst I'm on the go is sometimes very cumbersome.

thebentern:
@drewzh this has been already with implemented with #486

garthvh:
> > With messages taking up to 10 minutes the ability to copy and paste a message will remain the solution.
> 
> @garthvh I understand the desire to not allow re-transmission of a message that may already be in play (and why would you, the system does that for you), but when the max-transmission limit has been reached, would it not be nice to give the user a quick retry option, perhaps with a small link that's easy to click - or still really useful, an option on the context menu to re-send?
> 
> For me, copying and pasting whilst I'm on the go is sometimes very cumbersome.

https://support.apple.com/en-ca/guide/iphone/iph6d01d862/ios

PR #486, opened 2024-02-13
feature: add simple retry mechanism

Austinpayne:
Implements a simple retry mechanism similar to iMessages.

IMG_E2F6ECF29094-1

IMG_0B3B60297364-1

Closes https://github.com/meshtastic/Meshtastic-Apple/issues/456


garthvh (commented):
Would probably be good to just make it available on the max transmission error, the rest are not really likely to be corrected on re-send.  Only maxRetransmit or timeout are likely to be fixed by trying again.

garthvh (approved):

Austinpayne:
> Would probably be good to just make it available on the max transmission error, the rest are not really likely to be corrected on re-send. Only maxRetransmit or timeout are likely to be fixed by trying again.

Good point, thanks for updating!

drewzh:
Amazing, I just saw this. One improvement suggestion however if it's possible... can you update the status of the message back to 'Waiting for acknowledgment"? When you click retry, there's no UI update and you can continue to hit the (i) button and click Retry... it's just a bit janky from a UX perspective. Great that we have this feature now as I'll be making use of this regularly, so thank you very much for the efforts spent!

Austinpayne:
That's how it was working when I originally submitted the MR. Will investigate, seems like a bug that slipped in.

PR #511, opened 2024-02-21
feature: add power configuration

Austinpayne:
Adds ability to configure power settings in the app. Also refactors two shared config views to greatly reduce lines of code: `ConfigHeader` and `SaveConfigButton`. I recommend reviewing without whitespace changes visible.

IMG_FD41C01F4170-1
IMG_FD41C01F4170-2


garthvh:
People really mis use a lot of these settings.

garthvh:
Looks good, I will merge it as it will probably get dangerous to leave it open too long.

The power settings need a bunch of validation and conditional logic, and some of those settings probably just don't need to be in the app.

Austinpayne:
> People really mis use a lot of these settings.

They did seem quite advanced but I found the ADC override helpful to tune battery percentage on my Heltecs. What are your thoughts on a toggle for advanced settings in the app? I've been thinking of working on that next. If so I can hide these away in there and add some "user beware" type messages.

garthvh:
I have never seen a good mobile app with advanced settings

garthvh:
Not a fan at all of that approach, we need to reduce confusing settings, not hide them.

PR #552, opened 2024-03-11
Device metrics chart improvements

wnagele:
1. Filter out the [magic charging battery level (101%)](https://github.com/meshtastic/firmware/blob/master/src/modules/Telemetry/DeviceTelemetry.cpp#L13) as it just confuses the plot
2. Change the interpolation to `linear` to get clearer readability especially if fast changes occur
3. Reduce the size of the scatter plot points so they are better visible on a busy plot


garthvh:
Do you have some screenshots? There have been several changes from .cardinal and they have all looked worse

wnagele:
Here you go:
Screenshot 2024-03-11 at 22 36 32

garthvh:
And why remove the fully charged data?

wnagele:
Yeah you do have a point there. Basically we would need a dedicated `charging` state instead of using a magic (`101%`) value to reflect that - only then can we properly show both SoC and the charging indicator. Maybe I'll find some time to put that into the firmware going forward - then we can revisit. For now I've removed this part.

garthvh:
I had the magic number adjusted in the protobufs and firmware from 110 to 101 for the chart, never seen any issues with it.

PR #571, opened 2024-03-26
Fixing UserDefault wrapper crashing with enums

ChDel:
I tested this and it's not crashing. 
The enums are stored as rawValues 
to restore them the enum needs to be Decodable and the compiler will complain is they are not so it's compiler safe
I'm then turning the stored value into Json for decoding...

### DIFF
--- a/Meshtastic/Extensions/UserDefaults.swift
+++ b/Meshtastic/Extensions/UserDefaults.swift
@@ -13,6 +13,36 @@
 
 import Foundation
 
+@propertyWrapper
+struct UserDefault<T: Decodable> {
+	let key: UserDefaults.Keys
+	let defaultValue: T
+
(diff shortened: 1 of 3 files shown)


Meshtastic/Extensions/UserDefaults.swift
```diff
@@ -13,6 +13,36 @@
...
+		self.defaultValue = defaultValue
+	}
+
+	var wrappedValue: T {
+		get {
+			if defaultValue as? any RawRepresentable != nil {
+				let storedValue = UserDefaults.standard.object(forKey: key.rawValue)
+
+				guard let storedValue,
+				let data = ("\"\(storedValue)\"".data(using: .utf8)),
```
ChDel: I need to change this a bit to work with numbers too

garthvh (approved):
Looks good, I will push it into TestFlight

PR #576, opened 2024-03-29
Add mqtt status icon in channels with uplink/downlink enabled

daviesgeek:
First and foremost, please let me know if there are any styling issues with my code, I'm very rusty on iOS code generally and a relative newbie when it comes to SwiftUI.

This PR adds an icon in the channels detail view when uplink or downlink is enabled and also adds the status of the connection in the color of the icon as well.
This also adds the same connection icon in the MQT config settings screen

Here are some screenshots:  

Frame 1


garthvh (approved):
Looks good, I like it.

PR #583, opened 2024-04-03
Better mqtt status icons, including up/down icons

daviesgeek:
Apologies, meant to get this out sooner than i did, but ran out of time.

Updated the MQTT status icon to display up/down information as well as connected/not connected
Also now includes a popover to display the currently connected topic

image
image


garthvh (approved):
Looks really good!

PR #584, opened 2024-04-04
2.3.4 working changes

garthvh:
* Improved MQTT Status Icons
* String Length validation bug fix
* Show admin channel in the channels list
* Updated share channels tip
* Show connection errors for mqtt client proxy
* Swedish translation
* Stop resetting app settings now that fixed position has been updated.
* Add timezone to device config, create it automatically from your timezone.


daviesgeek:
There's currently a bug in my MQTT icon code where it shows connected all the time even when a channel doesn't have MQTT enabled. Will be submitting a PR shortly

PR #598, opened 2024-04-11
Made New Node notifications off by default

RCGV1:
The new node notification behavior should be optional as users will most likely start complaining about a large amount of notifications in large meshes. These notifications are intended more for power users.


garthvh:
I think notifications should all be on by default, they can disabled now, and some more types will be added to the settings

PR #627, opened 2024-04-30
completed translations for zh-Hans (except those for logs)

shengminchen:
Translations are added to complete the zh-Hans language pack.


garthvh (approved):
Looks good

PR #663, opened 2024-06-04
Refactor `NodeMapSwiftUI.swift`

72A12F4E:
## What

This change breaks down `NodeMapSwiftUI.swift` into computed properties to improve compile performance, and adds warnings to the project when a type-check exceeds 200ms.

## How

These stats were from a 2019 Macbook Pro. I expect it'll be much faster on some more recent ARM hardware.

## Why

The goal here isn't entirely to optimize for compile performance. Instead, I'm using this statistic as a rough indicator as to which View hierarchies are overly complex and could use some TLC

## How did you test this?

Compiled & ran the app on a simulator, but haven't been able to build to a device due to entitlements issues.


72A12F4E:
> Can you take a look at the conflicts?

Done ✅

PR #684, opened 2024-06-10
Fixed translation for better user understanding

Oliver0804:
Fixed translation for better user understanding


Oliver0804:
I have made localization modifications to the USER CONFIG page so that the currently displayed strings support localization features. If you have any suggestions regarding the code style or modifications, I can follow them to make changes. Thank you.

截圖 2024-06-10 晚上10 42 29

PR #700, opened 2024-06-17
Text formatting fix for Hops Away

njh:
The 'Hops Away' contextual menu item for messages was suffixed with a close bracket and 'dB'.
But hops are not measured in dB.

I believe this is a copy/paste bug.


garthvh:
Good catch

PR #701, opened 2024-06-18
Fix coredata persistence issues with nodeinfo

72A12F4E:


Meshtastic/Extensions/CoreData/MyInfoEntityExtension.swift
```diff
@@ -5,9 +5,35 @@
...
+			if (channel as? ChannelEntity)?.name?.lowercased() == "admin" {
+				return Int32(index)
+			}
+		}
+		return 0
+	}
+	
+	var bleName: String? {
+		myInfoNode?.user?.longName
+	}
```
72A12F4E, line 36: I was able to move these to computed properties, so we don't have to rely on saving them into this particular entity. Should simplify the data model a bit

PR #707, opened 2024-06-19
Fix MyInfo persistence that caused bluetooth desyncs

72A12F4E:
This change tries to address race conditions that occur when repeatedly pairing and unpairing a node via bluetooth. This was caused by the fetch-update code when persisting a `MyInfoEntity` packet instead of relying on CoreData to do the synchronization and merging.

I also factored out a method that updates the appropriate fields on `connectedPeripheral` whenever packets come in, so the node list & bluetooth screen will render correctly.

I also fixed an inverted condition in the message validation logic. We want to make sure messages are _NOT_ empty before we allow them to be sent.

### DIFF
--- a/Meshtastic/Helpers/BLEManager.swift
+++ b/Meshtastic/Helpers/BLEManager.swift
@@ -503,8 +503,11 @@ class BLEManager: NSObject, CBPeripheralDelegate, MqttClientProxyManagerDelegate
 	}
 
 	func peripheral(_ peripheral: CBPeripheral, didUpdateNotificationStateFor characteristic: CBCharacteristic, error: Error?) {
-		
-		Logger.services.error("💥 BLE didUpdateNotificationStateFor error: \(error?.localizedDescription ?? "Unknown")")
+		if let error {
+			Logger.services.error("💥 BLE didUpdateNotificationStateFor error: \(characteristic) \(error.localizedDescription)")
+		} else {
+			Logger.services.info("ℹ️ peripheral didUpdateNotificationStateFor \(characteristic)")
+		}
 	}
 
 	// MARK: Data Read / Update Characteristic Event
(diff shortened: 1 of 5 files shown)


Meshtastic/Helpers/MeshPackets.swift
```diff
@@ -92,52 +92,24 @@ func moduleConfig (config: ModuleConfig, context: NSManagedObjectContext, nodeNu
...
 	let logString = String.localizedStringWithFormat("mesh.log.myinfo %@".localized, String(myInfo.myNodeNum))
 	MeshLogger.log("ℹ️ \(logString)")
 
-	let fetchMyInfoRequest: NSFetchRequest<NSFetchRequestResult> = NSFetchRequest.init(entityName: "MyInfoEntity")
-	fetchMyInfoRequest.predicate = NSPredicate(format: "myNodeNum == %lld", Int64(myInfo.myNodeNum))
+	let myInfoEntity = MyInfoEntity(
+		context: context,
+		myInfo: myInfo,
+		peripheralId: peripheralId
+	)
```
72A12F4E, line 103: `MyInfoEntity` is uniqued on `myNodeNum`, so we won't get duplicate data here.

Meshtastic/Helpers/BLEManager.swift
```diff
@@ -941,7 +944,7 @@ class BLEManager: NSObject, CBPeripheralDelegate, MqttClientProxyManagerDelegate
 		}
 		
 		// Don't send an empty message
-		guard message.isEmpty else {
+		guard !message.isEmpty else {
```
72A12F4E, line 959: this was preventing any messages from being sent 🤦

garthvh (approved):
Awesome

PR #710, opened 2024-06-20
Bump Protos

72A12F4E:


garthvh:
Seems like this is going the wrong way

garthvh:
I think I updated these earlier and it looks like this removes the new ble logging setting

PR #726, opened 2024-06-27
Update Localizable.strings Redone into Portuguese

PhilRosa-Leeke:
Garth, if you like, instead of redoing the whole file, just send me the lines to translate and their locations (line numbers) and I will do it, no worries.


72A12F4E:
Hey @PhilRosa-Leeke! Thanks for doing this!

We recently merged a PR to adopt String Catalogs in Xcode 15, which should make translations a lot easier to manage across every locale. Unfortunately, these changes aren't compatible with the legacy localization files.

If you're up for it, can you create a new branch off of `2.3.12_Working_Changes` and follow the steps [here](https://developer.apple.com/documentation/xcode/localizing-and-varying-text-with-a-string-catalog#Add-your-translations-using-the-string-catalog-editor) to add the translations? 

Sorry for the moving target, but hopefully it should be an easier process in the long term.

PR #729, opened 2024-06-28
Extract the generated protobufs into its own Swift package

72A12F4E:
This change modifies the process for generating and integrating the Meshtastic protobufs into the client application.

* The generated Swift code is now in a local SPM package `MeshtasticProtobufs`
* An Xcode Workspace file `Meshtastic.xcworkspace` was created to more easily manage the new build targets.
* The code generation script for the protos was modified to generate the Swift code into the new location.
* The README.md was updated to reflect these changes.

NOTE: After merging this PR, do not open the project file `Meshtastic.xcodeproj`. You must use the workspace `Meshtastic.xcworkspace`

Extracting out the generated protobuf code into its own library enables several opportunities for the project. This is just a first step, but with some more modularization, a standalone Apple Watch app or other targets starts to become a little bit more achievable to implement.

After extracting the protobufs into a Swift package, I validate these changes by building and running the Meshtastic app to an iPhone 15 Pro Max, and tried changing some settings on a local node. I then messaged back and forth using two local nodes connected to two different iOS devices.


garthvh (approved):
Awesome, I really missed the workspace after reverting back to before it was implemented

PR #734, opened 2024-06-28
Migrate to String Catalogs

72A12F4E:
This change migrates the Localizable.strings to a [string catalog](https://developer.apple.com/documentation/xcode/localizing-and-varying-text-with-a-string-catalog). 

## What is it?

String Catalogs are a new feature in Xcode 15 that provide a centralized source of truth for managing localized string resources, and provide more insight into the completeness of the various translations.

There is more info in this WWDC session: https://developer.apple.com/videos/play/wwdc2023/10155

## How did you migrate?

I selected the existing Localized strings in the file navigator, right clicked, and selected "Migrate to string catalog". However, I noticed that some of the Portuguese translations were not migrated correctly, so I had to revert these changes, rename some files, and performed the migration steps again.

## How does it look now?

Screenshot 2024-06-28 at 17 45 11

## What's next?

With the new insights into the percentage of the app that is localized, contributors can now more easily provide translations for locales, and fill in any gaps we currently have. All supported locales show as ~30% translated, so there is ample opportunity to contribute here.


thebentern (approved):

thebentern:
Nice! This interface looks much nicer

garthvh (approved):
Looks great, there is an open pull for Portuguese stuff I didn't totally understand

thebentern:
> Looks great, there is an open pull for Portuguese stuff I didn't totally understand

I don't understand Portuguese either. It's all Greek to me ;-)

PR #735, opened 2024-06-29
Migrate away from using the .localized string extension which could c…

72A12F4E:
This change removes the extension on `String` that aliased `NSLocalizedString`. 

`NSLocalizedString` has some special magic under the hood that is mean to work specifically with static strings in the source code. By wrapping it in a computed property, a developer could inadvertently think dynamic string content was being localized. 

While this was fine in the majority of cases. we did find a few where it didn't work. Notably, this was occurring in some of the `localized` extensions on other types, because these types' `rawValue` is set at runtime.


72A12F4E:
Conflicted on this one. Its one of those things that was fine 99% of the time, but wasn't in a few cases

PR #749, opened 2024-07-02
New BLE LogRecord characteristic

thebentern:
Not my finest code. Just formatting things like the legacy characteristic and sending it through the same extracted method


72A12F4E (approved):

garthvh (approved):
Looks good

PR #777, opened 2024-07-10
Add First Heard and Last Heard times to node details

72A12F4E:
This change adds the last heard date to the node details screen. Unfortunately, it seems much of this data is buggy for me, and I have several nodes with lastHeard dates on Jan 1 1970, so I can't really validate that the change works on the happy path.

IMG_D7578A3F43FE-1

### DIFF
--- a/Localizable.xcstrings
+++ b/Localizable.xcstrings
@@ -10798,6 +10798,9 @@
         }
       }
     },
+    "Last heard" : {
+
+    },
     "Latitude" : {
 
     },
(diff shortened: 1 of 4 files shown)


garthvh:
Can we add the first heard as well? I added it to node info recently and it has been populating for a while

garthvh:
If you share your location from your phone it should get time

72A12F4E:
IMG_C944A89BAC8E-1

@garthvh hows this?

garthvh (approved):
Looks great!

PR #781, opened 2024-07-10
Delete dead code from node detail

72A12F4E:
This was leftover from the node actions refactor and can be deleted


garthvh (approved):
Kill it

PR #791, opened 2024-07-14
Restore the Exchange Positions and TraceRoute button back to the node list context menu

72A12F4E:
This change restores the Exchange Positions & Trace Route buttons back in the node list's context menu.

Unfortunately, the context menu doesn't render the alerts after a position packet is sent because the view is already removed from the render. It would take some refactoring to restore that to the node list, but the position packet and traceroute is still sent, and this should be functional enough for now.

Longer term, it would be cool to use a non-blocking UI action sheet-type of toast message for a lot of these confirmation dialog things.


garthvh (approved):

garthvh:
I am trying to slow people down a little for trace routes, position exchanges and admin message as they are kind of expensive messages that fail more when they get sent repeatedly

PR #794, opened 2024-07-16
prototype node map notification

garthvh:


Meshtastic/Persistence/UpdateCoreData.swift
```diff
@@ -357,6 +357,21 @@ func upsertPositionPacket (packet: MeshPacket, context: NSManagedObjectContext)
 					do {
 						try context.save()
 						Logger.data.info("💾 [Position] Saved from Position App Packet For: \(fetchedNode[0].num.toHex(), privacy: .public)")
+						// Notification for DM Positions (position exchange)
+						if packet.from != 4294967295 {
```
72A12F4E, line 361: Just curious what this magic number means?

garthvh: UInt32 max or FFFF,FFFF hex is used to identify messages that are being sent to a Channel and are not a DM to an individual user. In this case I want a notification when exchange / request position is used so I want to exclude regular position updates over the primary channel.

PR #798, opened 2024-07-16
Extract some magic numbers to named constants with some documentation

72A12F4E:


garthvh (approved):
Found the minimum value, renamed emptyNodeNum to maximumNodeNum since it makes more sense.

Issue #799, opened 2024-07-17
🚀 [Feature Request]: Filter out MQTT Nodes you can't send DM's to


garthvh:
### OS

iOS, iPadOS, macOS

### Description

Don't show contacts that are multiple hops away on the default Channel via MQTT as you can't really message them anymore

### Participation

- [ ] I am willing to submit a pull request for this issue.

todd2982:
How do you see this impacting users connected to a private MQTT server that does not get forced to the zero hop count?

violarulan:
I think it's only should be applied to nodes that are discovered via public MQTT. With a private broker, you still can DM them hops away.

hayschan:
This is affecting private MQTT server as well, which should not be the intended target. 

It should only target the default public MQTT server only.

PR #814, opened 2024-07-23
Don't Show unreachable nodes in the contact list

garthvh:
* [ ] Filter out mqtt nodes that are not direct


garthvh:
Closes #799

PR #824, opened 2024-07-31
Remove Fetched Properties and Replace with FetchRequests

garthvh:
Conversion from Fetched Properties to just FetchRequests, seems to be performing better than my similar attempt earlier, @72A12F4E this probably makes your refactor easier as well.

None of this seems to be iOS 18 dependent, but does fix what is currently broken in iOS 18 as well


72A12F4E:
Can you expand on what changed in the data model? Hard to see from the XML diffs

72A12F4E:
On the surface this LGTM ✅ 

@garthvh you'll probably want to cherry-pick out the only these two commits, and create a fresh PR on main. 

410acf8fdbfc7b584483b1d8437e2f34d551d54c and ae01eecd9921c941bf193fc09b70761cd049d9dc  

This PR currently includes Xcode 16 specific stuff which you won't be able to submit to App Store

72A12F4E:
I've opened a PR here with a cleaner commit history: https://github.com/meshtastic/Meshtastic-Apple/pull/825

PR #825, opened 2024-07-31
Remove Fetched Properties from channel and message models

72A12F4E:
cleaned up from https://github.com/meshtastic/Meshtastic-Apple/pull/824 which included merge commits & app icon updates

@ChrDel-2 and @garthvh are the original authors of the changes

### DIFF
--- a/Meshtastic.xcodeproj/project.pbxproj
+++ b/Meshtastic.xcodeproj/project.pbxproj
@@ -1634,7 +1634,7 @@
 				INFOPLIST_FILE = Meshtastic/Info.plist;
 				INFOPLIST_KEY_CFBundleDisplayName = Meshtastic;
 				INFOPLIST_KEY_LSApplicationCategoryType = "public.app-category.utilities";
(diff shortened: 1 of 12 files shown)


garthvh (approved):
Looks good I will deploy a test flight build

garthvh:
@72A12F4E The Fetched properties in core data were essentially the same as the predicates being used now.

ChrDel (approved):
Looks good to me!
Thank you for opening a PR for this!
🙏

72A12F4E:
The message list view is kinda borked still tho

Discussion in discord for a fix

72A12F4E:
We need to do a data purge or a migration to combine the timestamps as discussed in Discord, then this should work as expected. It might be necessary to explicitly add sorting as well

garthvh:
> We need to do a data purge or a migration to combine the timestamps as discussed in Discord, then this should work as expected. It might be necessary to explicitly add sorting as well

I just removed it, messageTimestamp was always loaded in parallel

PR #833, opened 2024-08-02
Exchange Position Response Notifications

garthvh:
Notification that a new position sent as a DM as been received

* [x] Only show the notification for DM's
* [x] Only Show the notification for DM's sent to the connected node
* [ ] Open the notification in the node map view using the NavigationState passed in from the notification

### DIFF
--- a/Meshtastic/Router/Router.swift
+++ b/Meshtastic/Router/Router.swift
@@ -79,14 +79,14 @@ class Router: ObservableObject {
 	}
 
 	func routeNodes(_ components: URLComponents) {
-		   let nodeId = components.queryItems?
-			   .first(where: { $0.name == "nodenum" })?
-			   .value
-			   .flatMap(Int64.init)
-		   let detail = components.queryItems?
-			   .first(where: { $0.name == "detail" })?
-			   .value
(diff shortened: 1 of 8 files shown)


Meshtastic/Router/Router.swift
```diff
@@ -78,12 +78,15 @@ class Router: ObservableObject {
...
-		route(to: .nodes(selectedNodeNum: nodeId))
+	func routeNodes(_ components: URLComponents) {
+		   let nodeId = components.queryItems?
+			   .first(where: { $0.name == "nodenum" })?
+			   .value
+			   .flatMap(Int64.init)
+		   let detail = components.queryItems?
+			   .first(where: { $0.name == "detail" })?
+			   .value
+		   route(to: .nodes(NodeListNavigationState(selectedNodeNum: nodeId, details: detail != nil ? NodeDetails(rawValue: detail!) : nil)))
```
72A12F4E: No need to force unwrap, just use a `map`

```
route(
    to: .nodes(
        NodeListNavigationState(
            selectedNodeNum: nodeId, 
            details: detail?.map(NodeDetails.init(rawValue:))
        )
    )
)

PR #837, opened 2024-08-02
Add support for navigating to currently connected node info

### LINKED ISSUE #836: 🚀 [Feature Request]: Navigate to the connected radio's node info or device metrics log from the bluetooth screen
### OS

iOS

### Description

Currently, unless I'm missing something obvious, in order to get to the device details of the currently connected radio, I have to navigate to the list of nodes, find my node, select it, and then go to the device metrics log.

I would like to be able to view the device metrics log directly from the connected radio bluetooth screen.

### Participation

- [X] I am willing to submit a pull request for this issue.

### Additional comments

_No response_

gabek:
No pressure to accept this PR if it doesn't make sense or I'm overlooking something obvious. 

Closes #836

This PR adds support for navigating to the currently connected node info screen directly from the bluetooth radio screen 
by tapping on the connected radio.

It adds a small right pointing chevron after the radio is connected to show that tapping on it will navigate somewhere.

## What changed?

Updated `Connect.swift` to support navigating to the device node info screen.

## Why did it change?

To add a level of convenience to getting to your currently connected device's details.

## How is this tested?

It was tested with my local node on iOS hardware.

https://github.com/user-attachments/assets/4e67c99a-7e4c-4dbc-a31a-c529d76a36a4

## Checklist

### DIFF
--- a/Meshtastic/Views/Bluetooth/Connect.swift
+++ b/Meshtastic/Views/Bluetooth/Connect.swift
@@ -308,7 +308,8 @@ struct Connect: View {
 			)
 			.navigationDestination(isPresented: $navigateToMetricsLog) {
 				if let node {
-					DeviceMetricsLog(node: node)
+					NodeDetail(bleManager: _bleManager, connectedNode: node, node: node)
+						.navigationBarTitleDisplayMode(.inline)
 				}
 			}
 			.onAppear {
--- a/Meshtastic/Views/Bluetooth/Connect.swift
+++ b/Meshtastic/Views/Bluetooth/Connect.swift
@@ -27,6 +27,7 @@ struct Connect: View {
 	@State var invalidFirmwareVersion = false
 	@State var liveActivityStarted = false
 	@State var selectedPeripherialId = ""
+	@State var navigateToMetricsLog = false
 
 	init () {
 		let notificationCenter = UNUserNotificationCenter.current()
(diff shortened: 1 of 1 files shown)


72A12F4E:
Wondering if we should nav to node details instead? From an IA perspective, its similar to tapping a node in the node list.

Curious what others think here

gabek:
Here's a question then, how would you feel about adding current battery, channel utilization, and airtime as values to the node details screen if that node is your currently connected node (or even for all nodes, but that's less important imho)? Then, in this case that I'm trying to optimize for, there'd be no need to navigate to device metrics at all and node details would have everything important about your current device at a glance.

gabek:
Got it. I honestly just want to see my battery status without navigation diving :) Right now, device metrics seems to be the only way to get that information. So maybe separately there can be some brainstorming of how this value might be surfaced.

gabek:
It now navigates to Node Info.

https://github.com/user-attachments/assets/cf57fed3-a3da-4cab-88fe-b6348949bcd7

gabek:
> Maybe it can be as simple as adding the battery level of the connected device there on the BLE screen?

Love that. I'll file something separately for changing that.

garthvh:
Yeah this seems pretty confusing and not really necessary, the nav needs to be finished and then we can look at the value of some of this linking back and forth.

garthvh:
The details on the new navigation structures are here, eventually it should be pretty easy to build all sorts of custom shortcuts for things like this https://github.com/meshtastic/Meshtastic-Apple/issues/857

PR #864, opened 2024-08-10
0 Message Time Logic

garthvh:
Use RX time for messages unless it is off by more than 10 seconds in either direction and then use the current time from the phone to prevent bad times in the past or future from affecting message sort.

Everything else will continue to use RX time and NSOrderedSets

### DIFF
--- a/Meshtastic/Helpers/MeshPackets.swift
+++ b/Meshtastic/Helpers/MeshPackets.swift
@@ -808,8 +808,7 @@ func textMessageAppPacket(
 			let fetchedUsers = try context.fetch(messageUsers)
 			let newMessage = MessageEntity(context: context)
 			newMessage.messageId = Int64(packet.id)
-			/// For message display if the rx time is off by more than 20 seconds set the timestamp to now to assist sorting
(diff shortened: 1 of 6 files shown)


Meshtastic/Helpers/MeshPackets.swift
```diff
@@ -808,7 +808,12 @@ func textMessageAppPacket(
 			let fetchedUsers = try context.fetch(messageUsers)
 			let newMessage = MessageEntity(context: context)
 			newMessage.messageId = Int64(packet.id)
-			newMessage.messageTimestamp = Int32(bitPattern: packet.rxTime)
+			/// For message display if the rx time is off by more than 20 seconds set the timestamp to now to assist sorting
+			if Date().timeIntervalSince1970 < Double(packet.rxTime - 10000) || Date().timeIntervalSince1970 > Double(packet.rxTime + 10000) {
```
thebentern: Logic looks sound. I guess the only caveat is that this will potentially always be true for messages that are queued up in the phone API on a disconnected device once connected and they enter the app.

thebentern (approved):

Issue #920, opened 2024-09-08
🚀 [Feature Request]: allow users to hide node "possible locations" ranges on MeshMap


powersjcb:
### OS

iOS

### Description

**Problem**

There's a handful of users in my area that have very large position uncertainty configured. These are causing 1-10 mi circles to appear on the map for each of them.

<img width="231" alt="image" src="https://github.com/user-attachments/assets/0c2f4934-e7e1-445b-99cc-261598f36ab5">

**Proposed Solution**

* Add a new MeshMap setting to disable range circles
* Add a new position uncertainty value to the node detail view on MeshMap.

### Participation

- [X] I am willing to submit a pull request for this issue.

### Additional comments

Perhaps in the future, deep linking to a node could launch the map UI and center on that node. With the node selected, it enables the uncertainty distance circle for just that node and sets the minimum zoom distance to be ~2-3x the size of that circle. (so the user is made aware if the position is not accurate)

powersjcb:
<img width="369" alt="image" src="https://github.com/user-attachments/assets/48036c12-093c-47d9-a1f6-f4fa394a21b3">

powersjcb:
wow, someone on reddit separately posted a question about what these circles are about a week ago.

They included a nearly nearly identical screenshot of the issue we are seeing on the SLC area map. 

https://www.reddit.com/r/meshtastic/comments/1f5xit0/what_do_the_rings_mean_on_the_meshtastic_app_map/

powersjcb:
splitting out this ticket to help users understand these circles when they are shown. https://github.com/meshtastic/Meshtastic-Apple/issues/924

garthvh:
Lines seems great, hiding imprecise locations does not match the direction the project is going.

PR #921, opened 2024-09-08
MeshMap: add setting to hide location precision circles

powersjcb:
Resolves: https://github.com/meshtastic/Meshtastic-Apple/issues/920

## What changed?
Added menu option to hide location precision circles for nodes

## Why did it change?
Allows for cleaner UIs when map areas have users with very low location precision

## How is this tested?
Manually on iPhone

https://github.com/user-attachments/assets/9ea4086b-7e19-435e-8840-48984ad6ac81

## Checklist

- [x] I have made corresponding changes to the documentation.

### DIFF
--- a/Meshtastic/Extensions/UserDefaults.swift
+++ b/Meshtastic/Extensions/UserDefaults.swift
@@ -50,6 +50,7 @@ extension UserDefaults {
 		case meshMapRecentering
 		case meshMapShowNodeHistory
 		case meshMapShowRouteLines
+		case meshMapShowLocationPrecision
 		case enableMapConvexHull
 		case enableMapRecentering
 		case enableMapNodeHistoryPins
@@ -96,6 +97,9 @@ extension UserDefaults {
 
 	@UserDefault(.meshMapDistance, defaultValue: 800000)
 	static var meshMapDistance: Double
+	
+	@UserDefault(.meshMapShowLocationPrecision, defaultValue: true)
+	static var meshMapShowLocationPrecision: Bool
 
 	@UserDefault(.enableMapWaypoints, defaultValue: false)
 	static var enableMapWaypoints: Bool
--- a/Localizable.xcstrings
+++ b/Localizable.xcstrings
@@ -19884,6 +19884,9 @@
     "Show Alerts" : {
 
     },
+    "Show Location Precision" : {
+
+    },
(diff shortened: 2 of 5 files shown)


powersjcb:
I'm blocked testing this change at the moment since I don't have the ability to load real node data.

Meshtastic/Extensions/UserDefaults.swift
```diff
@@ -96,6 +97,9 @@ extension UserDefaults {
 
 	@UserDefault(.meshMapDistance, defaultValue: 800000)
 	static var meshMapDistance: Double
+	
+	@UserDefault(.meshMapShowLocationPrecision, defaultValue: true)
```
powersjcb, line 101: there's three default values related to these configs (which is a bit confusing)

powersjcb:
I finally figured out how to get this compiling for my phone. However, even with the business logic disabled that hides location precision I'm only seeing a circle on the map for my own node.

TBD what's going on here, but I suspect we only ingest position precision when the iphone app sees certain events have happened.

garthvh (changes requested):
Since most locations are going to be imprecise now I am not a huge fan of this change, if we did implement it, it would need to filter the data in the query and not just hide the circle, we don't use user defaults for the filters.

powersjcb:
At the moment, this is just an attempt to allow users to "opt-out of clutter" if they don't care about knowing how precise node positions are. 

See this example where entire regions of the map are filled by Location Precision Circles for nodes that have ranges of 5mi and 15mi. When zoomed in the entire map has fill colors driven by the two nodes.

image
image

@garthvh, as a more minimalist change, what would you think about removing or allowing users to toggle fill color for Location Precision Circles?

This wouldn't change the ability to see uncertainty, but it would prevent very large Location Precision Circles from applying fill color to all the nodes they are encapsulating them.

Do you have any sense for how stable we are with this UI? Given the comments on reddit, it feels like it would be helpful to build a tooltip map-legend UI once it's stable.
> Press `(?)` icon on map -> launch modal with map-legend.

powersjcb:
image

garthvh:
I really am not seeing a need for this right now, the circles are like 20% and several can stack fine on top of each other. May revisit after 2.5 is the majority firmware.

PR #960, opened 2024-10-10
Changes for 2.5.9

garthvh:
## What changed?
* Update protos
* Update Device hardware JSON

### DIFF
--- a/Meshtastic/Helpers/BLEManager.swift
+++ b/Meshtastic/Helpers/BLEManager.swift
@@ -840,7 +840,6 @@ class BLEManager: NSObject, CBPeripheralDelegate, MqttClientProxyManagerDelegate
 						let logString = String.localizedStringWithFormat("mesh.log.traceroute.received.direct %@".localized, String(snr))
 						MeshLogger.log("🪧 \(logString)")
 					} else {
-						
 						guard let connectedNode = getNodeInfo(id: Int64(connectedPeripheral.num), context: context) else {
 							return
 						}
@@ -880,8 +879,8 @@ class BLEManager: NSObject, CBPeripheralDelegate, MqttClientProxyManagerDelegate
 								if decodedInfo.packet.rxTime > 0 {
 									hopNode?.lastHeard = Date(timeIntervalSince1970: TimeInterval(Int64(decodedInfo.packet.rxTime)))
 								}
-								hopNodes.append(traceRouteHop)
 							}
+							hopNodes.append(traceRouteHop)
 							routeString += "\(hopNode?.user?.longName ?? (node == 4294967295 ? "Repeater" : String(hopNode?.num.toHex() ?? "unknown".localized))) \(hopNode?.viaMqtt ?? false ? "MQTT" : "") (\(traceRouteHop.snr > 0 ? hopNode?.snr ?? 0.0 : 0.0)dB) --> "
 						}
 						let destinationHop = TraceRouteHopEntity(context: context)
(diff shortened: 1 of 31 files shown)


Meshtastic/Views/Layouts/TraceRoute.swift
```diff
@@ -56,13 +56,13 @@ struct TraceRoute: Layout {
 
 			subview.place(at: point, anchor: .center, proposal: .unspecified)
 
-			DispatchQueue.main.async {
+		//	DispatchQueue.main.async {
```
bjpetit, line 59: I was running this branch earlier today and noticed that this change was causing the arrows between nodes to no longer angle correctly. They all pointed one direction until I uncommented these.

Meshtastic/Helpers/BLEManager.swift
```diff
@@ -874,11 +881,26 @@ class BLEManager: NSObject, CBPeripheralDelegate, MqttClientProxyManagerDelegate
...
+						destinationHop.num = traceRoute?.node?.num ?? 0
+						if let mostRecent = traceRoute?.node?.positions?.lastObject as? PositionEntity, mostRecent.time! >= Calendar.current.date(byAdding: .hour, value: -24, to: Date())! {
+							destinationHop.altitude = mostRecent.altitude
+							destinationHop.latitudeI = mostRecent.latitudeI
+							destinationHop.longitudeI = mostRecent.longitudeI
+							traceRoute?.hasPositions = true
+						}
+						hopNodes.append(destinationHop)
+						/// Add the destination node to the end of the route towards string and the beginning of teh route back string
+						routeString += "\(traceRoute?.node?.user?.longName ?? "unknown".localized) \((traceRoute?.node?.num ?? 0).toHex()) \(traceRoute?.node?.snr ?? 0 > 0 ? traceRoute?.node?.snr ?? 0 : 0.0)dB)"
```
GUVWAF, line 901: Isn't this checking whether the SNR is > 0? I'm seeing 0.0dB when I expect a negative SNR. 
Also, if it's an unknown SNR (due to old firmware), this value will be -128.

Meshtastic/Views/Layouts/TraceRoute.swift
GUVWAF: For me, the app crashed when going from a multi-hop traceroute to a direct traceroute in the list. With this, this doesn't happen anymore, but indeed the arrows don't angle correctly anymore.

Meshtastic/Helpers/BLEManager.swift
GUVWAF: I think it would also be better to use the SNR values from the traceroute and not those corresponding to a node, because those will change when subsequent packets are received.

Meshtastic/Views/Layouts/TraceRoute.swift
bjpetit: Ok. I reproduced that crash. Looks like it's probably referencing the subview data after it's getting torn down. 
On the flip side, those assignments in this state trigger a "Modifying state during view update, this will cause undefined behavior” warning.

Meshtastic/Helpers/BLEManager.swift
```diff
@@ -840,24 +840,31 @@ class BLEManager: NSObject, CBPeripheralDelegate, MqttClientProxyManagerDelegate
...
+						guard let connectedNode = getNodeInfo(id: Int64(connectedPeripheral.num), context: context) else {
+							return
+						}
 						var hopNodes: [TraceRouteHopEntity] = []
 						let connectedHop = TraceRouteHopEntity(context: context)
-						connectedHop.name = traceRoute?.node?.user?.longName ?? "unknown".localized
 						connectedHop.time = Date()
+						connectedHop.num = connectedPeripheral.num
+						connectedHop.name = connectedNode.user?.longName ?? "???"
+						connectedHop.snr = Float(routingMessage.snrBack.last ?? 0 / 4)
```
powersjcb, line 851: Is `0 / 4` intended here? Seems like this would resolve to `0`.

PR #976, opened 2024-10-23
Traceroute SNR cleanup

bjpetit:
## What changed?

This change updates traceroute handling of SNR values.

1. For direct traceroute results, use the SNR from the routing message, rather than pulling it from the node entry. Pulling from the node entry causes every direct traceroute entry to show the current SNR for the node, rather than the SNR at the time of the traceroute
2. cleans up a couple spots where the SNR is divided by 4. The change casts them to Floats first then divides so that the decimals aren't truncated.
3. Clean up some incorrect snr checks against 0. SNR can be negative. An unknown SNR is represented as -32. When the SNR value is encoded as an int in the routing message, the value is -128.
4. Fix up some formatting issues with the routeString and routeBackString

## Why did it change?

These changes clean up some areas which were returning incorrect SNR values as well as some incorrect formatting of the route strings

## How is this tested?

This change was tested with multiple traceroute runs and results

## Checklist

### DIFF
--- a/Meshtastic/Helpers/BLEManager.swift
+++ b/Meshtastic/Helpers/BLEManager.swift
@@ -835,8 +835,8 @@ class BLEManager: NSObject, CBPeripheralDelegate, MqttClientProxyManagerDelegate
 					let traceRoute = getTraceRoute(id: Int64(decodedInfo.packet.decoded.requestID), context: context)
 					traceRoute?.response = true
 					if routingMessage.route.count == 0 {
-						let snr = routingMessage.snrBack.count > 0 ? routingMessage.snrBack[0] / 4 : 0
-						traceRoute?.snr = Float(snr)
+						let snr = routingMessage.snrBack.count > 0 ? Float(routingMessage.snrBack[0]) / 4 : 0.0
+						traceRoute?.snr = snr
 						let logString = String.localizedStringWithFormat("mesh.log.traceroute.received.direct %@".localized, String(snr))
 						MeshLogger.log("🪧 \(logString)")
 					} else {
@@ -848,7 +848,7 @@ class BLEManager: NSObject, CBPeripheralDelegate, MqttClientProxyManagerDelegate
 						connectedHop.time = Date()
 						connectedHop.num = connectedPeripheral.num
 						connectedHop.name = connectedNode.user?.longName ?? "???"
-						connectedHop.snr = Float(routingMessage.snrBack.last ?? 0 / 4)
+						connectedHop.snr = Float(routingMessage.snrBack.last ?? 0) / 4
 						if let mostRecent = traceRoute?.node?.positions?.lastObject as? PositionEntity, mostRecent.time! >= Calendar.current.date(byAdding: .hour, value: -24, to: Date())! {
 							connectedHop.altitude = mostRecent.altitude
 							connectedHop.latitudeI = mostRecent.latitudeI
@@ -866,7 +866,7 @@ class BLEManager: NSObject, CBPeripheralDelegate, MqttClientProxyManagerDelegate
 							let traceRouteHop = TraceRouteHopEntity(context: context)
 							traceRouteHop.time = Date()
 							if routingMessage.snrTowards.count >= index + 1 {
-								traceRouteHop.snr = Float(routingMessage.snrTowards[index] / 4)
+								traceRouteHop.snr = Float(routingMessage.snrTowards[index]) / 4
 							}
 							if let hn = hopNode, hn.hasPositions {
 								if let mostRecent = hn.positions?.lastObject as? PositionEntity, mostRecent.time! >= Calendar.current.date(byAdding: .hour, value: -24, to: Date())! {
@@ -888,7 +888,7 @@ class BLEManager: NSObject, CBPeripheralDelegate, MqttClientProxyManagerDelegate
 						let destinationHop = TraceRouteHopEntity(context: context)
 						destinationHop.name = traceRoute?.node?.user?.longName ?? "unknown".localized
 						destinationHop.time = Date()
-						destinationHop.snr = Float(routingMessage.snrTowards.last ?? 0 / 4)
+						destinationHop.snr = Float(routingMessage.snrTowards.last ?? 0) / 4
 						destinationHop.num = traceRoute?.node?.num ?? 0
 						if let mostRecent = traceRoute?.node?.positions?.lastObject as? PositionEntity, mostRecent.time! >= Calendar.current.date(byAdding: .hour, value: -24, to: Date())! {
 							destinationHop.altitude = mostRecent.altitude
@@ -910,7 +910,7 @@ class BLEManager: NSObject, CBPeripheralDelegate, MqttClientProxyManagerDelegate
 							traceRouteHop.time = Date()
 							traceRouteHop.back = true
 							if routingMessage.snrBack.count >= index + 1 {
-								traceRouteHop.snr = Float(routingMessage.snrBack[index] / 4)
+								traceRouteHop.snr = Float(routingMessage.snrBack[index]) / 4
 							}
 							if let hn = hopNode, hn.hasPositions {
 								if let mostRecent = hn.positions?.lastObject as? PositionEntity, mostRecent.time! >= Calendar.current.date(byAdding: .hour, value: -24, to: Date())! {
--- a/Meshtastic/Helpers/BLEManager.swift
+++ b/Meshtastic/Helpers/BLEManager.swift
@@ -867,6 +867,9 @@ class BLEManager: NSObject, CBPeripheralDelegate, MqttClientProxyManagerDelegate
 							traceRouteHop.time = Date()
 							if routingMessage.snrTowards.count >= index + 1 {
 								traceRouteHop.snr = Float(routingMessage.snrTowards[index]) / 4
+							} else {
+								// If no snr in route, try last snr from node
+								traceRouteHop.snr = hopNode?.snr ?? 0
 							}
 							if let hn = hopNode, hn.hasPositions {
 								if let mostRecent = hn.positions?.lastObject as? PositionEntity, mostRecent.time! >= Calendar.current.date(byAdding: .hour, value: -24, to: Date())! {
@@ -883,7 +886,7 @@ class BLEManager: NSObject, CBPeripheralDelegate, MqttClientProxyManagerDelegate
 								}
 							}
 							hopNodes.append(traceRouteHop)
-							routeString += "\(hopNode?.user?.longName ?? (node == 4294967295 ? "Repeater" : String(hopNode?.num.toHex() ?? "unknown".localized))) \(hopNode?.viaMqtt ?? false ? "MQTT" : "") (\(traceRouteHop.snr > 0 ? hopNode?.snr ?? 0.0 : 0.0)dB) --> "
+							routeString += "\(hopNode?.user?.longName ?? (node == 4294967295 ? "Repeater" : String(hopNode?.num.toHex() ?? "unknown".localized))) \(hopNode?.viaMqtt ?? false ? "MQTT" : "") (\(traceRouteHop.snr)dB) --> "
 						}
 						let destinationHop = TraceRouteHopEntity(context: context)
 						destinationHop.name = traceRoute?.node?.user?.longName ?? "unknown".localized
@@ -898,8 +901,8 @@ class BLEManager: NSObject, CBPeripheralDelegate, MqttClientProxyManagerDelegate
 						}
 						hopNodes.append(destinationHop)
 						/// Add the destination node to the end of the route towards string and the beginning of teh route back string
-						routeString += "\(traceRoute?.node?.user?.longName ?? "unknown".localized) \((traceRoute?.node?.num ?? 0).toHex()) \(traceRoute?.node?.snr ?? 0 > 0 ? traceRoute?.node?.snr ?? 0 : 0.0)dB)"
-						var routeBackString = "\(traceRoute?.node?.user?.longName ?? "unknown".localized) \((traceRoute?.node?.num ?? 0).toHex()) \(traceRoute?.node?.snr ?? 0 > 0 ? traceRoute?.node?.snr ?? 0 : 0.0)dB) --> "
+						routeString += "\(traceRoute?.node?.user?.longName ?? "unknown".localized) \((traceRoute?.node?.num ?? 0).toHex()) (\(traceRoute?.node?.snr ?? traceRoute?.node?.snr ?? 0.0)dB)"
+						var routeBackString = "\(traceRoute?.node?.user?.longName ?? "unknown".localized) \((traceRoute?.node?.num ?? 0).toHex()) (\(traceRoute?.node?.snr ?? traceRoute?.node?.snr ?? 0.0)dB) --> "
 						traceRoute?.hopsBack = Int32(routingMessage.routeBack.count)
 						for (index, node) in routingMessage.routeBack.enumerated() {
 							var hopNode = getNodeInfo(id: Int64(node), context: context)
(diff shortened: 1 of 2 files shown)


bjpetit:
Here's an example of the traceroute output.
IMG_0526

Meshtastic/Helpers/BLEManager.swift
```diff
@@ -848,7 +848,7 @@ class BLEManager: NSObject, CBPeripheralDelegate, MqttClientProxyManagerDelegate
 						connectedHop.time = Date()
 						connectedHop.num = connectedPeripheral.num
 						connectedHop.name = connectedNode.user?.longName ?? "???"
-						connectedHop.snr = Float(routingMessage.snrBack.last ?? 0 / 4)
+						connectedHop.snr = Float(routingMessage.snrBack.last ?? 0) / 4
```
powersjcb: Can we add code comments about why we are dividing by 4? (if anyone knows the reason)

Meshtastic/Helpers/BLEManager.swift
```diff
@@ -866,7 +866,10 @@ class BLEManager: NSObject, CBPeripheralDelegate, MqttClientProxyManagerDelegate
 							let traceRouteHop = TraceRouteHopEntity(context: context)
 							traceRouteHop.time = Date()
 							if routingMessage.snrTowards.count >= index + 1 {
-								traceRouteHop.snr = Float(routingMessage.snrTowards[index] / 4)
+								traceRouteHop.snr = Float(routingMessage.snrTowards[index]) / 4
+							} else {
+								// If no snr in route, try last snr from node
+								traceRouteHop.snr = hopNode?.snr ?? 0
```
powersjcb: Why we are pulling readings from the nodeDb here?

Isn't this going to possibly be `SNR(selfNode <- currentHopNode)` when we actually want, `SNR(currentHopNode <- previousHopNode)` for the path of the trace? (the only time this would maybe be correct is if we are certain the nodeDB updated from the trace message already and this was the final hop)

I'd much rather see us display "unknown" if trace data is missing than something we know is likely incorrect.

Meshtastic/Helpers/BLEManager.swift
```diff
@@ -835,8 +835,8 @@ class BLEManager: NSObject, CBPeripheralDelegate, MqttClientProxyManagerDelegate
 					let traceRoute = getTraceRoute(id: Int64(decodedInfo.packet.decoded.requestID), context: context)
 					traceRoute?.response = true
 					if routingMessage.route.count == 0 {
-						let snr = routingMessage.snrBack.count > 0 ? routingMessage.snrBack[0] / 4 : 0
-						traceRoute?.snr = Float(snr)
+						let snr = routingMessage.snrBack.count > 0 ? Float(routingMessage.snrBack[0]) / 4 : 0.0
```
powersjcb: zero is a bad default value for these unless it's impossible for SNR reading values to have that. It seems like it's possible tho since it frequently ranges from -20 to +10.

Meshtastic/Helpers/BLEManager.swift
```diff
@@ -898,8 +901,8 @@ class BLEManager: NSObject, CBPeripheralDelegate, MqttClientProxyManagerDelegate
 						}
 						hopNodes.append(destinationHop)
 						/// Add the destination node to the end of the route towards string and the beginning of teh route back string
-						routeString += "\(traceRoute?.node?.user?.longName ?? "unknown".localized) \((traceRoute?.node?.num ?? 0).toHex()) \(traceRoute?.node?.snr ?? 0 > 0 ? traceRoute?.node?.snr ?? 0 : 0.0)dB)"
-						var routeBackString = "\(traceRoute?.node?.user?.longName ?? "unknown".localized) \((traceRoute?.node?.num ?? 0).toHex()) \(traceRoute?.node?.snr ?? 0 > 0 ? traceRoute?.node?.snr ?? 0 : 0.0)dB) --> "
+						routeString += "\(traceRoute?.node?.user?.longName ?? "unknown".localized) \((traceRoute?.node?.num ?? 0).toHex()) (\(traceRoute?.node?.snr ?? traceRoute?.node?.snr ?? 0.0)dB)"
+						var routeBackString = "\(traceRoute?.node?.user?.longName ?? "unknown".localized) \((traceRoute?.node?.num ?? 0).toHex()) (\(traceRoute?.node?.snr ?? traceRoute?.node?.snr ?? 0.0)dB) --> "
```
powersjcb: Thoughts about using a default of something like `"??"` (unknown) rather than `0.0`? 

Default 0.0 is confusing since it's in the range of possible values.

Meshtastic/Helpers/BLEManager.swift
bjpetit: I'll put a comment in this code. 
The 'quick' answer is that for the routing messages the field is an int32, scaled buy 4. This then represents a float snr as an int. You can see this noted in the protobuf spec for the Routing messages.

Meshtastic/Helpers/BLEManager.swift
bjpetit: Agreed. Let me rework these spots a bit.

Meshtastic/Helpers/BLEManager.swift
bjpetit: I've updated the comments on what we are doing with the snr fields from the routing message. I've also noted the snr value that represents unknown.

Meshtastic/Helpers/BLEManager.swift
bjpetit: I've removed the use of the node snr here. If the routing message doesn't have a valid snr we make it unknown to reflect what really happened during the traceroute.

Meshtastic/Helpers/BLEManager.swift
bjpetit: The firmware will set unknown/unset snr values in routing messages to -128. When we divide by 4 that becomes -32 which is outside the range of valid snr values.
This code is now setting the snr as unknown when it sees -32.

Meshtastic/Helpers/BLEManager.swift
bjpetit: Done.

Meshtastic/Helpers/BLEManager.swift
```diff
@@ -898,8 +906,8 @@ class BLEManager: NSObject, CBPeripheralDelegate, MqttClientProxyManagerDelegate
 						}
 						hopNodes.append(destinationHop)
 						/// Add the destination node to the end of the route towards string and the beginning of teh route back string
```
GUVWAF: In general this looks good to me. 

One thing is that if `hop_start` is not set in the traceroute response (due to nodes with firmware <2.3 in the route), `routeBack` might be incomplete, since we can't add unknown nodes and SNR in this case (because we don't know how many missing hops there were). I think it would be best to skip showing the route back in this case, like what is done for [Android](https://github.com/meshtastic/Meshtastic-Android/blob/b668a21b68465c13d15ca224df785c86ebe4f3ad/app/src/main/java/com/geeksville/mesh/model/RouteDiscovery.kt#L17) and the [CLI](https://github.com/meshtastic/python/blob/646aa981d5fc191f1900da830615dec91cc77403/meshtastic/mesh_interface.py#L584).

bjpetit: I've put a conditional around the routeBack logic. This should match the logic seen in the Andriod app.

Meshtastic/Helpers/BLEManager.swift
```diff
@@ -883,12 +890,13 @@ class BLEManager: NSObject, CBPeripheralDelegate, MqttClientProxyManagerDelegate
 								}
 							}
 							hopNodes.append(traceRouteHop)
-							routeString += "\(hopNode?.user?.longName ?? (node == 4294967295 ? "Repeater" : String(hopNode?.num.toHex() ?? "unknown".localized))) \(hopNode?.viaMqtt ?? false ? "MQTT" : "") (\(traceRouteHop.snr > 0 ? hopNode?.snr ?? 0.0 : 0.0)dB) --> "
+							routeString += "\(hopNode?.user?.longName ?? (node == 4294967295 ? "Repeater" : String(hopNode?.num.toHex() ?? "unknown".localized))) \(hopNode?.viaMqtt ?? false ? "MQTT" : "") (\(traceRouteHop.snr != -32 ? String(traceRouteHop.snr) : "unknown ".localized)dB) --> "
```
GUVWAF: One more thing: "Repeater" is not really accurate, it can also just be a node that doesn't have the encryption key, or for the route back it can be old firmware.

It's used in multiple places. I would just also use "unknown" here.

bjpetit: In that case, would it be reasonable to just drop the "node == 4294967295" handling and let it fall through to the remaining options? It would try to use the node number or fall back to unknown.

GUVWAF: Yes, sounds reasonable to me.

bjpetit: Done.

GUVWAF (approved):

Meshtastic/Helpers/BLEManager.swift
```diff
@@ -835,8 +835,11 @@ class BLEManager: NSObject, CBPeripheralDelegate, MqttClientProxyManagerDelegate
 					let traceRoute = getTraceRoute(id: Int64(decodedInfo.packet.decoded.requestID), context: context)
 					traceRoute?.response = true
 					if routingMessage.route.count == 0 {
-						let snr = routingMessage.snrBack.count > 0 ? routingMessage.snrBack[0] / 4 : 0
-						traceRoute?.snr = Float(snr)
+						// Routing messages snr values are snr * 4 stored as an int
+						// If a traceroute snr value is unknown this field will contain INT8_MIN or -128
+						// After converting to a float here, -32 is our unknown value.
```
powersjcb, line 840: 🥇

Meshtastic/Helpers/BLEManager.swift
```diff
@@ -883,12 +890,13 @@ class BLEManager: NSObject, CBPeripheralDelegate, MqttClientProxyManagerDelegate
 								}
 							}
 							hopNodes.append(traceRouteHop)
-							routeString += "\(hopNode?.user?.longName ?? (node == 4294967295 ? "Repeater" : String(hopNode?.num.toHex() ?? "unknown".localized))) \(hopNode?.viaMqtt ?? false ? "MQTT" : "") (\(traceRouteHop.snr > 0 ? hopNode?.snr ?? 0.0 : 0.0)dB) --> "
+							routeString += "\(hopNode?.user?.longName ?? String(hopNode?.num.toHex() ?? "unknown".localized)) \(hopNode?.viaMqtt ?? false ? "MQTT" : "") (\(traceRouteHop.snr != -32 ? String(traceRouteHop.snr) : "unknown ".localized)dB) --> "
```
powersjcb: nit: this is kinda hard to read with text wrapping.
```
routeString += "\(hopNode?.user?.longName ?? String(hopNode?.num.toHex() ?? "unknown".localized)) \(hopNode?.viaMqtt ?? false ? "MQTT" : "") (\(traceRouteHop.snr != -32 ? String(traceRouteHop.snr) : "unknown ".localized)dB) --> "
```

```
let hopName = hopNode?.user?.longName ?? String(hopNode?.num.toHex() ?? "unknown".localized)
let mqttLabel = hopNode?.viaMqtt ?? false ? "MQTT" : ""
let snrLabel = (traceRouteHop.snr != -32 ? String(traceRouteHop.snr) : "unknown ".localized

routeString += "\(hopName) \(mqttLabel) (\snrLabel)dB) --> "
```
^ after refactoring it becomes clear that we have an unnecessary whitespace if mqttLabel is blank.

```
let hopName = hopNode?.user?.longName ?? String(hopNode?.num.toHex() ?? "unknown".localized)
let mqttLabel = hopNode?.viaMqtt ?? false ? "MQTT " : ""
let snrLabel = (traceRouteHop.snr != -32 ? String(traceRouteHop.snr) : "unknown ".localized

routeString += "\(hopName) \(mqttLabel)(\snrLabel)dB) --> "
```
^ fixt `"MQTT "` + `\(hopName) \(mqttLabel)(\snrLabel)`

Meshtastic/Helpers/BLEManager.swift
```diff
@@ -897,41 +905,50 @@ class BLEManager: NSObject, CBPeripheralDelegate, MqttClientProxyManagerDelegate
...
+									}
+								}
+								traceRouteHop.num = hopNode?.num ?? 0
+								if hopNode != nil {
+									if decodedInfo.packet.rxTime > 0 {
+										hopNode?.lastHeard = Date(timeIntervalSince1970: TimeInterval(Int64(decodedInfo.packet.rxTime)))
+									}
+								}
+								hopNodes.append(traceRouteHop)
+								routeBackString += "\(hopNode?.user?.longName ?? String(hopNode?.num.toHex() ?? "unknown".localized)) \(hopNode?.viaMqtt ?? false ? "MQTT" : "") (\(traceRouteHop.snr != -32 ? String(traceRouteHop.snr) : "unknown ".localized)dB) --> "
```
powersjcb: same nit about formatting + whitespace for mqtt string

Meshtastic/Helpers/BLEManager.swift
bjpetit: Yes. That's more readable. Will make that update.

bjpetit: Done

Meshtastic/Helpers/BLEManager.swift
bjpetit: Also done.

Meshtastic/Helpers/BLEManager.swift
```diff
@@ -942,7 +946,11 @@ class BLEManager: NSObject, CBPeripheralDelegate, MqttClientProxyManagerDelegate
 									}
 								}
 								hopNodes.append(traceRouteHop)
-								routeBackString += "\(hopNode?.user?.longName ?? (node == 4294967295 ? "Repeater" : String(hopNode?.num.toHex() ?? "unknown".localized))) \(hopNode?.viaMqtt ?? false ? "MQTT" : "") (\(traceRouteHop.snr != -32 ? String(traceRouteHop.snr) : "unknown ".localized)dB) --> "
+
+								let hopName = hopNode?.user?.longName ?? String(hopNode?.num.toHex() ?? "unknown".localized)
```
powersjcb: `node == 4294967295` was previously hardcoded here (value of 2^32-1). Any idea why that was the case?

edit: oh, just kidding part of the removed logic for identifying repeaters.

lgtm

powersjcb (approved):

Meshtastic/Helpers/BLEManager.swift
garthvh: The logic identifying routers needs to be in there, why did it get removed?

GUVWAF: It's not necessarily a router (or repeater). It can be a node that can't decrypt because it doesn't have the channel key, or in the route back it's likely a node on firmware before 2.5.

The only case that could cause a repeater to show as `0xFFFFFFFF` as well is if it was set to `ALL_SKIP_DECODING`.

garthvh: Could say repeater (no key) or something but it needs to be in there.  Most users are just using the app and are not digging through the docs.

GUVWAF: Why not just "unknown"? There are multiple options (old firmware is another), and we don't which one it is.

If it's a repeater not set to `ALL_SKIP_DECODING`, it will be its NodeNum, not `0xFFFFFFFF`.

GUVWAF: Ah, or will it show `0xFFFFFFFF` now when it doesn't know the long name? That would indeed be confusing.

bjpetit: Thoughts on next steps here?
It sounds like adding back special handling for 0xFFFFFFFF makes sense. But should it call that out as 'unknown' since a hop could come back with that set for multiple reasons? Or is something more precise that these entries should be named?

powersjcb: `unknown` feels like a clean answer if we cannot say who they are for any reason

garthvh: > Thoughts on next steps here? It sounds like adding back special handling for 0xFFFFFFFF makes sense. But should it call that out as 'unknown' since a hop could come back with that set for multiple reasons? Or is something more precise that these entries should be named?

Yes I want the special handling in there

bjpetit: Should the hopName end up as "unknown" or something else in this path?

garthvh: > Should the hopName end up as "unknown" or something else in this path?

Repeater like it was.

bjpetit: done

garthvh (approved):
Looks good

PR #1014, opened 2024-12-07
Notifications fix

garthvh:
## Checklist


garthvh:
After thinking about this a bit I think the better approach is to make a user if they don't exist.

PR #1044, opened 2025-01-04
Update Localizable.xcstrings

HJen-git:
Replaced incorrect german translation in 2 locations Old: key -> Taste (meaning key as in keyboard)
New: key -> Schlüssel (meaning key as in access key)

## What changed?
Resource entries for 2 german localizations.
Location in app: Settings for channel encryption, key and key length/size.

Line 31336 is an unexpected change and should be omitted.

## Why did it change?
The previous translation made no sense in that context.
The english word key was translated to german as a key you find on a keyboard or piano, but in this context an access/encryption key is meant. The correct translation for such a key is "Schlüssel" as in "Zugangsschlüssel" and "Schlüssellänge"

## How is this tested?
Test correct display of the umlaut (there was already one present, so no problem expected) and for cropping/line break through length.

n/a

## Checklist

- [n/a] I have commented my code, particularly in complex areas.
- [n/a] I have made corresponding changes to the documentation.
- [-] I have tested the change to ensure that it works as intended.

### DIFF
--- a/Localizable.xcstrings
+++ b/Localizable.xcstrings
@@ -14812,7 +14812,7 @@
         "de" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "Taste"
+            "value" : "Schlüssel"
           }
         },
         "sr" : {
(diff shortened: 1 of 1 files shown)


HJen-git:
I noticed a lot of missing translation strings as well.
I don't have any non-solo experience with git yet, but if the way I submitted is sufficient I would like to add any missing translations I find.
I mainly noticed missing strings for signal strength, device modes, status/error codes, ...

PR #1046, opened 2025-01-08
Update traceroute back route display to handle case where route back is not valid

bjpetit:
## What changed?

Fixed issue where traceroute info wasn't saved if route back wasn't valid.
Update route back handling to display "unknown" when the route back info is not valid.

## Why did it change?

Looks like the traceroute data could get dropped if the route back isn't valid. With that fixed, the route back could display "Direct" when the route back data is invalid which is incorrect.

Looking for feedback. The other option is that we just drop references to back messages. But I thought this was a little more clear that the route back data is invalid.

## How is this tested?

Ran various traceroutes and validated the data. I. don't have a setup that will return invalid route back info, so I tested that by artificially dropping the route back processing.

Here's an example of what the message will look like if the route back data is invalid. 
Screenshot 2025-01-08 at 09 16 23 jpeg

## Checklist


GUVWAF:
Great. "unknown" sounds good to me, then it's clear the data is missing.

PR #1037, opened 2024-12-30
Several EnvironmentMetrics enhancements

jake-b:
## What changed?

1. Added additional metrics to available columns/chart series -- Lux, Radiation, Distance, etc.
2. Better Y-axis scaling for the Environment Metrics chart (Temperature now has a 50-degree range, and Humidity 0-100%)
3. Save the user's preferences for the columns/chart series per-node. Added two attributes to the NodeInfoEntity that hold an array of metrics (one for the chart, and one for the table)
4. Better support for nil values in Environment Metrics -  This is a pretty big refactor, so you might not want to do this, but took the approach mentioned in #1027 (manual code generation)
5. A few new CompactWidgets for Distance/Radiation (Closes #991)

## How was it implemented:
1. Additional metrics are straightforward to add to the Environment Metrics Log view by adding implementations to the `EnvironmentDefaultColumns` and `EnvironmentDefaultSeries` files.
2. The API for `MetricsChartSeries` was extended to provide a `minumumYAxisSpan` and `initialYAxisRange`.  The forces the chart to a 0-100 scale when humidity is added, and keeps the temperature scale to be a 50-degree spread.  If the data requires larger scales, these ranges will be expanded as necessary.
3. `MeshtasticDataModelV 48.xcdatamodel` was created with changes to `NodeInfoEntity`:
- `telemetryColumns` and `telemetrySeries` were added as [String] arrays.  These will hold the list of chart series and the list of table columns that the user has active per-node.
4. The ability to differentiate nil values from zeros in Metrics was implemented as follows:
- Codegen for `TelemetryEntity` was disabled (Manual/none). Manual implementations were provided in `TelemetryEntity+CoreDataClass`and `TelemetryEntity+CoreDataProperties`
- A `ManagedAttribute` property wrapper was created to handle the translation of NSNumbers to scalar optionals.
-  `TelemetryEntity+CoreDataClass` is the manual NSManagedObject class implementation that uses the before metioned `@ManagedAttribute` to implement scalar optional accessors for various metrics attributes.
- `TelemetryEntity+CoreDataProperties` is the manual NSManagedObject extension that contains standard `@NSManaged` implementations
- `MeshPackets.swift` was updated to load nil values into the database when the `hasXXXX` is false, and the actual value with it is true.
- Various updates across the app to handle nil metrics.
5. Radiation and Distance compact widgets were added to expose these values (when not nil) on the NodeInfo view.
6. Minor fixes, typo corrections

## Why did it change?
[...]

PR #1112, opened 2025-02-21
Optional values for TelemetryEntity attributes

jake-b:
## What changed?
Better support for nil values in Environment Metrics

The ability to differentiate nil values from zeros in Metrics was implemented as follows:
- Codegen for TelemetryEntity was disabled (Manual/none). Manual implementations were provided in TelemetryEntity+CoreDataClass and TelemetryEntity+CoreDataProperties
- A ManagedAttribute property wrapper was created to handle the translation of NSNumbers to scalar optionals.
- TelemetryEntity+CoreDataClass is the manual NSManagedObject class implementation that uses the before mentioned @ManagedAttribute to implement scalar optional accessors for various metrics attributes.
- TelemetryEntity+CoreDataProperties is the manual NSManagedObject extension that contains standard @NSManaged implementations
- MeshPackets.swift was updated to load nil values into the database when the hasXXXX is false, and the actual value with it is true.
- Various updates across the app to handle nil metrics.

This pull request is part of the work done in draft PR #1037 which will be withdrawn.

Also, where I touched the source code, I tried to implement `.formatted()` and minimize the creation of unnecessary localization keys. (as suggested by @4np)

## Why did it change?
Currently, the app assumes that all nodes that provide telemetry are providing a `temperature` value.  While this is probably true, it is also possible to have nodes that do not provide temperature.  "Zero" values may be displayed in the UI when really no values are present. 

## How is this tested?
Built and running on my development device.  Recommend full TestFlight as the changes touch a large portion of the app has been touched to handle the optionals.  For example, I could not fully test the PowerMetrics stuff from @daviesgeek as I don't have the hardware.

Not much in the app visibly changes with this pull request, but it sets the stage for more Compact Widgets on the NodeInfo page and other features.

## Checklist

### DIFF
--- a/Localizable.xcstrings
+++ b/Localizable.xcstrings
@@ -36,6 +36,12 @@
         }
       }
     },
+    " %@%%" : {
+
+    },
+    "--" : {
+      "shouldTranslate" : false
+    },
     ": %@" : {
       "localizations" : {
         "sr" : {
@@ -67,6 +73,9 @@
           }
         }
       }
+    },
+    "?" : {
+
     },
     "(Re)define PIN_GPS_EN for your board." : {
       "localizations" : {
@@ -728,6 +737,9 @@
           }
         }
       }
+    },
+    "%f%%" : {
+
     },
     "%lf" : {
       "localizations" : {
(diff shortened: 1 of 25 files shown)


jake-b:
Accidentally merged #1108 in my branch.  I can back that out if necessary, but if 1108 is accepted then it won't be necessary.

PR #1116, opened 2025-02-28
Filter variation selectors from Short Name field

jake-b:
## What changed?

The iOS keyboard appends Unicode 'variation selectors' (extra bytes) to some characters.  This will cause the validation on the shortName field to block using these emojis because they appear as too many bytes. 

This pull request will filter out the variation selectors from the shortName field allowing more emoji to be used in short names.

## Why did it change?

Some, but not all emoji have 2 presentations.  A colored graphical representation (❤️ ) and a black and white representation (♥).  The unicode standard defines additional bytes, called "variation selectors" that allow you to select the variant you want.

In the case of short name, we probably don't care.  When the variation selector is not provided, [the default is the emoji style](https://en.wikipedia.org/wiki/Emoticons_(Unicode_block)#Variant_forms).

This pull request will filter the variation selectors from the short name field, unlocking additional Emoji.  Note that there are still quite a few emoji that are more than 4 bytes (such as the country flags) that are still not allowed. 

## How is this tested?

Generally testing it locally.  I am NOT an expert in Unicode, so I've tried to explain myself above.  I am hopeful that this does not cause bad data to find its way into the Short Name field and cause downstream problems, but my node happily has the Short Name of ☢️ .

image

## Checklist

### DIFF
--- a/Meshtastic/Extensions/String.swift
+++ b/Meshtastic/Extensions/String.swift
@@ -90,4 +90,15 @@ extension String {
 		let end = index(start, offsetBy: range.upperBound - range.lowerBound)
 		return String(self[start ..< end])
 	}
+
+	// Filter out variation selectors from the string
+	var withoutVariationSelectors: String {
+		return self.unicodeScalars
+			.filter { scalar in
+				return !scalar.properties.isVariationSelector
+			}
+			.compactMap { UnicodeScalar($0) }
+			.map { String($0) }
+			.joined()
+	}
 }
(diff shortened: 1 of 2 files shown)


garthvh:
Does this handled large values pasted in?

jake-b:
> Does this handled large values pasted in?

Yes. This maintains the current method of handling large values.  When `totalBytes > 4`, the shortName is truncated with `dropLast()`. This triggers the onChange again, and if the value is still larger than 4 bytes, it will truncate again, until the string passes the 4 byte check.

This could be improved using a loop, but it's not just a matter of something like `.prefix(4)`.  It would involve walking the string, adding character by character until we hit a maximum of 4 bytes to make sure the boundary doesn't fall within a multi-byte character.

PR #1137, opened 2025-03-15
Improvements to emoji handling for node names

jake-b:
## What changed?
See #1116 for a discussion on variation selectors and how they work.

This pull request:
- Removes variation selectors from the longName field (previously only shortName) to save a few bytes.
- adds a `addingVariationSelectors` function to the String extension to add back a preference for the graphical/color variant of any emoji in the string.
- uses this new extension in the `CircleText` and `ConnectedDevice` views so that they prefer the graphical/color representations of any Emoji.

## Why did it change?
The graphical/color emojis are mostly preferred on iOS by default.  But on Mac Catalyst, its seems the B&W glyphs are preferred. This will hopefully make the behavior consistent across devices.

## How is this tested?
Tested on Mac OS and iOS with various emoji combinations, but probably more extensive testing is necessary.
Also there may be other places in the app that need `addingVariationSelectors` applied to a node long or short name.

## Checklist

### DIFF
--- a/Meshtastic/Extensions/String.swift
+++ b/Meshtastic/Extensions/String.swift
@@ -101,4 +101,30 @@ extension String {
 			.map { String($0) }
 			.joined()
 	}
+
+	// Adds variation selectors to prefer the graphical form of emoji.
+	// Looks ahead to make sure that the variation selector is not already applied.
+	var addingVariationSelectors: String {
+		var result = ""
+		var scalars = self.unicodeScalars
+		var index = scalars.startIndex
+		while index < scalars.endIndex {
+			let currentScalar = scalars[index]
+			result += String(currentScalar)
+			if currentScalar.properties.isEmoji && !currentScalar.properties.isEmojiPresentation {
+				// Check if the next scalar is U+FE0F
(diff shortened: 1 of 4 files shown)


jake-b:
I'm going to move this to draft, because I think we need more though on this.
Adding variation selectors to every eligible character results in some characters getting emoji-ified when we probably don't want it.  Such as the 3000 below: 
image

Unicode is fun.

jake-b:
I've done some more testing. This PR works as follows:

- Variation selectors are only stripped by `withoutVariationSelectors` on emoji's who's base character is ASCII.  (For example 1️⃣2️⃣3️⃣ as these need the variation selector to differentiate them as the emoji variants.
- Variation selectors are only added back to the string in `addingVariationSelectors` when the previous character is an emoji but not if the character is an ASCII character.  This prevents the function from adding variation selectors to `1` `2` `3` etc.
- Added `addingVariationSelectors` to the longName in specific places throughout the app to explicitly ask for the "emoji/color/graphical" version of the emoji.

I think this results in sane behavior, consistent rendering across iOS and Mac Catalyst, and the minimum number of bytes to use emojis with different variants.

Issue #1081, opened 2025-01-28
🐞 [Bug]: Waypoints not visible by default


tekstrand:
### Firmware Version

2.5

### What did you do?

1. Receive a waypoint from another node/client on a net new install

### Expected Behavior

Waypoint should be visible on the map because the default position of "Show Waypoints" toggle is true

### Current Behavior

While the Show Waypoints toggle _displays_ true/on, waypoints are not visible until toggling the switch back and forth

### Participation

- [ ] I am willing to submit a pull request for this issue.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #1203, opened 2025-05-06
Fixed waypoints popup form

RCGV1:
## What changed?

I set all default occurrences of enableMapWaypoints to be true. I also made it so that any unsent waypoints made during the creation process are deleted if a user cancels or leaves the screen.
## Why did it change?

There was a discrepancy in the state of enableMapWaypoints (reported in #1081 ), which caused the toggle switch to be on for waypoints but the actual state to be off, so by default, waypoints would not show up on the mesh map. Also, when creating waypoints but never sending them and cancelling or dismissing the view, a permanent waypoint would show up that was never sent to the mesh and cannot be removed.

## How is this tested?

Tested on two iPhones one with and one without the fix.
Before the Fix:

https://github.com/user-attachments/assets/18bd69ea-ffab-488f-92b8-09a8f83afeaa

https://github.com/user-attachments/assets/86f31466-1706-4270-9b23-dd989d0b6d47

After the Fix:

https://github.com/user-attachments/assets/e292bba9-b326-4563-9d01-2eba50d73b29

## Checklist

### DIFF
--- a/Meshtastic/Extensions/UserDefaults.swift
+++ b/Meshtastic/Extensions/UserDefaults.swift
@@ -98,7 +98,7 @@ extension UserDefaults {
 	@UserDefault(.meshMapDistance, defaultValue: 800000)
 	static var meshMapDistance: Double
 
-	@UserDefault(.enableMapWaypoints, defaultValue: false)
+	@UserDefault(.enableMapWaypoints, defaultValue: true)
 	static var enableMapWaypoints: Bool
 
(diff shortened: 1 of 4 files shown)


thebentern (approved):

RCGV1:
Also commited a fix to a very annoying issue where other gestures overide the gesture to tap a node circle on the map or waypoint circle

garthvh (approved):
Looks good

PR #1209, opened 2025-05-07
Fixed waypoint updating logic and waypoint notification

RCGV1:
## What changed?

Stops making duplicates of waypoints instead of updating a waypoint. Also updates the notification of a new waypoint to show who sent it.
## Why did it change?

An update/deletion of a waypoint created a duplicate of that same waypoint as the packet.id was used for the waypoint id instead of the actual waypoint id.
## How is this tested?

Tested on my iPhone, sending, receiving, and modifying waypoints.

IMG_3750
Realtime updating and deletion of a waypoint:

https://github.com/user-attachments/assets/6738abee-ef08-426e-816c-726abff55404

## Checklist

### DIFF
--- a/Meshtastic/Helpers/LocationsHandler.swift
+++ b/Meshtastic/Helpers/LocationsHandler.swift
@@ -8,6 +8,7 @@
 import SwiftUI
 import CoreLocation
 import OSLog
+import CoreData
 
 // Shared state that manages the `CLLocationManager` and `CLBackgroundActivitySession`.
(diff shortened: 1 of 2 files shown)


RCGV1:
Added a commit to add an extra fallback for device location so it stores the last known location to use just in case.

RCGV1:
An example of the bug:
The user is not that far from the waypoint but because his phone is not getting its own position its falling back to the location of Apple Park

image

PR #1215, opened 2025-05-10
2.6.3 Working Changes

garthvh:
* Auto favorite nodes you send DM's to prevent No Channel errors.
* Mesh Map Favorites filter
* Contact sharing QR code with latest alpha firmware (2.6.9)
* Trace Route Rate Limit Countdown Timer
* Unmessagable Users for Unmonitored nodes based on role. Granular control is available with latest alpha firmware (2.6.9)
* Message conversation timestamps
* Waypoint Fixes
* Trace route notifications link fix
* Conversation Message Timestamp header
* Remove insecure legacy admin
* Device disconnect improvements
* Translation updates
* Voiceover enhancements


RCGV1:
@garthvh I don't think that all nodes you dm should be favorited. Maybe a new feature should be added to firmware to just mark recently dmed nodes as more "interesting" since favoriting a node always shows them at the top and would be strange if that action was done for you automatically and not nodes specifically favorited by the user.

thebentern:
> @garthvh I don't think that all nodes you dm should be favorited. Maybe a new feature should be added to firmware to just mark recently dmed nodes as more "interesting" since favoriting a node always shows them at the top and would be strange if that action was done for you automatically and not nodes specifically favorited by the user.

I think this is a good sensible default actually. It's a good way to preserve interesting nodes from rolling off the nodedb in a way that doesn't require user interaction. It's also pretty trivial unfavorite.

Localizable.xcstrings
```diff
@@ -2136,6 +2136,12 @@
           }
         }
       }
+    },
+    "Add Contact" : {
+
+    },
+    "Add Meshtastic Node %@ as a contact" : {
```
dzienisz, line 2150: Love this change <3 can we make the same feature on web client?

PR #1225, opened 2025-05-14
Add share contact via QR (and untested scanning of contact QR codes)

thebentern:
Can't actually test the entitlement related url handling changes until it is released to testflight sadly

https://github.com/user-attachments/assets/36912da5-dfcf-4728-8f56-99ed1b7e2132

### DIFF
--- a/Localizable.xcstrings
+++ b/Localizable.xcstrings
@@ -15089,10 +15089,10 @@
         }
       }
     },
-    "Import Contact" : {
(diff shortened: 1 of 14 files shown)


thebentern:
@garthvh I updated the protobufs here

garthvh:
Cool I want to gate it by firmware version but we can add that

garthvh (approved):

PR #1226, opened 2025-05-14
Import Contact App Shortcut and add a share url button to the page

RCGV1:
IMG_3766
IMG_3765

### DIFF
--- a/Meshtastic/Views/Nodes/Helpers/ShareContactQRDialog.swift
+++ b/Meshtastic/Views/Nodes/Helpers/ShareContactQRDialog.swift
@@ -61,6 +61,13 @@ struct ShareContactQRDialog: View {
                 .font(.subheadline)
                 .multilineTextAlignment(.center)
                 .foregroundColor(.secondary)
+			ShareLink("Share QR Code & Link",
+						item: Image(uiImage: qrImage),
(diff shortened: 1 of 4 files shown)


Meshtastic/Views/Nodes/Helpers/ShareContactQRDialog.swift
```diff
@@ -61,6 +61,13 @@ struct ShareContactQRDialog: View {
                 .font(.subheadline)
                 .multilineTextAlignment(.center)
                 .foregroundColor(.secondary)
+			ShareLink("Share QR Code & Link",
+						item: Image(uiImage: qrImage),
+					  subject: Text("Import Meshtastic Node \(node.user.shortName) as a contact"),
```
thebentern: I think it would good to remain consistent with terminology and usage of longname vs shortname. 
Add vs Import and shortname vs longname

RCGV1:
@thebentern Are we calling this feature import contact or add contact?

thebentern:
Add Contact sounds more friendly and less like a business process. Let's go with that.

> @thebentern Are we calling this feature import contact or add contact?

thebentern (approved):

PR #1222, opened 2025-05-12
Update serbian translations

nikdale:
## What changed?
Added Serbian Cyrillic translations for various Meshtastic app and device strings related to networking functionality, privacy settings, and device configuration options.

## Why did it change?
To improve accessibility for Serbian users who prefer Cyrillic script. This localization effort enhances the user experience by providing native language support for important system messages, consent notices, and technical terminology.

## How is this tested?
Manual verification by native Serbian speakers to ensure linguistic accuracy and appropriate technical terminology. UI testing will be needed to verify proper rendering of Cyrillic characters across all affected screens and confirm text fits within UI elements.

## Screenshots/Videos (when applicable)
N/A - Text-only changes

## Checklist
- [x] My code adheres to the project's coding and style guidelines.
- [x] I have conducted a self-review of my code.
- [x] I have commented my code, particularly in complex areas.
- [x] I have verified whether these changes require an update to existing documentation or if new documentation is needed, and created an issue in the [docs repo](http://github.com/meshtastic/meshtastic/issues) if applicable.
- [ ] I have tested the change to ensure that it works as intended.
```

PR #1233, opened 2025-05-21
Revert "Update serbian translations"

garthvh:
Reverts meshtastic/Meshtastic-Apple#1222


garthvh:
Broke localized.string file, reverting

PR #1241, opened 2025-06-03
Allow Unmessegable Nodes to be DMed, but instead present a disclamer

RCGV1:
## What changed?

Allows unmessegable nodes to be dmed but instead provides a big disclamer locked at the top that the nodes messages will not be read.
## Why did it change?

Unmessegable is confusing for newer users since its set by defualt on a lot of nodes. Also lots of power users like sending dms to their actually unmessegable routers to test propogation. Even though you can send trace routes dms are less expensive than TRs so we should encourage this type of behavior. Simply hiding unmmesegable nodes is confusing for people and actually discourages its use for power users.

## How is this tested?

Tesed on my iPhone

https://github.com/user-attachments/assets/ad78b767-ac6d-414f-8193-f4a2221b0510

IMG_3946

## Checklist

### DIFF
--- a/Meshtastic/Views/Nodes/NodeList.swift
+++ b/Meshtastic/Views/Nodes/NodeList.swift
@@ -94,7 +94,6 @@ struct NodeList: View {
 			FavoriteNodeButton(bleManager: bleManager, context: context, node: node)
 			/// Don't show message, trace route, position exchange or delete context menu items for the connected node
 			if connectedNode.num != node.num {
-				if !(node.user?.unmessagable ?? true) {
 					Button(action: {
 						if let url = URL(string: "meshtastic:///messages?userNum=\(node.num)") {
 						   UIApplication.shared.open(url)
@@ -102,7 +101,6 @@ struct NodeList: View {
 					}) {
 						Label("Message", systemImage: "message")
 					}
-				}
 				TraceRouteButton(
 					bleManager: bleManager,
 					node: node
--- a/Localizable.xcstrings
+++ b/Localizable.xcstrings
@@ -31205,6 +31205,9 @@
         }
       }
     },
+    "This node is unmessegable and your message will likley not be read" : {
+
+    },
     "This will disable fixed position and remove the currently set position." : {
       "localizations" : {
         "it" : {
--- a/Meshtastic/Views/Messages/UserList.swift
+++ b/Meshtastic/Views/Messages/UserList.swift
@@ -47,7 +47,7 @@ struct UserList: View {
 						  NSSortDescriptor(key: "userNode.lastHeard", ascending: false),
 						  NSSortDescriptor(key: "longName", ascending: true)],
 		predicate: NSPredicate(
-		  format: "userNode.ignored == NO AND unmessagable = NO"
+		  format: "userNode.ignored == NO"
 		), animation: .spring
 	)
 	var users: FetchedResults<UserEntity>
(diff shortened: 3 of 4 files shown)


Meshtastic/Views/Nodes/NodeList.swift
```diff
@@ -94,15 +94,13 @@ struct NodeList: View {
 			FavoriteNodeButton(bleManager: bleManager, context: context, node: node)
 			/// Don't show message, trace route, position exchange or delete context menu items for the connected node
 			if connectedNode.num != node.num {
-				if !(node.user?.unmessagable ?? true) {
```
jedahan, line 97: how about defaulting to false, so that way we can respect those who marked a node unmessagable? And keep everything else you added?

powersjcb:
image

@RCGV1 Have thoughts about adding a setting to display un-messagable nodes which is disabled by default?

We have a similar setting to display "Ignored" nodes. This still complies with @garthvh's original intent of hiding these nodes from the UI, but it continues to permits power users to leverage DMs for range checks.

jedahan:
Traceroutes have been less reliable than DMs in my experience in a dense network using defaults.

powersjcb:
^ Even for non-default configs, SLC mesh is on LF/51 and we still see very low success rates for traceroutes compared to DMs.

The entire point is that we want to be able to explicitly test the mesh performance to remote nodes that are not clients.

erayd:
For what it's worth, I also find the recent app change extremely annoying, and more than a little short-sighted. For a product this widely used, IMO there is a duty of care on the devs to not needlessly break people's workflows.

Changing the mandatory inability to send a DM with a warning, or providing an option in the app to send the DM to an unmessageable node *anyway*, would resolve the issue, while still achieving the underlying goal of making users aware that the node is not monitored for message traffic. 

My particular use-case is for testing paths on the mesh with larger payloads, to verify proper functionality under realistic usage scenario. Often on a marginal path, especially via sites susceptible to collisions, small packets (e.g. traceroute, location exchange etc) will make it through, but larger packets that require more airtime will not. There is currently no other way to run these tests than sending DMs to the target node.

pqhf5kd:
I like seeing ACKs, even if there's never going to be a human reply.

powersjcb:
One interesting thing I noticed yesterday while trying to revert to doing my testing with a laptop was that the App Meshsense is actually running auto-`traceroutes` by default to determine hops-away for all nodes. (meaning that we have airtime contention with bots when attempting to use that feature)

image

RCGV1:
@powersjcb Meshsense is a big ch util user. Don't recommend using it except for really specific cases.

powersjcb:
https://discord.com/channels/867578229534359593/871553765105348668/1379869509522952242

>  it treats any ack, including an implicit ack, as sufficient, which isn't really in line with what folks expect for a DM

^It also appears that the python SDK doesn't actually support real `--sendtext` ACKs. This generally means that the only mechanism we had to message routers and view the ACKs has been removed from the ecosystem. :-\

garthvh:
> > > DMing a node is by far the most accurate way to test router connectivity and this only negatively impacts infra maintainers.
> > 
> > 
> > Should be using traceroutes
> 
> We're quite consistently discouraged from using Trace Routes

The primary firmware developer just told you to use trace routes, I would follow that advice.

powersjcb:
Traceroutes are significantly less reliable if you need to diagnose a busy mesh. He's saying we are discouraged from using them because the firmware de-prios and drops them.

The other point that I don't know if people are getting is that this was the only remaining way to send messages to routers. From some quick research the python app doesn't actually listen for ACKs.

PR #1265, opened 2025-06-15
Update italian translation

Blackmane:
## What changed?
Changed some translations for Italian language

## Why did it change?
There are some translations that do not make sense in Italian, while for others the translation has been made homogeneous between different texts. Added missing translations.

## How is this tested?
Not tested

### DIFF
--- a/Localizable.xcstrings
+++ b/Localizable.xcstrings
@@ -1362,7 +1362,7 @@
         "it" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "🦕 Versione di fine vita 🦖 ☄️"
+            "value" : "🦕 Versione a fine vita 🦖 ☄️"
           }
         },
         "sr" : {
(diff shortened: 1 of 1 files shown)


garthvh:
Did you update this file manually or via Xcode?  It is not resolving properly in Xcode for me.

Blackmane:
I think I've found the problem, a missing closing bracket. I've pushed the branch, can I open again a PR? 
Or do you prefer another approach to the translations?

garthvh:
Happy to merge it if the json is not mangled, please open a new pull, I just can't break all the built in tools and manage a bunch of languages I can't read.

Blackmane:
I understand. I have checked the file with  `jq . Localizable.xcstrings > /dev/null` and now no problem has been reported. I opened a new PR.
I apologize again for the problem, I will be more careful in the future, thanks for your understanding

PR #1269, opened 2025-06-16
Update italian translation

Blackmane:
## What changed?
Changed some translations for Italian language

Why did it change?
There are some translations that do not make sense in Italian, while for others the translation has been made homogeneous between different texts. Added missing translations.

How is this tested?
Not tested with XCode, but I checked it is a valid json


garthvh (approved):
Worked perfect this time, thanks!

PR #1283, opened 2025-06-19
App option for purging nodes not heard in certain amount of time

garthvh:
## What changed?

Add an option to the App Setting to purge nodes entities after they haven't been heard for a certain
number of days.
This purges entries from the app's node list, but not from the connected node.
This will not clear out nodes that have been favorited or ignored
Create a periodic task to check for stale node entries and delete them.

## Why did it change?

I'm in an area where there quite a few nodes coming and going. Node lists in the app can get into the 200-300 range pretty quickly. As a result, we end up with quite a few stale node entities in the app DB.
A number of folks in the area regularly clear their node DB to clear out the stale nodes. This would help manage that.

## How is this tested?

Have been running with various stale times and favorite, ignored settings. Ensured only the expected nodes were deleted.

## Checklist

### DIFF
--- a/Localizable.xcstrings
+++ b/Localizable.xcstrings
@@ -1379,6 +1379,12 @@
         }
       }
     },
+    "0" : {
+
+    },
+    "1" : {
+
+    },
     "1 byte" : {
       "localizations" : {
         "it" : {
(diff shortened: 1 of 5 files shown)


garthvh:
Looks great @bjpetit I did make a few security and timing updates and changed the function to return a bool so that I can re-run the want config to get the nodes from the device in case we have deleted nodes in the node db while we are connected. If not currently connected when the purge happens those nodes will come back on the next connection to the device.

PR #1298, opened 2025-07-04
Datadog RUM monitoring

thebentern:


Meshtastic/MeshtasticApp.swift
```diff
@@ -26,6 +29,28 @@ struct MeshtasticAppleApp: App {
...
+		let clientToken = "pub4427bea20dbdb08a6af68034de22cd3b"
+		let environment = "testflight"
+
+		Datadog.initialize(
+			with: Datadog.Configuration(
+				clientToken: clientToken,
+				env: environment,
+				site: .us5
+			),
+			trackingConsent: .granted
```
thebentern: This is definitely not production-worthy

garthvh: Updated with an opt out user default, matches android

PR #1321, opened 2025-07-18
Add Developer AppIcon

chase9:
Adds a blue developer icon to help tell the dev app apart from the release app. I added this because I couldn't tell the apps apart on my phone and thought it would help others too. I'm happy to build this out into an actual icon picker if you think that's need.

## What changed?
Developer app builds now have a blue icon while release builds maintain the green icon.

## Why did it change?
So developers can tell between the release and development builds on their devices.

## How is this tested?
👀

Simulator Screenshot - iPhone 16 - 2025-07-17 at 20 47 08
Develop Build

Simulator Screenshot - iPhone 16 - 2025-07-17 at 20 51 14
Release Build

## Checklist


thebentern:
@chase9 I really like this idea! Are we sure it won't just replace the icon in the app store though once this gets merged into the next app store release? I assumed the icon defined there was what the app store would use as the manifest

chase9:
I added this as a separate icon bundle and created a new key for the development app icon. In the screenshots above, the green icon is actually this same build but built for release! I think as long as the App Store release is using production builds, we should be good!

chase9:
Another way to test might be through a distribution with test flight.

BTW, I'm also happy to make this into a more "normal" app icon picker feature. That way you can pick to have the dev icon if you like. I was also thinking it would be fun to have icons representing the different mesh clubs around the world.

chase9:
I changed this to be a full App Icon picker, along with a developer icon and an icon representing my local mesh in the Twin Cities!

The developer icon is loosely based on the Xcode icon.

The icon for my local mesh has the colors of the Minnesota flag. I figured that keeping it simple would allow other local meshes to easily add their own icons without changing the logo too much. What do you think about this?

chase9:
@dzienisz can you clarify a bit?

garthvh (changes requested):
If we can use the blue color that matches the TestFlight dot for debug builds that would be fine, and if chirpy could be selected as an alternative icon that would be cool, the local mesh piece is not compliant with or trademark or branding.

garthvh:
> I’ll work on the changes you requested. If you have any ideas about the chirpy icon, let me know!
> 
> For local meshes, the app Apollo had way over 8 icons so maybe it’s more of a “soft” rule? I think it would be very fun for the hobby! I’m guessing not being compliant with the trademark or branding means to remove the meshtastic logo?

The app needs to have the meshtastic m logo (or the mascot chirpy), and there are literally hundreds of community groups so I just don't think that is going to work.

Chirpy is here https://github.com/meshtastic/design/tree/master/chirpy

PR #1341, opened 2025-08-08
Transports Interface to Support TCP for all Platforms and Serial on Mac

garthvh:
* Refactor to add transport interface to allow TCP connections on iOS and Serial for Mac
* Upgrade to swift 6


thebentern (approved):

steffann:
Awesome to see this. Much appreciated and many thanks!

PR #1351, opened 2025-08-15
Remove redundant gotFirstUnreadMessage 

RCGV1:
## What changed?

Remove redundant gotFirstUnreadMessage variable, which prevents everything from being set as read until it is true
## Why did it change?

This could improve performance, as everything will be marked as read on load instead of only when scrolled to the first unread item.
## How is this tested?

Tested on my iPhone
## Checklist


garthvh:
Was this part of the scrolling functionality?

RCGV1:
Technically yes. I wanted it to send the view to the first unread and only then make everything as read but removing this redundant check does the same exact thing.

PR #1409, opened 2025-09-16
Use Hops away value for DM's

garthvh:
Set hopLimit for DM messages (DM's and Exchange position) to the hopsAway value for the node you are sending to.

### DIFF
--- a/Meshtastic/Accessory/Accessory Manager/AccessoryManager+Position.swift
+++ b/Meshtastic/Accessory/Accessory Manager/AccessoryManager+Position.swift
@@ -27,7 +27,7 @@ extension AccessoryManager {
 		}
 	}
 
-	public func sendPosition(channel: Int32, destNum: Int64, wantResponse: Bool) async throws {
+	public func sendPosition(channel: Int32, destNum: Int64, hopsAway: Int32 = 0, wantResponse: Bool) async throws {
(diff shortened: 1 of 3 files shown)


nullrouten0:
Path lengths are highly variable in our mesh …. and this change would introduce a problem.  

 For example, you learn a Node info packet at 2 hops, but then it takes 6 to get there the next time….  this is an extremely common occurrence.

 I can do traceroute to the same destination 10 times in a row and get different hop counts ranging from 2-6 (and values in between) regularly.

DaneEvans:
So, no moving around I take it? 

As far as unilateral poor decisions go, this is one of them

jschrempp:
I see this PR as a problem for the reasons cited above. 

I think a better approach is to only happen when hopsAway is greater than the configured hop limit. Sometimes I want to DM a node that has a hopsAway of 6, but my node has a configured hopLimit of 3. In this case I would like to have the hopLimit of the message I send be 6. 

So, change the hopLimit to the hopsAway value of the destination node when the hopsAway value is greater than the hopLimit; otherwise use the hopLimit.

Stated another way:  Use the configured hopLimit unless the hopsAway for the destination node is greater than hopLimit; in that case use hopsAway for the message.

thebentern:
Right, something analogous to Math.max(lora.hop_limit, node.hops_away) is what would work best

PR #1466, opened 2025-10-16
Getting app to ignore Range Test messages when Range Test mode is off

emilyboda:
## What changed?
This should make the app ignore range test message when range test mode is off

## Why did it change?
When range tests are conducted on the primary channel with default settings, iOS users receive the messages regardless of whether range test mode is turned on. This is annoying for iOS users, since often times people conduct range tests on the primary channel (whether they don't want to change their primary channel from default, or don't realize this is bothersome because they only use Meshtastic on android). This also makes the behavior consistent with the Android app.

## How is this tested?
I am not able to test this and I don't know anyone who can test this. Can you help me out?

None

## Checklist

I'm sorry, I'm not an iOS developer. I just thought this really needed to be done and it seems like a small-ish change. Assistance with the following things would be very much appreciated.


garthvh:
How is this any different than the existing wantRangeTestPackets checking the config value?

PR #1467, opened 2025-10-16
Fix: New messages rendering below the fold requiring manual scrolling

martinbogo:
# Fix: New messages rendering below the fold requiring manual scrolling

## Problem

When new messages arrive in either public channels or private conversations, they frequently render "below the fold" (outside the visible area) and are not immediately visible. Users have to manually drag/scroll the view to see new messages, which disrupts the messaging experience.

## Root Cause

The issue stems from the incomplete scrolling implementation after commit [5842864a](https://github.com/meshtastic/Meshtastic-Apple/commit/5842864a614d67eb2c78dfa052ca27e7254ed6e1) which removed custom scrolling logic in favor of SwiftUI's built-in `.defaultScrollAnchor(.bottom)` modifier. While this simplified the code, it proved unreliable for automatically scrolling to new messages as they arrived.

The previous implementation had:
- Manual `onChange(of: allPrivateMessages.count)` handlers that were removed
- Complex async dispatch queue timing workarounds in `.onAppear`
- iOS 18-only modifiers (`.defaultScrollAnchorTopAlignment()` and `.defaultScrollAnchorBottomSizeChanges()`) that don't work on iOS 17

## Solution

This PR implements a modern, reliable scrolling solution using SwiftUI's `scrollPosition` binding (available in iOS 17+):

### Key Changes

1. **Added `scrollPosition` state binding** - Tracks and controls scroll position programmatically
2. **Implemented `.scrollPosition(id:anchor:)` modifier** - Leverages iOS 17's native scroll position binding
3. **Added `onChange(of: allPrivateMessages.count)` handlers** - Detects when new messages arrive
4. **Set `scrollPosition = nil` to trigger auto-scroll** - When combined with `.defaultScrollAnchor(.bottom)`, this reliably scrolls to bottom
5. **Removed iOS 18-only modifiers** - Ensures compatibility with the app's iOS 17.5 deployment target
6. **Simplified `.onAppear` logic** - Removed unnecessary async dispatch queue workarounds
7. **Added smooth animations** - Uses `withAnimation` for polished scrolling transitions

### Technical Details

The `scrollPosition` binding approach works by:
- Maintaining a state variable that tracks the current scroll position ID
- Setting it to `nil` when we want to return to the default anchor (bottom)
- The `.defaultScrollAnchor(.bottom)` modifier defines where `nil` should scroll to
- This is synchronous and reliable, unlike `scrollTo()` which requires timing delays

### Files Modified

- `Meshtastic/Views/Messages/ChannelMessageList.swift`
- `Meshtastic/Views/Messages/UserMessageList.swift`

## Testing

✅ Tested on macOS (Apple Silicon) - Debug build
- New channel messages now automatically scroll into view
- Direct messages automatically scroll into view
- Keyboard focus correctly scrolls to bottom
- Smooth animations throughout
- No manual scrolling required

## Benefits

- **More reliable**: Uses SwiftUI's native binding mechanism instead of timing-dependent workarounds
- **Better UX**: New messages are immediately visible without user intervention
- **Cleaner code**: Removes complex async dispatch queue timing logic
- **Future-proof**: Uses modern iOS 17+ APIs while maintaining compatibility
- **Consistent**: Works uniformly across channels and direct messages

## Compatibility

- ✅ iOS 17.5+ (deployment target)
- ✅ macOS (Catalyst)
- ✅ iPadOS 17.5+

## Related Issues

Addresses the "messages below the fold" issue where new messages don't automatically scroll into view.

### DIFF
--- a/Meshtastic/Views/Messages/ChannelMessageList.swift
+++ b/Meshtastic/Views/Messages/ChannelMessageList.swift
@@ -22,6 +22,7 @@ struct ChannelMessageList: View {
 	@State private var redrawTapbacksTrigger = UUID()
 	@AppStorage("preferredPeripheralNum") private var preferredPeripheralNum = -1
 	@State private var messageToHighlight: Int64 = 0
+	@State private var scrollPosition: Int64?
 	@FetchRequest private var allPrivateMessages: FetchedResults<MessageEntity>
 	
 	init(myInfo: MyInfoEntity, channel: ChannelEntity) {
@@ -84,12 +85,7 @@ struct ChannelMessageList: View {
 							  if !message.read && UIApplication.shared.applicationState == .active {
 								  message.read = true
 								  LocalNotificationManager().cancelNotificationForMessageId(message.messageId)
-								  // Race condition, sometimes the app doesn't update unread count if we run this too early
-								  // So, run it in the main queue after everything saves and stabilizes
-								  DispatchQueue.main.async {
-									  markMessagesAsRead()
-									  scrollView.scrollTo("bottomAnchor", anchor: .bottom)
-								  }
+								  markMessagesAsRead()
 							  }
 						  }
 						  .id(redrawTapbacksTrigger)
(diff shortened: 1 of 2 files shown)


garthvh:
Anyone using iOS 17 can upgrade to 18, not really interested in this extra complexity and performance hit.

martinbogo:
The issue occurs all the way through iOS 26.  New messages appear 'below the fold' and are not visible when they arrive.

garthvh:
All of the modifiers being used are supported on 18 and 26, 17 is not supported but anyone running 17 can upgrade at least to 18.

PR #1471, opened 2025-10-17
Fix: "Retrieving nodes" significantly slower after reconnect (#1424)

### LINKED ISSUE #1424: 🐞 [Bug]: "Retrieving nodes" gets significantly slower after disconnect and reconnect
### Before submitting

### Firmware Version

iOS 2.7.3 and macOS 2.7.0

### What did you do?

The "Retrieving nodes" step is relatively fast the first time I connect after starting the macOS or iOS app. (1-2 seconds, TCP, 250 nodes in NodeDB.) After disconnect, and then reconnect, it crawls very slowly (like 10+ seconds).

Quitting the macOS app (2.7.0) and starting the app again makes it load fast again!

Same happens to me on a different node with the iOS app (2.7.3) to a BLE node. "Retrieving nodes" very slow, 10+ seconds, after disconnect and reconnect. 

Definitely more noticable on TCP, but still noticable on BLE too.

Seems likely to be an issue on the iOS/macOS app side, since force-killing fixes things, and `meshtastic --nodes` from the Python CLI doesn't change in performance.

### Expected Behavior

"Retrieving nodes" should be approximately the same speed when reconnecting after disconnect

### Current Behavior

"Retrieving nodes" is significantly slower when reconnecting after disconnect

### Participation

### Additional comments

_No response_

### Code of Conduct

martinbogo:
## Problem

After disconnect and reconnect, the "Retrieving nodes" step became significantly slower (10+ seconds vs 1-2 seconds on first connect). With 250 nodes in NodeDB, the performance degradation was very noticeable. Force-quitting and restarting the app would restore fast performance temporarily.

## Root Cause

The `nodeInfoPacket()` function called `context.save()` for every single NodeInfo packet received during database sync. With 250 nodes, this meant 250 individual CoreData save operations. On first connection CoreData is fresh and fast, but on reconnect CoreData has accumulated change tracking, undo management, and memory pressure, making each save progressively slower.

## Solution

- Added `deferSave` parameter to `nodeInfoPacket()` function
- During database retrieval (`.retrievingDatabase` state), defer all individual saves
- Perform a single batch save when database retrieval completes (when `NONCE_ONLY_DB` configCompleteID is received)
- Reduces N individual saves to 1 batch save

## Changes

- `Meshtastic/Helpers/MeshPackets.swift`: Added `deferSave` parameter and conditional save logic
- `Meshtastic/Accessory/Accessory Manager/AccessoryManager+FromRadio.swift`: Check state and pass `deferSave: true` during retrieval
- `Meshtastic/Accessory/Accessory Manager/AccessoryManager.swift`: Batch save when database sync completes

## Testing

- ✅ Compiled successfully for macOS (Apple Silicon)
- ✅ App launches and runs correctly
- ✅ Tested with actual node connections
- ✅ "Retrieving nodes" now consistently fast (1-2 seconds) on both first connect and reconnect

Fixes #1424

### DIFF
--- a/Meshtastic/Accessory/Accessory Manager/AccessoryManager+FromRadio.swift
+++ b/Meshtastic/Accessory/Accessory Manager/AccessoryManager+FromRadio.swift
@@ -106,8 +106,11 @@ extension AccessoryManager {
 			return
 		}
 
+		// Check if we're in database retrieval mode to defer saves for performance
+		let isRetrievingDatabase = if case .retrievingDatabase = self.state { true } else { false }
+		
 		// TODO: nodeInfoPacket's channel: parameter is not used
-		if let nodeInfo = nodeInfoPacket(nodeInfo: nodeInfo, channel: 0, context: context) {
+		if let nodeInfo = nodeInfoPacket(nodeInfo: nodeInfo, channel: 0, context: context, deferSave: isRetrievingDatabase) {
 			if let activeDevice = activeConnection?.device, activeDevice.num == nodeInfo.num {
 				if let user = nodeInfo.user {
 					updateDevice(deviceId: activeDevice.id, key: \.shortName, value: user.shortName ?? "?")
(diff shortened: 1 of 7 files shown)


garthvh:
Can you resubmit this without the UIKit text selection feature?  Probably not going to merge that and encumber myself with a bunch of UIKit to maintain.

jake-b:
Created branch `deferred_commit` with just the deferred commit stuff.  The pull request contains alternate message list views that are not part of this change.

martinbogo:
Thank you jake.

garthvh:
Closing in favor of a pull without the UIKit text selection

PR #1495, opened 2025-11-02
Fix: hoplimit of dms would always fallback to hops away of the node even when configured hops was higher

RCGV1:
## What changed?

We now fetch from our own LoRa config instead of doing it from the to node.
## Why did it change?

There was a bug in the code that tried to fetch the hop limit of the node we are sending to, but because we can't get it, we fall back to 0, and this will always be smaller than the hopsaway of the node.
## How is this tested?

I noticed this when analyzing packets; my dm packet would have fewer hops than my configured amount.
## Checklist

### DIFF
--- a/Meshtastic/Views/Nodes/Helpers/Actions/ExchangePositionsButton.swift
+++ b/Meshtastic/Views/Nodes/Helpers/Actions/ExchangePositionsButton.swift
@@ -11,7 +11,7 @@ struct ExchangePositionsButton: View {
 	@State private var isPresentingPositionFailedAlert: Bool = false
 
     var body: some View {
-		let hopsAway = Int32(truncatingIfNeeded: node.hopsAway > connectedNode.loRaConfig?.hopLimit ?? 0 ? node.hopsAway : node.loRaConfig?.hopLimit ?? 0)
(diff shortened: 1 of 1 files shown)


Meshtastic/Views/Nodes/Helpers/Actions/ExchangePositionsButton.swift
```diff
@@ -3,14 +3,15 @@ import SwiftUI
...
+	var connectedNode: NodeInfoEntity
 
 	@EnvironmentObject var accessoryManager: AccessoryManager
 
 	@State private var isPresentingPositionSentAlert: Bool = false
 	@State private var isPresentingPositionFailedAlert: Bool = false
 
     var body: some View {
-		let hopsAway = Int32(truncatingIfNeeded: node.hopsAway > node.loRaConfig?.hopLimit ?? 0 ? node.hopsAway : node.loRaConfig?.hopLimit ?? 0)
+		let hopsAway = Int32(truncatingIfNeeded: node.hopsAway > connectedNode.loRaConfig?.hopLimit ?? 0 ? node.hopsAway : node.loRaConfig?.hopLimit ?? 0)
```
compumike: @RCGV1 I think the final `node.loRaConfig?.hopLimit` is supposed to be `connectedNode.loRaConfig?.hopLimit`

PR #1513, opened 2025-11-26
Fix adding channels

RCGV1:
## What changed?

We now create the initial value for index i var outside the for loop so multiple cchannels dont try to overwrite each other. Also we call channel packet function to update the UI with new channels.
## Why did it change?

Before adding multiple cahnnels via QR would only add one and this would not be reflected in the UI.
## How is this tested?

Tested sending multiple qr codes from android to iOS
## Checklist


Meshtastic/Accessory/Accessory Manager/AccessoryManager+ToRadio.swift
```diff
@@ -450,93 +448,75 @@ extension AccessoryManager {
...
-							guard let mutableChannels = myInfo.channels!.mutableCopy() as? NSMutableOrderedSet else {
-								throw AccessoryError.appError("No channels or channel")
-							}
-							if mutableChannels.first(where: {($0 as AnyObject).name == cs.name }) is ChannelEntity {
-								throw AccessoryError.appError("Channel already exists")
-							}
-						}
-					} catch {
-						Logger.data.error("Failed to find a node MyInfo to save these channels to: \(error.localizedDescription, privacy: .public)")
+					guard let mutableChannels = myInfo.channels?.mutableCopy() as? NSMutableOrderedSet else {
```
Vaidios, line 478: Should it be mutable if we are not mutating it anywhere?

Meshtastic/Accessory/Accessory Manager/AccessoryManager+ToRadio.swift
```diff
@@ -450,93 +448,75 @@ extension AccessoryManager {
...
+				let fetchMyInfoRequest = MyInfoEntity.fetchRequest()
+				fetchMyInfoRequest.predicate = NSPredicate(format: "myNodeNum == %lld", Int64(deviceNum))
+
+				let fetchedMyInfo = try context.fetch(fetchMyInfoRequest)
+				if fetchedMyInfo.count != 1 {
+					throw AccessoryError.appError("MyInfo not found")
+				}
+				
+				// We are trying to add a channel so lets get the last index
+				myInfo = fetchedMyInfo[0]
```
Vaidios, line 466: We could use only the Set of channels instead of entire parent object. Maybe we can have an optional set of channels instead.
So `var myInfo: MyInfoEntity!` would become `var myInfoChannels: NSOrderedSet?`

PR #1536, opened 2025-12-31
Don't alert (with sound: .default) when updating Live Activity

compumike:
I don't need my phone to vibrate every 15 minutes for this. (I suspect it was used to be that way, because the other version was commented out.)


garthvh:
Going to have to revert this, it still only works once with no notification sound.

compumike:
> Going to have to revert this, it still only works once with no notification sound.

@garthvh see https://github.com/meshtastic/Meshtastic-Apple/pull/1548 which may fix this!

PR #1542, opened 2026-01-05
Accessorymanager background discovery

garthvh:
Don't append new devices to the list of discovered devices if the app is in the background, seems to cause crashes


compumike:
@garthvh FYI this change seems to break automatic reconnection when the app is in the background: `shouldAutomaticallyConnectToPreferredPeripheral` / `connectToPreferredDevice` relies on it being present in `devices`.

Test sequence:
- connect to a node (BLE)
- put app in background
- reboot the node
- watch the app logs

With this PR, it just stops at `[Discovery] Found preferred peripheral` and doesn't attempt to reconnect.

PR #1547, opened 2026-01-05
TAK Server (for bridging Meshtastic into TAK clients)

thebentern:
@niccellular's original work. I'm just opening the PR for review


Meshtastic/Helpers/TAK/GenericCoTHandler.swift
```diff
@@ -0,0 +1,399 @@
...
+		/// Use EXI compression on ATAK_FORWARDER port (small, no fountain)
+		case exiDirect
+		/// Use EXI + Fountain coding on ATAK_FORWARDER port (large)
+		case exiFountain
+	}
+
+	/// Classify a CoT message to determine send method
+	func classifySendMethod(for cot: CoTMessage) -> CoTSendMethod {
+		// Self PLI (position)
+		if cot.type.hasPrefix("a-f-G") || cot.type.hasPrefix("a-f-g") {
```
thebentern, line 52: Are there any other atoms designations that could be a PLI packet?

Meshtastic/Helpers/TAK/TAKDataPackageGenerator.swift
```diff
@@ -0,0 +1,261 @@
...
+	// MARK: - Data Package Generation
+
+	/// Generate a TAK data package for ITAK client configuration
+	/// - Parameters:
+	///   - serverHost: The server hostname/IP (default: 127.0.0.1 for localhost)
+	///   - port: The server port
+	///   - useTLS: Whether to use TLS (ssl) with mTLS or plain TCP
+	///   - description: Description shown in TAK client
+	/// - Returns: URL to the generated zip file, or nil if generation failed
+	func generateDataPackage(
```
thebentern, line 28: Maybe configurable at some point, so that we can actually distribute one or more of these over a local net and the apple device can be a TAK server for other clients?

Meshtastic/Helpers/TAK/GenericCoTHandler.swift
niccellular: I believe all TAK clients will use a-f-g as a default, (atom - friendly - ground).

ATAK lets you change this, perhaps other TAK clients can change it too, but the default is typically that.

Meshtastic/Helpers/TAK/TAKDataPackageGenerator.swift
niccellular: The code could be updated to bind 0.0.0.0 vs 127.0.0.1 but that would need testing also unsure how that would impact App Store approval.

thebentern: Additionally, you may want to send a DP with your ipv4 address on the LAN if you want external hosts to connect. Although now that I say this out loud, have we tested that you actually _can_ externally connect to the TAK server?

niccellular: No, all my testing has been with localhost. The current DP generation is out of band of TAK. As in the user, from the Meshtastic-Apple Settings menu, downloads a .zip to save locally. From there they can dispense the DP how they see fit.

niccellular:
The logging is all filterable with "TAK" but yes there is an absurd amount of verbose logging including the output of the entire CoT XML -> Protobuf conversion including byte dumping.

Issue #1402, opened 2025-09-14
🐞 [Bug]: iOS app stalls BLE→MQTT forwarding in background; only resumes on foreground (needs CoreBluetooth restoration + background reconnect)


shinteza:
### Before submitting

- [x] I have searched existing issues to make sure this bug hasn't already been reported
- [x] I have updated to the latest version of the software to verify the issue still exists

### Firmware Version

2.7.1

### What did you do?

**Environment**
- iPhone 14 (latest iOS)
- Heltec T114 v2 (BLE-only, no Wi‑Fi)
- Meshtastic firmware 2.7.7
- Meshtastic iOS app (latest App Store build)
- MQTT broker: Mosquitto (TLS optional)

**Steps**
1. On the node, enabled **MQTT uplink** on the active channel (default public channel; downlink disabled).
2. In the iOS app, enabled **Client Proxy** and configured the MQTT broker (host/port/creds; TLS optional).
3. Verified in **foreground** that telemetry/messages from the node arrive on broker topics.
4. Ensured iOS settings that maximize background continuity:
   - Background App Refresh = **ON** (Meshtastic)
   - Low Power Mode = **OFF**
   - Bluetooth = **ON**
   - Location permission = **Always Allow**
   - App **not** force‑quit (left in app switcher).
5. Sent the app to **background** and waited ~5–30+ minutes (also tested 60–120).
6. Watched broker topics throughout the background interval.
7. Brought the app back to **foreground** and watched topics again.

**Observed outcome**
- Topics update normally while in **foreground**.
- After the app sits in **background**, topics go **quiet**.
- Upon **foregrounding**, updates **resume immediately**.

---

## Technical context / code pointers (what I looked at)

> Goal: Explain why background forwarding stalls and where the code likely needs changes.

### 1) CoreBluetooth central has **no state restoration**
- Expected: initialize the central with a **restoration identifier** and handle `willRestoreState`.
- Current pattern observed (representative):
```swift
// BLETransport.swift (representative)
centralManager = CBCentralManager(
    delegate: delegate,
    queue: .global(qos: .utility)
    // ❌ no CBCentralManagerOptionRestoreIdentifierKey
)

// ... later ...
// ❌ Restoration callback exists but is commented or not wired:
/// func centralManager(_ central: CBCentralManager, willRestoreState dict: [String : Any]) { ... }
```

### 2) MQTT client is **process-lifetime bound**
[...]

PR #1548, opened 2026-01-05
BLE handling & Live Activity update changes (wrapping `Task` with `beginBackgroundTask`) for more reliable background/idle/suspended app behavior

compumike:
@garthvh I think this may resolve https://github.com/meshtastic/Meshtastic-Apple/pull/1536#issuecomment-3709019701 .

I think what's going on is that, per the [Core Bluetooth Programming Guide](https://developer.apple.com/library/archive/documentation/NetworkingInternetWeb/Conceptual/CoreBluetooth_concepts/CoreBluetoothBackgroundProcessingForIOSApps/PerformingTasksWhileYourAppIsInTheBackground.html):

> In addition, the system wakes up your app when any of the [CBCentralManagerDelegate](https://developer.apple.com/documentation/corebluetooth/cbcentralmanagerdelegate) or [CBPeripheralDelegate](https://developer.apple.com/documentation/corebluetooth/cbperipheraldelegate) delegate methods are invoked, allowing your app to handle important central role events, such as when a connection is established or torn down, when a peripheral sends updated characteristic values, and when a central manager’s state changes.
> 
> [...]
> 
> Apps woken up for any Bluetooth-related events should process them and return as quickly as possible so that the app can be suspended again.
> 
> [...]
> 
> Ideally, it should complete the task as fast as possible and allow itself to be suspended again.

My guess is that wrapping `await meshActivity?.update(updatedContent)` in `Task { ... }` enqueues the task, but the app gets suspended before that task runs, so the update never gets pushed.

There may be a similar issue with `BLEConnectionDelegate` handlers all going through `Task { ... }` but that's a harder one to wrap my head around right now.

At least for my testing on iOS 26.2, I've had consistent Live Activity updates for over an hour without opening the app. Curious to see if that works for you as well!

### DIFF
--- a/Meshtastic/Accessory/Accessory Manager/AccessoryManager+Discovery.swift
+++ b/Meshtastic/Accessory/Accessory Manager/AccessoryManager+Discovery.swift
@@ -52,12 +52,7 @@ extension AccessoryManager {
 							existing.rssi = newDevice.rssi
 							self.devices[index] = existing
 						} else {
-							// This is a new device, add it to our list if we are in the foreground
-							if !(self.isInBackground) {
-								self.devices.append(newDevice)
-							} else {
-								Logger.transport.debug("🔎 [Discovery] Found a new device but not in the foreground, not adding to our list: peripheral \(newDevice.name)")
-							}
+							self.devices.append(newDevice)
 						}
 						
 						if self.shouldAutomaticallyConnectToPreferredPeripheral,
@@ -75,6 +70,14 @@ extension AccessoryManager {
 					
 					case .deviceReportedRssi(let deviceId, let newRssi):
 						updateDevice(deviceId: deviceId, key: \.rssi, value: newRssi)
+
+						// If we see RSSI for the preferred device while idle, attempt auto-connect
+						if self.shouldAutomaticallyConnectToPreferredPeripheral,
+						   UserDefaults.autoconnectOnDiscovery,
+						   !self.isConnected, !self.isConnecting,
+						   deviceId.uuidString == UserDefaults.preferredPeripheralId {
+							self.connectToPreferredDevice()
+						}
 					}
 				} catch {
 					break


garthvh:
Those are a different type of background task that meshtastic does not use, the  beginBackgroundTask/endBackgroundTask stuff is for web connected apps. This issue shows a bunch of similar stuff that is not related to meshtastic https://github.com/meshtastic/Meshtastic-Apple/issues/1402

garthvh:
In short: It is generally not a good idea to wrap every didUpdateValueFor call in a beginBackgroundTask.

While your intention is to ensure the app doesn't get suspended while processing data, this approach introduces significant overhead and architectural risks.

Why this is problematic
1. High Frequency & Overhead

didUpdateValueFor can be called dozens of times per second (e.g., if you are streaming sensor data or heart rate). Starting and ending a UIBackgroundTask that frequently is expensive for the OS and can lead to performance bottlenecks or even system-level throttling.

2. The "Background Execution" Fallacy

Core Bluetooth has its own built-in mechanisms for background execution. If you have the bluetooth-central background mode enabled:

iOS will wake your app automatically when a value is updated.

The OS grants you a small slice of execution time (around 10 seconds) specifically to handle that delegate callback.

You don't need a UIBackgroundTask to "stay alive" for the immediate processing of that packet; the system has already provided the window.

3. Race Conditions with Task

By wrapping doUpdate in a Task, you are moving the work to a background thread. Since Task is asynchronous, there is no guarantee of execution order. If your BLE device sends "Packet A" then "Packet B," a Task might finish processing "Packet B" before "Packet A" is done, leading to data corruption or UI inconsistencies.

4. Main Thread Dependency

Calling UIApplication.shared.endBackgroundTask inside DispatchQueue.main.async adds further latency. If the main thread is busy, your background task might expire before the "end" call even executes, leading to the "expiration reached" log and potentially a crash if the system thinks you are leaking tasks.

PR #1580, opened 2026-02-03
fix: Mesh Map missing some nodes

adoxner:
## What changed?
Loosens the filter for Mesh Map to show assumed exact locations so more appear on the map.

## Why did it change?
Nodes with unset or zero-precision (likely manually set) do not appear on the map, though the user expects to see them. 

The Android Meshtastic app doesn't seem to be doing _any_ filtering based on precision, leading to a discrepancy between nodes that can be seen in the Android vs iOS, furthering user confusion. Some filtering based on accuracy is probably warranted, so I kept it at what was potentially the original intent of this `if` statement and allowed all values greater than or equal to 12 bits.

Note the definition of `isPreciseLocation` in PositionEntityExtension.swift:
```swift
var isPreciseLocation: Bool {
  precisionBits == 32 || precisionBits == 0
}
```

## How is this tested?
I have not tested this yet as I haven't figured out how to run on-device so I can test connected to a node.

n/a

## Checklist

Fixes https://github.com/meshtastic/Meshtastic-Apple/issues/1581

### DIFF
--- a/Meshtastic/Views/Nodes/Helpers/Map/MapContent/MeshMapContent.swift
+++ b/Meshtastic/Views/Nodes/Helpers/Map/MapContent/MeshMapContent.swift
@@ -62,7 +62,7 @@ struct MeshMapContent: MapContent {
 					// Imprecise location: fuzz slightly so overlapping nodes are visible and clickable at highest zoom levels.
 					position.fuzzedCoordinate
 				}
-				if 12...15 ~= position.precisionBits || position.isPreciseLocation {
+				if position.precisionBits >= 12 || position.isPreciseLocation {
 					
 					let nodeColor = UIColor(hex: UInt32(position.nodePosition?.num ?? 0))
 					let positionName = position.nodePosition?.user?.longName ?? "?"


pdxlocations:
I'm not sure if this is the issue I was reporting on Discord.  After a "Clear App Data" I now have different set of nodes on the map, but still clearly missing some.  Here's a position packet from one node currently not showing on the map:
```
from: 363606076
to: 1602236699
channel: 15
decoded {
  portnum: POSITION_APP
  payload: "latitude_i: 448430080 longitude_i: -1224835072 altitude: 1200 time: 1770078404 location_source: LOC_MANUAL ground_speed: 0 ground_track: 0 precision_bits: 16"
  request_id: 1091335698
  bitfield: 0
}
id: 3281381893
rx_time: 1770078416
rx_snr: -6.5
hop_limit: 2
priority: RESPONSE
rx_rssi: -109
hop_start: 5
relay_node: 121
```

pdxlocations:
I've created a bug report for tracking the issue I reported.  This PR may still be valuable and may or may not be related.
https://github.com/meshtastic/Meshtastic-Apple/issues/1581

adoxner:
> I'm not sure if this is the issue I was reporting on Discord. After a "Clear App Data" I now have different set of nodes on the map, but still clearly missing some. Here's a position packet from one node currently not showing on the map:
> 
> ```
> from: 363606076
> to: 1602236699
> channel: 15
> decoded {
>   portnum: POSITION_APP
>   payload: "latitude_i: 448430080 longitude_i: -1224835072 altitude: 1200 time: 1770078404 location_source: LOC_MANUAL ground_speed: 0 ground_track: 0 precision_bits: 16"
>   request_id: 1091335698
>   bitfield: 0
> }
> id: 3281381893
> rx_time: 1770078416
> rx_snr: -6.5
> hop_limit: 2
> priority: RESPONSE
> rx_rssi: -109
> hop_start: 5
> relay_node: 121
> ```

Thanks for sharing this, it's super helpful. If you look at the line I changed it seems to only show nodes when `precisionBits` are `12`, `13`, `14`, `15`, or `32` (and now `0` with this PR). Your example is just one bit outside the threshold at `16` :) 

If we match the filtering logic on Android (i.e. do not filter anything based on accuracy) it seems like it would fix this altogether. Though I think some filtering based on accuracy is probably warranted. I added a PR that allows all values of `precisionBits` greater than or equal to 12, which _should_ fix the problem in your case @pdxlocations . I updated the PR description as well. Let me know if you think this will fix it for you.

garthvh:
The logic is correct as is, those imprecise values are precise enough to be located.

Deuteranomalous1:
So the intent of the current state is to not let users dox their close bust still appropriate location on the map. 

And the solution for users who want to appear on the map is to dox their precise location? Brilliant.

PR #1625, opened 2026-03-13
feat: persist node list filters

bruschill:
## What changed?
- persist node list filters to `@AppStorage`

## Why did it change?
I wanted my filter selections to persist across app sessions.

## How is this tested?
1. Open app
1. Go to 'Nodes' tab
1. Click filters button
1. Make filter selections
1. Close App
1. Reopen App
1. Go to 'Nodes' tab
1. Click filters button
1. See that filter selections were persisted

n/a

## Checklist

- [X] My code adheres to the project's coding and style guidelines.
- [X] I have conducted a self-review of my code.
- [X] I have commented my code, particularly in complex areas.
- [X] I have verified whether these changes require an update to existing documentation or if new documentation is needed, and created an issue in the [docs repo](http://github.com/meshtastic/meshtastic/issues) if applicable.
- [X] I have tested the change to ensure that it works as intended.


garthvh (changes requested):
Thanks for the PR — the feature idea is solid and the `@AppStorage` approach is the right tool for persisting scalar filter values. However there are a few things that need to be addressed before this can merge:

## 1. Needs rebase onto `main`

`NodeFilterParameters.swift` has grown significantly since this PR was filed. `main` now includes:
- `isFiltering` computed property
- `matches(_ node: NodeInfoEntity)` — full in-memory filter logic (online, distance, PKI, role, hops, via LoRa/MQTT, ignored, environment)
- Distance calculation helpers

The PR diff does not include these additions, so this will have **merge conflicts** against current `main`. Please rebase.

## 2. `matches()` has no test coverage

The new `matches(_ node: NodeInfoEntity)` method on `main` is the most complex part of this class and has zero test coverage in this PR. Since `NodeInfoEntity` is a SwiftData `@Model`, tests for `matches()` require an **in-memory `ModelContainer`** — not Core Data (Core Data has been removed from this project entirely). Example setup:

```swift
let container = try ModelContainer(
    for: NodeInfoEntity.self,
    configurations: ModelConfiguration(isStoredInMemoryOnly: true)
)
let context = container.mainContext
```

Please add `matches()` tests using this pattern covering at minimum: online filter, role filter, hops filter, distance filter, PKI filter, ignored filter, and environment filter.

## 3. Tests must use Swift Testing, not XCTest

Per project conventions, **all new test files must use Swift Testing** — not XCTest. Please convert `NodeFilterParametersTests.swift`:

```swift
// Instead of:
import XCTest
class NodeFilterParametersTests: XCTestCase {
    func testFoo() async throws { XCTAssertEqual(...) }
}

// Use:
import Testing
@Suite("NodeFilterParameters")
struct NodeFilterParametersTests {
    @Test func foo() async throws { #expect(...) }
}
```

See `MeshtasticTests/RouterTests.swift` for an existing example in this project.

## 4. Minor: `clearAllFilterDefaults()` leaks into `UserDefaults.standard`

Tests should use a dedicated `UserDefaults` suite (e.g. `UserDefaults(suiteName: "test")`) rather than clearing `UserDefaults.standard` directly — the latter can affect other tests running in parallel and is fragile. Consider injecting the `UserDefaults` store into `NodeFilterParameters` as an init parameter (defaulting to `.standard`) so tests can pass an isolated suite.

garthvh (changes requested):
Thanks for this, @bruschill — persisting filter selections is a genuinely nice quality-of-life win, and on busy meshes it also helps perf (users who narrow the list keep it narrowed instead of re-rendering the full node set every cold start). The injected-`UserDefaults` `init(store:)` for test isolation and the dedicated test file are great touches.

First, an apology: a couple of the issues below are **our fault, not yours**. When you opened this in March, `NodeFilterParameters` had far fewer consumers. Since then the SwiftData conversion (#1668, May 3) and map filtering (#1814, May 16) both merged and now use this same object in **three** places — `NodeList`, `MeshMap`, and `UserList`. That changed the ground under the PR, so the cross-screen behavior below isn't something you could have designed around originally. Sorry for the moving target.

Requesting changes on two correctness points:

### 1. `@AppStorage` inside an `ObservableObject` doesn't publish changes
`@AppStorage` only drives SwiftUI updates when it's declared in a `View` — inside an `ObservableObject` it reads/writes `UserDefaults` but does **not** trigger `objectWillChange`. So views observing `filters` via `@StateObject`/`@ObservedObject` won't be notified when a filter toggles. Impact varies by consumer:

- **MeshMap** filters once **per render** (`filteredPositions`, `MeshMap.swift:73`) → toggling a filter no longer re-renders, so **map pins won't update live**.
- **UserList** filters per-render (`visibleUsers`, `UserList.swift:93`) → same regression.
- **NodeList** happens to mask it — it re-reads filters in its 350 ms `.task` poll (`NodeList.swift:324`), so it still works with a slight lag.

It's also inconsistent: `deviceRoles` keeps `@Published` (still publishes), while the rest don't — so on the map, role filtering would update live but online/favorite/distance wouldn't. The current test plan (set → relaunch → verify persisted) wouldn't surface this since it doesn't exercise live toggling.

**Suggested fix:** keep the fields `@Published` (so observers still update) and persist separately — mirror to `UserDefaults` in a `didSet`, the same way `deviceRoles` already does. That restores reactivity everywhere and makes all fields consistent.

### 2. All three contexts now share one persisted key space (ours to account for)
Because `NodeList`, `MeshMap`, and `UserList` each create their own `NodeFilterParameters()` but the persisted keys are global (`nodeFilter.*`), a filter set on the Nodes tab will also apply to the Map pins and the Messages contact list (at least on relaunch / re-init). Since we're the ones who fanned this out to three places, we should decide the intended behavior:
- If the three should filter **independently**, namespace the keys per context (e.g. pass a key prefix into `init`, like the `store` you already inject).
- If a single app-wide filter **is** intended, that should be a deliberate choice with one shared instance rather than three diverging `@StateObject`s.

### Smaller notes
- Consider **not** persisting `searchText` — relaunching into a stale search that hides most nodes is confusing (and it's currently shared across all three screens too).
- Persisted-but-hidden filters can read as "missing nodes." MeshMap shows a filled filter icon when active (`MeshMap.swift:306`); worth mirroring that + an easy reset on NodeList/UserList.

Happy to pair on #1 or push a small commit with the `@Published` + `didSet`-persist pattern and per-context keys if that's easier — just say the word. Thanks again for the contribution and for your patience while main shifted underneath it.

garthvh (approved):

PR #1628, opened 2026-03-17
fix/wind-speed-shows-wrong-unit

### LINKED ISSUE #1617: 🐞 [Bug]: Wind speed shows wrong unit
### Before submitting

### Firmware Version

2.7.15

### What did you do?

Wind speed on the display badge shows in km/h with the correct value. On the environment log table it seems to display in m/s judging from the value being 1/3.6 of the badge value, but labeled as km/h. Same for the chart. 

### Expected Behavior

Show value matching to unit.

### Current Behavior

Value does not match unit on table and chart.

### Participation

### Additional comments

_No response_

### Code of Conduct

bruschill:
## What changed?
Now conditionally choosing units for windspeed based on initial value, based on my assumption that the persisted/fetched value is m/s.

## Why did it change?
Assuming the unit of the persisted value is m/s, the display would show km/s regardless w/o conversion. Fixes #1617. 

## How is this tested?
[Written tests.](https://github.com/bruschill/Meshtastic-Apple/blob/75e8b15332ac8796f092669495a56e8bb9b69e98/Meshtastic/Views/Nodes/Helpers/Metrics%20Columns/WindSpeedColumnTests.swift)

n/a

## Checklist

### DIFF
--- a/Meshtastic.xcodeproj/project.pbxproj
+++ b/Meshtastic.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		0529232A2F68DB9C00930463 /* WindSpeedColumnTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 052923282F68DB7E00930463 /* WindSpeedColumnTests.swift */; };
 		102B5EAB2E172F41003D191E /* DatadogCore in Frameworks */ = {isa = PBXBuildFile; productRef = 102B5EAA2E172F41003D191E /* DatadogCore */; };
 		102B5EAD2E172F41003D191E /* DatadogCrashReporting in Frameworks */ = {isa = PBXBuildFile; productRef = 102B5EAC2E172F41003D191E /* DatadogCrashReporting */; };
 		102B5EAF2E172F41003D191E /* DatadogLogs in Frameworks */ = {isa = PBXBuildFile; productRef = 102B5EAE2E172F41003D191E /* DatadogLogs */; };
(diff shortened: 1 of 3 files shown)


garthvh:
This needs to be fixed in the firmware if there are different values, the apps should not have to determine the unit.  It looks like the firmware values are all meters per second.

garthvh:
@copilot resolve the merge conflicts in this pull request

garthvh:
Thanks for the report and the PR — the underlying bug was real. However, \`main\` already has the correct fix in place.

The protobuf definition is authoritative and unambiguous:

```swift
// telemetry.pb.swift
/// Wind speed in m/s
public var windSpeed: Float
```

Both display paths in \`main\` now unconditionally construct a \`Measurement\` with \`UnitSpeed.metersPerSecond\` and let the system format it into the user's preferred locale units:

**\`Meshtastic/Views/Nodes/Helpers/NodeDetail.swift\`** (line 322):
```swift
  let windSpeedMeasurement = Measurement(value: Double(windSpeed), unit: UnitSpeed.metersPerSecond)
```

**\`Meshtastic/Views/Nodes/Helpers/Metrics Columns/EnvironmentDefaultColumns.swift\`** (line 202):
```swift
  let windSpeed = Measurement(value: Double(\$0), unit: UnitSpeed.metersPerSecond)
```

This approach is preferable to conditionally guessing the unit from the value magnitude — the proto contract guarantees m/s, so no guessing is needed.

Closing as superseded. Thanks again for surfacing this! Closing as superseded by the existing fix in main — see comment above.

PR #1780, opened 2026-05-14
TAK V2 (TAK Packet SDK)

thebentern:
## What changed?

Switches the iOS TAK pipeline from the legacy V1 codec stack to the
TAKPacket-SDK V2 wire format, with a per-send V1 fallback for older firmware:

- **Codec swap.** Deletes `EXICodec.swift`, `FountainCodec.swift`, and
  `GenericCoTHandler.swift`. CoT XML compression and on-wire packing are
  now handled by the `MeshtasticTAK` Swift package
  (`MeshtasticTAK.CotXmlParser` → `TakCompressor` → TAKPacketV2). New
  `RouteDataPackageGenerator.swift` produces the ATAK data package zip.
- **V1/V2 fork on `AccessoryManager.supportsTAKv2`.** `sendCoTToMeshV2`
  ships zstd-dictionary-compressed `TAKPacketV2` on `ATAK_PLUGIN_V2 = 78`
  for firmware ≥ 2.8.0. Older firmware falls back to the original bare
  `TAKPacket` on `ATAK_PLUGIN = 72` (PLI + GeoChat only). The fork is
  per-send so a radio upgrading mid-session starts using V2 immediately.
- **Proto bindings.** Added the missing `PortNum.atakPluginV2 = 78`
  case (declaration, `init?(rawValue:)`, `rawValue`, `allCases`, and the
  SwiftProtobuf `_protobuf_nameMap` bytecode) and refreshed `atak.pb.swift`,
  `admin.pb.swift`, etc. against the upstream protobuf submodule. Wired up
  receiver dispatch in `AccessoryManager.swift` so incoming port-78 packets
  land in `handleATAKPluginV2Packet`.
- **Settings UX.** Both "TAK Server" entries (Module Configuration and the
  dedicated TAK section) now route to `TAKServerConfig`. A new
  `TAKIdentitySection` is embedded at the top of that screen, surfacing
  the firmware team/role pickers (previously only reachable via the now-
  unused `.takConfig` route) inline with the in-app server controls. The
  identity section saves via `accessoryManager.saveTAKModuleConfig` under
  an inline button that appears only when there are unsaved changes.
- **Test/cruft cleanup.** Removes pbxproj and `MeshtasticTests` references
  to the three deleted codec files so Xcode no longer fails with
  "Build input files cannot be found".
- **Localizations.** Adds new strings for the TAK Server screen and a few
  unrelated debug/device strings picked up from upstream.

## Why did it change?

[...]

PR #1781, opened 2026-05-14
Tak v2 electric boogaloo

thebentern:
## What changed?

Adds a version-gated V1/V2 TAK send/receive fork on iOS. New radios go
through the TAKPacket-SDK on port 78 with zstd dictionary compression;
older radios keep using the existing Apple V1 codec stack (EXI on port
257, bare `TAKPacket` protobuf on port 72). Also consolidates the TAK
Server settings page so firmware team/role identity and the in-app
server controls live on one screen.

- **V1/V2 fork on `AccessoryManager.supportsTAKv2`** in
  `TAKMeshtasticBridge.sendToMesh`:
  - **V2 (firmware ≥ 2.8.0):** `sendCoTToMeshV2` runs the CoT XML
    through `MeshtasticTAK.CotXmlParser` → `TakCompressor` (zstd dict)
    and ships `TAKPacketV2` on `ATAK_PLUGIN_V2 = 78`. Carries the full
    typed payload schema (shape, marker, route, casevac, emergency,
    task) — not just PLI/GeoChat.
  - **V1 (firmware ≤ 2.7.x):** unchanged from upstream. The bridge
    classifies the CoT via `GenericCoTHandler.classifySendMethod` and
    dispatches `.takPacketPLI` / `.takPacketChat` → bare
    `MeshtasticProtobufs.TAKPacket` on `ATAK_PLUGIN = 72`, or
    `.exiDirect` / `.exiFountain` → zlib-compressed CoT XML on
    `ATAK_FORWARDER = 257`, LT-coded into multi-packet transfers when
    larger than the LoRa MTU.
  - The fork is per-send, so a radio upgrading mid-session jumps to
    V2 on its next packet.
- **`PortNum.atakPluginV2 = 78` proto binding.** Hand-patched
  `portnums.pb.swift` — case declaration, `init?(rawValue:)`,
  `rawValue`, `allCases`, and the SwiftProtobuf `_protobuf_nameMap`
  bytecode (added `\u{1}ATAK_PLUGIN_V2\0` and adjusted the next delta
  byte from `\u{2}#` to `\u{2}\"` so `GROUPALARM_APP` still resolves
  at 112). Refreshed `atak.pb.swift`, `admin.pb.swift`, `mesh.pb.swift`,
  `config.pb.swift`, `telemetry.pb.swift`, and new `serial_hal.pb.swift`
  against the upstream protobuf submodule.
- **Receiver dispatch.** `AccessoryManager.swift`'s portnum switch
  now routes:
  - port 72 → `handleATAKPluginPacket` (unchanged V1 PLI/GeoChat path)
  - port 78 → `handleATAKPluginV2Packet` (zstd decompress → forward
    raw CoT XML to TAK clients, plus side-effects for `b-m-r` routes
    that iTAK can't stream — they're written as a KML data package zip
    to `Documents/TAK Routes/` with a local notification)
  - port 257 → `handleATAKForwarderPacket` (existing V1 EXI/Fountain
    reassembly via `GenericCoTHandler`)
- **TAK Server settings UX.** Both "TAK Server" entries (Module
  Configuration and the dedicated TAK section) now route to
  `TAKServerConfig`. A new `TAKIdentitySection` sits at the top of
  that screen with the firmware team/role pickers (previously only
  reachable via `TAKModuleConfig`) and an inline "Save TAK Identity"
  button that only appears when there are unsaved changes. The section
  saves via `accessoryManager.saveTAKModuleConfig` and reports the
  specific failure (local node info / local user record / target user
  record) so the user knows whether to reconnect or wait.
- **Bridge plumbing.** `TAKServerManager.ensureBridge()` is lazy now,
  there's an offline outgoing-mesh queue with TTL/cap that flushes on
  the first TAK client connect, `TAKConnection` gets a pre-XML pong
  fast-path and a `sendRawXML` helper, and `CoTMessage.toXML()` drops
  the `<?xml ?>` prologue so it can be safely interleaved on a TAK TCP
  stream (mid-stream declarations tear down the connection).
- **`RouteDataPackageGenerator`.** New `Helpers/TAK/` file that
  converts a `b-m-r` route CoT into a KML-inside-a-zip ATAK data
  package, saved to `Documents/` (`UIFileSharingEnabled = YES` added
  to `Info.plist` so iOS Files surfaces the directory).
- **Copilot review fixes.**
  - `handleATAKPluginV2Packet` XML prologue stripping now uses
    `^\s*<\?xml[^>]*\?>` regex instead of a literal substring, so any
    declaration form (single quotes, `standalone="yes"`, reordered
    attributes) is stripped before the XML hits the TAK TCP stream.
  - `sendCoTToMeshV2` now `throw`s `AccessoryError.ioFailed(...)` when
    the compressor refuses an oversized payload even after stripping
    `<remarks>` — previously it logged a warning and returned, which
    the bridge's `do/catch` treated as a successful send.
  - `TAKIdentitySection.saveIdentity` splits a single confusing
    "Not connected to a node." error into three specific messages so
    the user knows which lookup failed.
- **Localizations.** Adds new strings for the TAK Server screen
  (TAK Identity, Save TAK Identity, Saving…) plus a few unrelated
  debug / device strings picked up from upstream.

## Why did it change?

Without `ATAK_PLUGIN_V2` support, iOS couldn't interop with the V2
wire format that firmware ≥ 2.8.0 and Android-on-tak_v2 are already
shipping. V2 uses zstd dictionary compression keyed on TAKPacket-SDK
dictionaries — dramatic median ratio gains over the V1 path and, more
importantly, full typed payload coverage (shapes, routes, markers,
casevac, emergency) that the V1 bare-protobuf schema can't represent.

V1 stays exactly as it was. Older firmware only understands ports 72
and 257, so swapping their codec out from under them would silently
drop everything except PLI/GeoChat. Keeping the V1 EXI/Fountain stack
intact means iOS-to-iOS-on-old-firmware still gets the full Apple-only
generic CoT pipeline it had before the SDK landed.

The combined TAK Server settings page is a UX fix: previously the
"TAK Server" nav labels routed to the firmware module config page
(team/role only), so users saw the identity controls but couldn't
reach the actual in-app TAK Server screen (Enable, channel, mTLS
certs, data package export) that the screen title promised.

## How is this tested?

- Paired two TAK-role nodes on firmware 2.8.0.3a0c08b and confirmed:
  - PLIs and GeoChat round-trip via V2 in both directions.
  - Outgoing packets ship with `portnum=ATAK_PLUGIN_V2` (port 78) and
    the `28 b5 2f fd` zstd magic header on the wire (verified via
    Android debug logs on the receiving node).
  - Compressed wire payloads land in the 60-100 B range for typical
    PLIs.
- Spoofed `supportsTAKv2 = false` and confirmed the V1 branch in
  `TAKMeshtasticBridge.sendToMesh` classifies via
  `GenericCoTHandler.classifySendMethod` and dispatches PLI/GeoChat
  through `sendTAKPacket` on port 72 and other CoT types through
  `GenericCoTHandler.sendGenericCoT` on port 257 (no regression on
  the legacy stack).
- Built `MeshtasticProtobufs` standalone (`swift build`) to validate
  the hand-patched `portnums.pb.swift` (bytecode name map + enum
  case + `init?(rawValue:)`).
- Opened the combined TAK Server settings page and confirmed team /
  role pickers populate from `node.takConfig`, the Save button only
  appears on change, saving fires `saveTAKModuleConfig` end-to-end,
  and the new specific error messages surface when guards fail.
- `plutil -lint Meshtastic.xcodeproj/project.pbxproj` clean after
  pbxproj edits (file references for the V1 stack and the new
  `TAKServerConfig` / `RouteDataPackageGenerator` / `TAKIdentitySection`
  additions).

_To add: TAK Server settings page showing the new TAK Identity
section embedded above the server status / Enable toggle / channel
picker._

## Checklist


niccellular:
1. 
 #1780 and #1781 directly contradict each other, and #1780 is already merged.
 
#1780 (merged) deleted EXICodec.swift, FountainCodec.swift, GenericCoTHandler.swift. 

#1781 depends on all three still existing its V1 branch calls GenericCoTHandler.shared.classifySendMethod, sendGenericCoT, etc. As it stands, #1781 either won't compile against current main or will silently re-resurrect the deleted files. Someone has to decide which approach is canonical (delete V1 vs. keep V1) before #1781 can land at all.

  2. 
broadcastRawXml mutates connections while iterating over it.
  for (connectionId, connection) in connections {
      ...
      await removeConnection(connectionId)   // mutates `connections` mid-iteration
  }   

This is undefined-behavior bug on the new V2 receive path. Fix: collect failed IDs, remove them after the loop.

thebentern:
> #1780 and #1781 directly contradict each other, and #1780 is already merged.
> 
> #1780 (merged) deleted EXICodec.swift, FountainCodec.swift, GenericCoTHandler.swift.
> 
> #1781 depends on all three still existing its V1 branch calls GenericCoTHandler.shared.classifySendMethod, sendGenericCoT, etc. As it stands, #1781 either won't compile against current main or will silently re-resurrect the deleted files. Someone has to decide which approach is canonical (delete V1 vs. keep V1) before #1781 can land at all.
> 
> 
> broadcastRawXml mutates connections while iterating over it. for (connectionId, connection) in connections { ... await removeConnection(connectionId) // mutates `connections` mid-iteration }
> 
> This is undefined-behavior bug on the new V2 receive path. Fix: collect failed IDs, remove them after the loop.

#1780 was inadvertently merged and then reverted

I'll take a look at #2

niccellular:
1. supportsTAKv2 defaults to V2 on unknown firmware. The doc comment openly says it returns true when the radio hasn't handshook yet. A radio mid-handshake then gets port-78 packets that firmware ≤ 2.7
  silently drops. V1 is the safe default for unknown version. Copilot never flagged this — it's arguably the most consequential logic issue in the PR.

  2. drainOfflineQueue ordering may make it a no-op. It's called in the .connected case right after connectedClients.append(...), then calls broadcast/broadcastRawXml — but those guard on the connections
  dictionary, not connectedClients. If the new connection isn't inserted into connections until after the .connected event is processed, every drained message hits the empty-guard and gets immediately 
  re-enqueued. Worth verifying the insertion order; if it's wrong, the whole offline-queue feature silently does nothing on reconnect.

niccellular (commented):
lgtm

PR #1784, opened 2026-05-14
Fix firstHeard/lastHeard timestamp bugs and hopsAway misparse

turbosnail11:
## Summary

- **Fix `lastHeard` regression from NodeDB dump** — `nodeInfoPacket()` was unconditionally overwriting `lastHeard` with the radio's stored `nodeInfo.lastHeard` value, which could be much older than a value already set by a live packet via `updateAnyPacketFrom()`. This caused the node detail view to show "First Heard" as more recent than "Last Heard". Now only updates `lastHeard` when `nodeInfo.lastHeard` is non-zero and more recent than the currently stored value.

- **Fix `hopsAway` always showing 0 due to proto misparse** — `upsertNodeInfoPacket()` was parsing the `NODEINFO_APP` payload as `NodeInfo` and reading `hopsAway` from it. The payload is actually a `User` proto: `User` field 9 is `is_unmessagable` (bool), while `NodeInfo` field 9 is `hops_away` (uint32). Protobuf parses without error but always yields 0 for normal nodes, silently overwriting the correct value computed from `hopStart - hopLimit`. For new nodes, `hopsAway` is now computed directly from the packet hop fields. For existing nodes, the misparse assignment is removed — `updateAnyPacketFrom()` already set the correct value.

## Test plan

- [ ] Connect to a mesh with nodes multiple hops away — confirm they no longer appear in the "Direct" filter
- [ ] Disconnect and reconnect BLE — confirm "Last Heard" does not regress to an older value after reconnect
- [ ] Confirm "First Heard" is never newer than "Last Heard" in node detail view

### DIFF
--- a/Meshtastic/Helpers/MeshPackets.swift
+++ b/Meshtastic/Helpers/MeshPackets.swift
@@ -453,7 +453,12 @@ actor MeshPackets {
 					
 					fetchedNode[0].id = Int64(nodeInfo.num)
 					fetchedNode[0].num = Int64(nodeInfo.num)
-					fetchedNode[0].lastHeard = Date(timeIntervalSince1970: TimeInterval(Int64(nodeInfo.lastHeard)))
+					if nodeInfo.lastHeard > 0 {
+						let candidate = Date(timeIntervalSince1970: TimeInterval(nodeInfo.lastHeard))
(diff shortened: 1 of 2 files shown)


garthvh (approved):
Awesome, clean, minimal, well-explained fixes for two real bugs with good test instructions.

PR #1788, opened 2026-05-14
Fix: display channel names without reformatting

turbosnail11:
## Summary

- **Stop applying `camelCaseToWords()` to channel names** — Channel names like "LongFast" were being displayed as "Long Fast" throughout the UI (ChannelList, ChannelMessageList, Channels settings, ShareChannels, TAKServerConfig). This is misleading because the actual name — used for MQTT topic matching and over-the-air channel identification — is the unformatted version. Adds `displayName` and `shortDisplayName` computed properties to `ChannelEntity` that return the name as-is, falling back to "Primary Channel" or "Channel N" for unnamed channels.

## Future Improvement Suggestion

New users don't understand the modem preset (LongFast, LongSlow, etc.) linking with the channel name and think they can rename the public channel — the moment someone customizes a channel name or switches presets it kills their public broadcasts. It may be worth surfacing the modem preset as the primary concept in LoRa settings, with the raw channel name field moved to an advanced section within LoRa to prevent unintended changes. This would leave only the secondary channels as editable in the channel settings and would allow the display name of the public channel to be separated from the modem preset (i.e. it would always show as "Public").

## Test plan

- [ ] Open Messages → channel list — confirm "LongFast" appears instead of "Long Fast"
- [ ] Open Settings → Channels — confirm channel names are unformatted
- [ ] Open Share Channels sheet — confirm channel names are unformatted
- [ ] Confirm unnamed primary channel still shows "Primary Channel" / "Primary"
- [ ] Confirm unnamed secondary channels show "Channel 1", "Channel 2", etc.

### DIFF
--- a/Meshtastic/Extensions/SwiftData/ChannelEntityExtension.swift
+++ b/Meshtastic/Extensions/SwiftData/ChannelEntityExtension.swift
@@ -9,6 +9,30 @@ import SwiftData
 import MeshtasticProtobufs
 
 extension ChannelEntity {
+	var displayName: String {
+		if let name, !name.isEmpty {
+			return name
+		}
+
+		if index == 0 && role == 1 {
+			return "Primary Channel"
+		}
+
+		return "Channel \(index)"
+	}
+
+	var shortDisplayName: String {
+		if let name, !name.isEmpty {
+			return name
+		}
+
+		if index == 0 {
+			return "Primary"
+		}
+
+		return "Channel \(index)"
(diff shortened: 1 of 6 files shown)


turbosnail11:
I understand the space might be a design decision so feel free to discard.  I just found it very confusing and thought it was the reason I wasn't getting messages and couldn't figure out why the name wasn't updating (without the space) when I changed it.

PR #1789, opened 2026-05-14
Fix: center map on user location at 100km on open

turbosnail11:
## Summary

- **Map no longer zooms out to fit all annotations on open** — The map was initialized with `MapCameraPosition.automatic`, which causes MapKit to expand the view to include all visible nodes — showing the whole country on a spread-out mesh. The initial position is now `userLocation`, which centers on the user without fitting annotations. On appear, the camera is explicitly set to 100km centered on the user's location. If the nearest node is farther than 100km, the camera scales out just enough to include it (capped at 1,000km). If user location isn't available yet, the map centers on the nearest known node instead. Existing users with the old 800km default stored in UserDefaults are migrated automatically on first launch.

## Test plan

- [ ] Open map with a spread-out mesh — confirm it no longer zooms to fit all nodes
- [ ] Fresh install — confirm map opens centered on user at ~100km
- [ ] Nearest node farther than 100km — confirm camera scales out to include it, not all nodes
- [ ] No location permission granted — confirm map centers on nearest node instead of fitting all
- [ ] Existing install with 800km stored in UserDefaults — confirm migrates to 100km on first open
- [ ] Switch away from map tab and back — confirm camera does not reset on return

### DIFF
--- a/Meshtastic/Enums/AppSettingsEnums.swift
+++ b/Meshtastic/Enums/AppSettingsEnums.swift
@@ -56,6 +56,7 @@ enum MeshMapDistances: Double, CaseIterable, Identifiable {
 	case tenMiles = 16093.4
 	case twentyFiveMiles = 40233.6
 	case fiftyMiles = 80467.2
+	case oneHundredKilometers = 100000
 	case oneHundredMiles = 160934
 	case twoHundredMiles = 321869
 	case fiveHundredMiles = 804672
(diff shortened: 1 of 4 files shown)


garthvh:
You can decrease the distance you want to display, this is not a user focused map page

TurboClawd:
> You can decrease the distance you want to display, this is not a user focused map page

Hmmm.  I use it to find the nodes around me, how is it meant to be used?  Doesn't seem helpful to see the whole country.

PR #1795, opened 2026-05-15
Fix translation availability log spam

garthvh:
"Language fr supported but not installed" was logged ~30 times every time the doc browser loaded (once per UI string translation call). 

Moved all availability status messages inside the existing dedup check (`lastAvailabilityStatusByLanguage`) so each status is logged only once per language per app session.

Before: ~30 identical log lines on browser load
After: 1 log line per language


garthvh:
@copilot resolve the merge conflicts in this pull request

PR #1894, opened 2026-05-29
Optimize large node and message views

RCGV1:
## Summary

This PR optimizes the app for very large node databases and long message histories. The main goal is to reduce repeated SwiftData relationship materialization, broad queries, repeated sorts, per-row body work, and heavy MapKit annotation rendering while preserving current behavior.

Key changes:
- Added a DEBUG-only seeded performance harness for thousands of nodes, telemetry rows, positions, and messages. Production startup remains unchanged unless the debug perf argument/env flags are used.
- Reworked node list filtering and row rendering to use precomputed parameters, batched lookup data, compact row summaries, and fewer repeated sorts/queries.
- Reworked direct/channel message lists to page results, poll with stable change tokens, precompute visible reply/tapback context, and avoid full-list scans from row bodies.
- Reduced ingest-time pruning overhead by amortizing expensive count/sort/delete work for positions, telemetry, pax counters, and messages.
- Optimized maps with pure-value position snapshots, visible-region/density limiting, deterministic spatial sampling, and lightweight tappable dense annotations. This does not introduce clustering.
- Converted node detail and log screens to bounded state-backed fetches/counts where broad relationship queries were doing unnecessary work.
- Preserved export behavior for logs that need all rows, including PAX and position exports.

## Why

With thousands of nodes, the app was repeatedly loading large SwiftData relationships and rebuilding sorted/filtered collections from SwiftUI body updates. Long message lists had similar issues: full message arrays, per-row reply/tapback scans, broad unread queries, and notification cancellation one item at a time. Map views also became expensive because every visible update could create heavy annotations for a dense node set.

This moves the hot paths toward bounded fetches, paged data, batched lookup dictionaries, and cheap value snapshots so list/map interactions scale with what is visible instead of the full database wherever practical.

## Performance validation

Validated locally on the node-scale simulator profile:
- `git diff --cached --check` passed.
- `xcodebuildmcp build_sim -quiet` succeeded for `Meshtastic` Debug on iPhone 17 Pro simulator.
- Seeded DEBUG harness launched with `MESHTASTIC_PERF_SEED_NODES=5000`, telemetry history `3`, position history `3`, reset store enabled, and compact node list enabled.
- Seeded launch displayed `Nodes (4948)` and runtime log tail had no crash/fatal/error signatures.
- 10 second idle sample on the seeded optimized app: average CPU `1.34%`, max CPU `1.80%`, average RSS `630378 KB`.

## Safety notes

- The seed harness is DEBUG-only and gated by explicit launch argument/env flags.
- Dense map mode still uses tappable annotations; no clustering was added.
- Generated/resource diffs from local build/image sync were intentionally excluded from this PR.
- Existing unrelated build warnings remain: SwiftLint is not installed locally, generated CoreData Swift resource warnings, and remote image sync content-type warnings for two SVGs.


garthvh:
@RCGV1 make a branch in the org that can be tested

PR #1929, opened 2026-06-12
Add LoRa coding rate preset override controls

RCGV1:
## Summary
- Adds a LoRa coding-rate control to the LoRa config screen.
- Keeps preset mode on preset bandwidth/spreading factor while allowing only firmware-valid higher coding-rate overrides.
- Uses a default toggle plus stepped slider so users can see when they are following the preset and when they are adding redundancy.
- Preserves protobuf default behavior by saving `coding_rate = 0` for preset-default CR instead of writing an explicit equal-to-default value.

## Reasoning
Firmware now allows `config.lora.coding_rate` to override a preset only when the value is a valid higher/more redundant CR. Equal or lower values should fall back to the preset CR. The UI mirrors that rule by filtering preset-mode choices to `0` or strictly higher values, while custom mode still exposes explicit `4/5...4/8` values.

## Screenshot
LoRa coding rate override slider in preset mode

## Verification
- `git diff --check` passed.
- Built for Benjamin's iPhone with:
  `xcodebuild -workspace Meshtastic.xcworkspace -scheme Meshtastic -configuration Debug -destination 'platform=iOS,id=00008130-000578880102001C' -derivedDataPath /tmp/meshtastic-device-derived -skipPackagePluginValidation -skipMacroValidation build`
- Initial device build hit a codesign internal error while disk was nearly full; after clearing this branch's derived-data caches, the same build succeeded.
- Installed the built app to the iPhone successfully. Launch was denied because the phone was locked.
- Focused test compile was attempted earlier, but the test bundle hit the same local code-signing/disk-space issue rather than a source failure.

### DIFF
--- a/Meshtastic/Enums/LoraConfigEnums.swift
+++ b/Meshtastic/Enums/LoraConfigEnums.swift
@@ -587,7 +587,7 @@ enum CodingRates {
 			return Array(validRange)
 		}
 		let defaultCodingRate = modemPreset?.defaultCodingRate ?? ModemPresets.longFast.defaultCodingRate
-		return [0] + validRange.filter { $0 >= defaultCodingRate }
+		return [0] + validRange.filter { $0 > defaultCodingRate }
 	}
 
 	static func normalized(_ codingRate: Int, usePreset: Bool, modemPreset: ModemPresets?) -> Int {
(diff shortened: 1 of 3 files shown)


garthvh:
I don't really like this pattern, once you go to custom stuff you are customizing things.

RCGV1:
I disagree, modifying your coding rate isn't "custom" it doesn't make your packets get ignored by other nodes running different coding rates. I want to encourage people to use different coding rates, as depending on your deployment, you may need a higher coding rate to get out. @NomDeTom can back me up here

PR #1930, opened 2026-06-13
Generate RF coverage overlays from nodes

RCGV1:
## Summary
- Port coverage generation for the default `https://site.meshtastic.org` path to the native on-device Site Planner engine described in the mobile design doc: https://github.com/meshtastic/meshtastic-site-planner/blob/main/docs/on-device-mobile.md
- Add `Tools > RF Site Planner`, a full-screen native MapKit planning page for ad hoc RF work without needing to start from an existing node detail screen.
- Polish the Tools entry so disconnected users see a clean native RF Planning row instead of an empty NFC section, and make the documented `meshtastic:///settings/tools` deep link land on the Tools screen.
- Coverage mode: tap any map point, generate an on-device RF coverage overlay for that point, annotate the RF settings into the GeoJSON, import it through the existing Map Data pipeline, and render it immediately on the planner map.
- Point-to-point mode: tap transmitter/receiver points, run the native ITM point-to-point path model, draw the path as green/orange/red by result, mark blocking terrain when found, and show loss, predicted signal, link margin, LOS/Fresnel clearance, and an elevation profile chart.
- Promote point-to-point analysis results into a draggable on-map profile panel with the terrain/LOS/Fresnel diagram, while keeping the detailed Link Profile sheet available from the panel.
- Refine the planner UI to use native MapKit `Marker`s, hide the app tab bar while planning, keep the map chrome clear of top-right app toolbar collisions, restore the native MapKit user-location button below the status area, and place Map Data/RF Settings actions in the bottom SwiftUI `.bar` footer.
- Add planner controls for high detail, coverage range, overlay opacity, color ramp, transmitter/receiver height above ground, frequency, transmit power, antenna gains, system loss, and receiver threshold.
- Add the Swift bridge declarations for the native SPLAT point-to-point API and reuse the existing terrain tile loader for both coverage and link analysis.
- Keep custom hosted Site Planner API URLs working for operators who run compatible `/predict`, `/status/{task_id}`, and `/result/{task_id}` endpoints.
- Improve contour quality with 640-cell native contour output, three-pass smoothing, and byte-limit fallback so overlays are less blocky while staying under the app's 10 MB import limit.
- Show persisted RF summary details in generated overlay success alerts, Map Options overlay rows, and Map Data management rows.
- Add a native Map Options opacity slider that globally scales all imported GeoJSON/RF overlay fills and strokes at render time without rewriting stored overlay files.

## Reasoning
The public Site Planner site is a browser app, not a hosted prediction API. The mobile design doc calls for mobile clients to run the engine on device/offline instead of relying on a shared server. This makes the app's default behavior match the documented architecture while preserving the custom hosted endpoint path for operators who run a compatible backend.

The Tools planner covers the workflow where the user wants to evaluate a prospective site, not only a node that already exists in the mesh. The map uses direct tap placement without a persistent center reticle, while the native user-location button gives users a familiar way to recenter the map.

The point-to-point result chart needs to be available during map work, not buried in a form. The draggable profile panel keeps the terrain profile visible while preserving map context and lets users move it away from the area they are inspecting.

The native path originally rendered valid coverage, but row-merged contours could look too low-poly at normal map zoom levels. The renderer now keeps more raster detail by default and smooths the field before generating RF bands. It still has an adaptive byte-budget fallback, so large overlays remain importable and fast enough for MapKit.

## Screenshots
Tools entry for the full-screen planner:

RF Site Planner Tools entry

RF Site Planner map with the native location control lowered below the status area and bottom planner controls:

RF Site Planner map

RF settings sheet with coverage detail, opacity, height, and RF controls:

RF Site Planner settings

Existing generated native Site Planner coverage overlay on MapKit:

Native Site Planner coverage overlay

## Verification
- `git diff --check` -> OK.
- `xcodebuild` simulator build through XcodeBuildMCP `build_run_sim` with `-disableAutomaticPackageResolution -skipPackagePluginValidation -skipMacroValidation CODE_SIGNING_ALLOWED=NO`, launching `gvh.MeshtasticClient --meshtastic-perf-seed` on simulator `DE3343B3-1E9D-4C2D-9AFE-B50AC465BEBA` -> succeeded.
- `xcodebuild` targeted router tests through XcodeBuildMCP `test_sim` with `-only-testing:MeshtasticTests/RouterURLRoutingTests -disableAutomaticPackageResolution -skipPackagePluginValidation -skipMacroValidation CODE_SIGNING_ALLOWED=NO` -> 51 passed, 0 failed.
- Opened `meshtastic:///settings/tools` on the simulator, tapped `RF Site Planner`, switched to Point to Point, and captured the updated planner screenshot above; confirmed the native MapKit tracking button is visible below the status/battery area and bottom planner controls remain reachable.
- Earlier branch verification: targeted GeoJSON/RF tests passed with `xcodebuild ... test -only-testing:MeshtasticTests/RFGeoJSONOverlayTests -only-testing:MeshtasticTests/GeoJSONFeaturePropertyTests` reporting 39 tests passed.
- Earlier planner smoke: opened `Tools > RF Site Planner`, confirmed the full MapKit planner page renders with the app tab bar hidden, the compact native footer visible, and no covered bottom controls.
- Native point-to-point engine smoke: compiled the SPLAT driver from the branch and ran a sea-level TX/RX path through `splat_point_to_point`; result returned `count=183 loss=177.421 signal=-165.281 distance=13.4305 azimuth=75.8168 err=0 profile=13.4305`.

Notes: the build still logs pre-existing duplicate source build-file warnings, SwiftData macro Sendable warnings, missing SwiftLint command, and warnings from the imported ITWOM source; none block the build, tests, install, or launch.


garthvh:
@RCGV1 needs a branch in the org

RCGV1:
It's not ready yet

garthvh:
Merged in sync with feature on android

PR #1940, opened 2026-06-15
feat: export whole device configuration (#1924)

### LINKED ISSUE #1924: 🚀 [Feature Request]: export whole configuration
### OS

iOS

### Description

Android app lets users export full device config (this is also possible from CLI). Please add that feature to iOS/macOS APP

### Participation

### Additional comments

_No response_

### Code of Conduct

bruschill:
## What changed?

Adds an **Export Configuration** action to **Settings → Tools** that saves the connected node's full configuration to a binary `DeviceProfile` (`.cfg`) file.

- `DeviceProfileExport.swift` — entity → protobuf converters for all 8 device configs and 15 module configs (the inverse of the proto → entity mapping in `UpdateSwiftData.swift`), plus `NodeInfoEntity.exportDeviceProfile()` which assembles a `DeviceProfile`: long/short name, `LocalConfig`, `LocalModuleConfig`, the channel-set URL, ringtone, canned messages, and the fixed position when one is set.
- `DeviceProfileDocument.swift` — a `FileDocument` wrapping the serialized profile bytes for `.fileExporter`.
- `Tools.swift` — the export button + `fileExporter`; shown with guidance/disabled when no node is connected.
- `UpdateSwiftData.swift` — persists MQTT `mapReportingShouldReportLocation` in the config-update branch (it was only set on insert), so the exported value isn't a stale default.

Closes #1924.

## Why did it change?

The Android app and the `meshtastic` CLI can export a device's full configuration; iOS/macOS could not. The `.cfg` binary `DeviceProfile` format matches Android, so an exported file can be backed up and re-imported across the Meshtastic ecosystem.

## How is this tested?

- New `DeviceProfileExportTests` (Swift Testing) — 16 tests covering the converters (enum mapping, negated-flag inversion, packed IPv4 bit-pattern, admin-key slot preservation, nested map-report settings, `isServer`, TAK/traffic/status) and `DeviceProfile` assembly including a binary serialization round-trip. All pass.
- Full app build succeeds for the iOS simulator (Debug).
- The converter set was reviewed field-by-field against the `upsert*ConfigPacket` functions in `UpdateSwiftData.swift` (the import mapping) to keep export/import symmetric.

## Checklist

- [x] I have verified whether these changes require updates to the in-app documentation under `docs/user/` or `docs/developer/`. A short note in the Tools/settings docs may be warranted; happy to add it here or apply the **`skip-docs-check`** label per maintainer preference.

---

### Notes for reviewers

- **Sensitive material is exported in plaintext.** The profile includes the private key, admin keys, WiFi PSK, and MQTT password. This matches the Android `.cfg` behavior (the format is meant to fully restore a device), but the exported file should be treated as secret. Worth a deliberate decision if iOS wants to diverge (e.g. prompt or redact).
- **Export ↔ import are two hand-maintained mirrors.** Each `protoConfig` converter is the inverse of an `upsert*ConfigPacket` mapping, and the two can silently drift as config fields are added (the file header comments call this out). A shared single source of truth between import and export would be the deeper fix for a future pass.

## Summary by CodeRabbit

* **New Features**
  * Added "Export Device Configuration" to save a connected node's configuration as a local `.cfg` file with a safe, Android-style filename and correct export type.
  * Shows a confirmation step before exporting sensitive security material.
* **Bug Fixes**
  * Improved device/module export fidelity (including channel URL, RTTTL ringtones, canned messages, and fixed-position rules).
  * Fixed persistence during syncs for MQTT location reporting and refreshed store-forward router/server flag.
  * Corrected admin-key slot alignment/trimming and improved optional/empty-field handling in exports.
* **Tests**
  * Added thorough device-profile export and export-filename tests, including serialization round-trips.
* **Documentation**
  * Updated Settings deep-link documentation (supported paths and query handling).


RCGV1 (commented):
Compatibility/hardening review focused on Android `.cfg` parity:

- [P2] Add a confirmation/warning before export because this file can include security/admin material. The export path serializes the full `DeviceProfile`, including `config.security`, admin keys/private key when present, and channel PSKs via `channelURL`. Right now `Tools.swift` starts export directly from the button. Please add a warning/confirmation or selection step that makes it clear this backup can grant mesh/admin access if shared.

- [P2] Make the filename/type match Android more closely. Android currently exports names like `Meshtastic_${shortName}_${yyyyMMdd}_nodeConfig.cfg`; this PR uses the long name, no explicit `.cfg` suffix in the default filename, and a dynamic `UTType(filenameExtension: "cfg")`. Please use a sanitized Android-style filename with an explicit `.cfg` extension, and consider defining a stable exported UTType instead of recreating it inline.

- [P3] Surface file-export write failures to the user. The `.fileExporter` completion logs failures, but the visible alert only covers serialization/preparation failures before the exporter opens. If iOS fails to write/save the document, the user should see an alert instead of only a log entry.

The raw file format itself looks compatible with Android: both sides are using raw serialized `DeviceProfile` protobuf bytes, not JSON or a wrapper container.

RCGV1:
@bruschill, I think this is a great feature. Can you address the issues I raised above

tylerpieper:
Seeing this inspired me to have the hallucination machine implement this for CLI and Web as well. Now we will have one import/export format common across all platforms!

garthvh:
This looks solid — nice symmetry with the `upsert*ConfigPacket` import mapping, and good call flagging the plaintext-secrets and hand-maintained-mirror tradeoffs.

Holding off on merging until the **import** side exists too — export alone isn't very useful without a way to bring a `.cfg` back in, and I'd rather land both together than ship a half feature. Marking this as a draft for now; I'll flip it back and merge once import is available (or ready to review as a companion PR).

bruschill:
Superseded by #2023, which contains this same export work **plus** the inverse import feature so the two ship together. Closing in favor of the combined PR.

PR #1962, opened 2026-06-19
Fix list text wrapping after performance row caps

RCGV1:
## Summary
- Let channel and direct-message preview rows grow past their old 62pt fixed height when text wraps.
- Let the shared node `IconAndText` label report its vertical size instead of being visually capped by list row measurement.
- Preserve the existing performance caching and deliberate one-line compact controls.

## Reasoning
I checked the current upstream tree and recent performance history for a global `lineLimit` / text cap. There is no app-wide text limiter in the latest code; the practical regression points are fixed-height list rows and shared row text that could not claim vertical space when wrapping.

## Verification
- `git diff --check`
- `xcodebuild -workspace Meshtastic.xcworkspace -scheme Meshtastic -configuration Debug -destination 'platform=iOS Simulator,id=DE3343B3-1E9D-4C2D-9AFE-B50AC465BEBA' -derivedDataPath /tmp/meshtastic-textlimit-derived -skipPackagePluginValidation -skipMacroValidation build` succeeded.
- Installed/launched the built app on the iPhone 16e simulator with `--meshtastic-perf-seed --meshtastic-perf-reset`; seeded Nodes list rendered successfully.

### DIFF
--- a/Meshtastic/Views/Messages/ChannelList.swift
+++ b/Meshtastic/Views/Messages/ChannelList.swift
@@ -110,6 +110,7 @@ struct ChannelList: View {
 						Text(LocalizedStringKey(mostRecent?.messagePayload ?? " "))
 							.font(.footnote)
 							.foregroundColor(.onSurfaceVariant)
+							.fixedSize(horizontal: false, vertical: true)
 					}
 				}
 			}
@@ -126,7 +127,7 @@ struct ChannelList: View {
 			.alignmentGuide(.listRowSeparatorLeading) {
 				$0[.leading]
 			}
-			.frame(height: 62)
+			.frame(minHeight: 62, alignment: .center)
 			.contextMenu {
 				if hasMessages {
 					Button(role: .destructive) {
--- a/Meshtastic/Views/Messages/UserList.swift
+++ b/Meshtastic/Views/Messages/UserList.swift
@@ -294,6 +294,7 @@ private struct DirectMessageUserRow: View {
 					Text(user.longName ?? "Unknown".localized)
 						.font(.headline)
 						.allowsTightening(true)
+						.fixedSize(horizontal: false, vertical: true)
 					Spacer()
 					if user.userNode?.favorite ?? false {
 						Image(systemName: "star.fill")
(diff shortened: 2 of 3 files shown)


garthvh:
Ran this branch on an iPhone 16 simulator to look at the actual list rendering. To exercise the rows this PR touches I seeded channel messages (`MESHTASTIC_PERF_CHANNEL_MESSAGES`) with one deliberately long message as the most-recent, so it becomes the list preview. A clean A/B of the **Channel List** preview row (same message, base commit vs this PR):



**Before** (`.frame(height: 62)`) — the preview truncates to **2 lines + `…`** in a compact fixed-height row:
> 🔓 **Perf Channel**  ·  Channel perf message 0: This is a deliberately very long message used to v**…**

**After** (this PR — `minHeight: 62` + `fixedSize(vertical:)`) — the same row grows to **~7 full lines**, no ellipsis, ~2–2.5× taller:
> 🔓 **Perf Channel**  ·  Channel perf message 0: This is a deliberately very long message used to verify list row wrapping behavior. It should keep going well past one line and well past two lines … (entire message, unwrapped)

### The concern
The height cap wasn't clipping text by accident — it was doing intentional **2-line truncation with an ellipsis**, which is the standard preview-row behavior (and matches Android). Removing it doesn't "fix wrapping" so much as let a single long message expand its row unbounded and dominate the list. Because the preview is narrow (footnote font), even ordinary multi-sentence messages will now push rows to 2–3+ lines.

### Suggestions
- For `ChannelList` / `UserList` preview rows, prefer **`.lineLimit(2)`** on the preview `Text` over removing the height cap — that keeps rows bounded *and* keeps explicit truncation.
- The `IconAndText` change in `NodeListItem` is the one piece with a plausible real case (a detail label getting cut off); worth evaluating on its own, separate from the message rows.

### Minor note
With the perf seed, `directMessageCount` / `channelMessageCount` default to `0`, so `--meshtastic-perf-seed` alone seeds nodes but **no messages** — the changed rows aren't exercised unless those env vars are set. Worth passing them when verifying message-list changes.

garthvh:
@RCGV1 not sure why we would want to remove this feature.

PR #1964, opened 2026-06-19
fix: reflect message ACK delivery status live in open conversations

bruschill:
## What changed?

Message **delivery-receipt (ACK) status now updates live** in an open Direct Message or Channel conversation, instead of only appearing after the user leaves and re-enters the conversation.

- Added `AppState.messageAckUpdate`, a signal bumped whenever a delivery ACK is persisted.
- Factored a shared `MeshPackets.saveAck(appState:caller:)` helper that commits the ACK and fires the signal, used by **both** ACK-writing paths — `routingPacket` (routing ACKs) and `adminResponseAck` (admin-response ACKs) — so the live update never depends on which path delivered the receipt.
- `UserMessageList` and `ChannelMessageList` observe the signal and reload, so any ACK-field change — `receivedACK`, `realACK`, or a later `ackError`/NAK on an already-acked message — surfaces immediately.
- The ACK write is committed synchronously so the reload reads the fresh state.

## Why did it change?

Since the message lists moved from `@Query` to a manually-fetched `@State` snapshot array + 5s poll (#1901), they only reload when their change token (latest cursor + message count) moves. An incoming ACK changes neither, and ACK fields are written on `MeshPackets`' isolated `@ModelActor` context — so the on-screen rows kept their stale "sent" state until a full re-fetch happened on re-entry. This restores the live delivery indicator users expect.

## How is this tested?

- Added `AckCrossContextRefreshTests`, which drives the **real** `routingPacket()` path on a separate `ModelContext` (exactly as the production `@ModelActor` does) and asserts that a fresh fetch on the view's **main context** — what `loadMessages()` does — sees the committed `receivedACK`/`realACK`/`ackError`. This proves the cross-context refresh the fix depends on, and runs in CI.
- `xcodebuild test` passes on the iOS 18 simulator; full app build succeeds.
- Note: not yet exercised against a live LoRa ACK round-trip (requires two physical nodes), but the data-layer mechanism it relies on is verified.

N/A — delivery-indicator timing change; covered by the automated test above.

## Checklist

- [ ] I have verified whether these changes require updates to the in-app documentation under `docs/user/` or `docs/developer/`, and updated accordingly. No user-facing view → doc mapping changes; requesting the **`skip-docs-check`** label.

### DIFF
--- a/Meshtastic.xcodeproj/project.pbxproj
+++ b/Meshtastic.xcodeproj/project.pbxproj
@@ -275,6 +275,7 @@
 		DD00001BINTCVRTST0000001 /* IntentConverterTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD00001BINTCVRTST0000000 /* IntentConverterTests.swift */; };
 		DD00001CMAPDATTST0000001 /* MapDataModelTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD00001CMAPDATTST0000000 /* MapDataModelTests.swift */; };
 		DD00001DCDENTTST00000001 /* EntityTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD00001DCDENTTST00000000 /* EntityTests.swift */; };
+		DD000031ACKXCTXTST000001 /* AckCrossContextRefreshTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD000031ACKXCTXTST000000 /* AckCrossContextRefreshTests.swift */; };
 		DD00001EURLDETTST0000001 /* URLExtensionDetailedTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD00001EURLDETTST0000000 /* URLExtensionDetailedTests.swift */; };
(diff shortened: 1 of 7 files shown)


garthvh:
Closing as **can't reproduce** — delivery-ACK status already updates live in open conversations in current builds.

The rows read the ACK fields directly off the message object (`ChannelMessageRow`/`UserMessageRow` use `message.receivedACK` / `realACK` / `ackError` / `canRetry`), and `MessageEntity` is an `@Observable` `@Model`. The `@State messages` array holds references to the main-context objects, so when an ACK is persisted (even from MeshPackets' `@ModelActor` context, which the main `ModelContext` merges), the on-screen row re-renders from per-object observation. That path is independent of the snapshot change-token + 5s poll that #1901 introduced — that token only governs adding/removing messages, not field updates on already-visible ones. This is also why it worked under `@Query` before, and it still works now.

The premise that the token doesn't move on an ACK is correct, but the conclusion that the indicator goes stale until re-entry doesn't hold, because the ACK updates the same object the row is already observing.

`AckCrossContextRefreshTests` asserts that a *fresh fetch* sees the committed ACK — it proves persistence + re-fetch, but doesn't demonstrate the live on-screen object going stale, so it doesn't reproduce the reported bug.

If you can capture a concrete repro (a clip of an ACK that does *not* update until you leave and re-enter the conversation), please reopen with it and we'll dig in. Thanks for the thorough write-up and the test either way.

PR #1965, opened 2026-06-19
fix: correct IPv4 byte order in Network Config (display + write)

bruschill:
## What changed?
Corrected the IPv4 ⇄ `UInt32` byte order in the static-IP helpers in `NetworkConfig.swift`. Both `ipStringToUInt32` and `uint32ToIpString` now use **little-endian** octet order (first octet = least-significant byte) instead of big-endian, and a Swift Testing suite (`NetworkConfigIPConversionTests`) was added to pin the byte order.

## Why did it change?
The helpers packed octets big-endian, but the firmware (Arduino `IPAddress`) and the Meshtastic Android app store them little-endian. This caused two problems:

- **Display:** static/DHCP IPv4 addresses read back from the device rendered byte-reversed in the iOS/macOS app (e.g. `192.168.1.1` shown as `1.1.168.192`).
- **Write:** a static IP typed into the app would have been written to the device byte-reversed.

The little-endian convention is confirmed by the Android reference implementation:

```kotlin
// formatIpAddress — first octet from the LOW byte
"${(ipAddress) and 0xFF}.${(ipAddress shr 8) and 0xFF}.${(ipAddress shr 16) and 0xFF}.${(ipAddress shr 24) and 0xFF}"

// convertIpAddressToInt
.split(".").map { it.toIntOrNull() }
.reversed() // little-endian byte order
.fold(0) { total, next -> total shl 8 or next }
```

No data migration is required: stored values are the firmware's raw `uint32` (kept verbatim in `UpdateSwiftData.swift`); only the display/encode order changed, so existing addresses now render correctly.

## How is this tested?
Added `MeshtasticTests/NetworkConfigIPConversionTests.swift` (Swift Testing):

- Known-value mapping in both directions (`192.168.1.1 ↔ 0x0101A8C0`)
- Asymmetric case (`10.20.30.40`) that catches any accidental byte swap a palindrome wouldn't
- 7 parameterized round-trips + subnet-mask check
- Edge cases: `0 → ""`, empty/`0.0.0.0 → 0`, and 5 malformed inputs → `0`

Suite passes on the iOS 18 simulator (`xcodebuild test -only-testing:MeshtasticTests/NetworkConfigIPConversionTests`).

> Note / follow-up: `ipStringToUInt32` still returns `0` for malformed input, so a typo in a static field saves as `0.0.0.0` with no validation feedback. This is pre-existing and out of scope for this byte-order fix; tracking input validation separately.

N/A — addressing values are not easily shown without a networked device; covered by unit tests instead.

## Checklist

- [x] I have verified whether these changes require updates to the in-app documentation under `docs/user/` or `docs/developer/`, and updated accordingly. No user-facing doc change needed (bugfix to existing field behavior) — please apply the **`skip-docs-check`** label.

### DIFF
--- a/Meshtastic.xcodeproj/project.pbxproj
+++ b/Meshtastic.xcodeproj/project.pbxproj
@@ -228,6 +228,7 @@
 		BD7B1A838F8245A0B6BE56C5 /* WatchSessionManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = 84CCF00CDD3D48C9B9DCA554 /* WatchSessionManager.swift */; };
 		C1B995B38DC97D13D90D9C1A /* DiscoverySummaryView.swift in Sources */ = {isa = PBXBuildFile; fileRef = F1B6FB13C4D4E03503BB1DF8 /* DiscoverySummaryView.swift */; };
 		C54AC9F4DA920A4F718CE21A /* DiscoveryScanEngineTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = AAD14FA280F0A130730DEB93 /* DiscoveryScanEngineTests.swift */; };
+		FDAF8808342245C4817C93EC /* NetworkConfigIPConversionTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 18295AFE686247778632C6DE /* NetworkConfigIPConversionTests.swift */; };
 		C79BFC99EDCC89DD45082753 /* DiscoveryScanView.swift in Sources */ = {isa = PBXBuildFile; fileRef = F5CA8BAC814CF82E127EEABF /* DiscoveryScanView.swift */; };
 		D02589957BD040969FCB8663 /* PositionEntity.swift in Sources */ = {isa = PBXBuildFile; fileRef = A97241903A924144A3EEB679 /* PositionEntity.swift */; };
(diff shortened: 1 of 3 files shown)


garthvh (approved):
LGTM — correct byte-order fix, well-tested, no migration needed.

Verified:
- The old helpers packed big-endian (`parts[0] << 24 | …`); the firmware (Arduino/ESP32 `IPAddress`) and the Android app store IPv4 little-endian (first octet = least-significant byte), so addresses displayed and would have written byte-reversed. The new encode/decode (`parts[0] | parts[1] << 8 | …`) matches the Android reference and the known mapping `192.168.1.1 ↔ 0x0101A8C0`.
- No data migration: `node.networkConfig?.ip` is the raw firmware `uint32` (read via `UInt32(bitPattern:)`); the helpers are used only to decode it for display (`loadValues`) and re-encode the typed string on save. The stored value is untouched, so existing addresses now render correctly and writes go out in the right order.
- Good test coverage — known value, the asymmetric `10.20.30.40` case (catches a swap a palindrome wouldn't), round-trips, subnet mask, and malformed/edge inputs.

Agree the malformed-input → `0` (no validation feedback) is pre-existing and fine to track separately. Applied `skip-docs-check` since this is a bugfix to existing field behavior with no doc-mapping change.

PR #1971, opened 2026-06-20
Reduce redundant list refresh work

RCGV1:
## Summary
- Avoid reassigning the throttled node-list snapshot when the visible node IDs, order, and SwiftData model instances are unchanged.
- Refresh direct-message summaries only for contacts that are currently visible after filtering.

## Reasoning
This keeps the existing throttled recompute behavior, but avoids extra SwiftUI list invalidation during live packet ingestion when the visible node set has not actually changed. The contact-list change narrows summary refresh work to rows that can be displayed, while leaving the existing message fetch semantics intact.

## Verification
- `git diff --check`
- `xcodebuild -workspace Meshtastic.xcworkspace -scheme Meshtastic -configuration Debug -destination 'platform=iOS Simulator,id=DE3343B3-1E9D-4C2D-9AFE-B50AC465BEBA' -derivedDataPath /tmp/meshtastic-round2-derived -skipPackagePluginValidation -skipMacroValidation build`
- Installed and launched the simulator app with `--meshtastic-perf-seed --meshtastic-perf-reset` and `MESHTASTIC_PERF_SEED_NODES=1000`; captured a smoke screenshot showing the seeded Nodes list rendering normally.

No screenshot attached because this PR has no intended visual change.

### DIFF
--- a/Meshtastic/Views/Messages/UserList.swift
+++ b/Meshtastic/Views/Messages/UserList.swift
@@ -129,7 +129,7 @@ private struct FilteredUserList: View {
 		let dateFormatString = (localeDateFormat ?? "MM/dd/YY")
 		let activeDeviceNum = Int64(accessoryManager.activeDeviceNum ?? 0)
 		let visibleUsers = users.filter { $0.num != activeDeviceNum }
-		let summaryUsers = allUsers.filter { $0.num != activeDeviceNum && $0.lastMessage != nil }
+		let summaryUsers = visibleUsers.filter { $0.lastMessage != nil }
 		let summaryRefreshKey = directMessageSummaryRefreshKey(for: summaryUsers)
 		let currentDay = Calendar.current.dateComponents([.day], from: Date()).day ?? 0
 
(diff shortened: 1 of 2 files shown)


garthvh (approved):
LGTM — correct, behavior-preserving, and a sensible reduction in redundant work.

Verified the main correctness concern (could skipping the `displayedNodes` reassignment leave the list stale?) and it's safe:

- `NodeListEntry` is just `(id, node)` with no baked-in display values, and rows render from the `@Observable` `NodeInfoEntity` directly — so per-node field changes still re-render the row via observation regardless of whether `displayedNodes` is reassigned.
- Order/membership changes are still caught: `allNodes` is `@Query(sort: \.lastHeard, .reverse)`, so a node moving re-sorts → `displayNodes()` returns a different order → `displayedNodeIDsDiffer` detects the id-position mismatch (and count changes catch add/remove). The only thing skipped is a genuine no-op reassignment when the visible set is identical (same ids, order, instances) — exactly the redundant `List` invalidation worth avoiding during off-screen packet churn. Worst case (instances differ) it falls back to current behavior, never stale.

The `UserList` change (scoping `summaryUsers` to `visibleUsers` instead of `allUsers`) is likewise a strict reduction — off-screen contacts aren't displayed, so there's no visible regression.

Nice low-risk win for heavy packet-ingestion on large meshes.

PR #1983, opened 2026-06-23
Harden channel QR import and sharing

RCGV1:
## Summary

This PR hardens channel QR sharing/import so iOS is more tolerant of supported Meshtastic channel links and less likely to produce broken or misleading channel QR flows.

- Adds a shared `MeshtasticChannelURL` helper for supported channel forms: canonical `https://meshtastic.org/e/#...`, explicit `meshtastic://` channel links, and raw payload imports used internally by the save sheet.
- Explicitly does **not** claim `/v/` links, which remain contact-share links handled by the existing contact flow.
- Validates channel QR saves before mutating local state or writing to the connected radio, including empty channel sets, slot limits, duplicate names, and replace-mode requirements.
- Aligns add-vs-replace behavior with Android expectations: add mode appends channels without applying LoRa config, while replace mode can apply LoRa settings from the QR payload.
- Removes the redundant manual import-channel-link UI from the share screen and relies on normal iOS universal-link/open-URL handling.
- Cleans up the share QR channel picker so it only appears when there are actual shareable channels and no longer leaves empty/off-center header content.
- Leaves associated-domain entitlements and DEBUG performance seeding out of this PR.

## Logic

The main behavior change is centralizing URL parsing and payload extraction before UI or radio writes happen. The app now decodes supported Meshtastic channel-link forms through one helper, then the save path performs validation before it touches Core Data or sends admin messages to the radio. That keeps bad imports from partially changing channel state.

For cross-platform safety, add mode intentionally clears LoRa config from the decoded payload before saving. That matches the safer Android-style behavior where adding channels should not silently overwrite modem/frequency settings. Replace mode keeps the existing full-import behavior and makes that explicit in the review copy.

The parser is deliberately scoped so `/v/` contact URLs and nested/non-channel paths do not get intercepted by channel handling.

## Screenshots

Share QR screen:

Share QR screen

Import review screen with shorter replace-mode copy:

Import review screen

## Tests Done

- Ran focused URL parser/save-link tests: `MeshtasticTests/MeshtasticChannelURLTests` passed via `xcodebuild test` on iPhone 16e simulator.
- Ran `git diff --check` successfully.

Known note: the project still emits existing build-time SwiftLint/Swift 6/SwiftData/duplicate build warnings unrelated to this change, but the focused tests completed successfully.

## Summary by CodeRabbit

* **New Features**
  * Added canonical channel URL generation and parsing to power channel-link and QR workflows.
  * Refined the “Save Channel QR Code” experience with selectable channels, add vs. replace mode, and an LoRa change preview.
* **Bug Fixes**
  * Improved channel-link handling across the app with stricter validation and consistent error behavior.
  * When replacing channels, preserves existing radio LoRa settings and avoids duplicate LoRa transmissions.
* **Documentation**
  * Updated messages documentation with a new “Channel QR Codes” subsection and refreshed help text.
* **Tests**
  * Added regression tests for channel-set saving/reboot scenarios and channel URL parse/generate round-trips.


garthvh (changes requested):
Thanks for this — the core direction is good. Centralizing channel-URL parsing into one helper, validating a `ChannelSet` *before* clearing existing channels (no more channel-wipe on a malformed URL), and adding tests are all real improvements worth keeping.

A few things need to change before this can land:

### 1. `/v/` is not a legacy channel format — it's the current *contact* format
The PR treats `meshtastic.org/v/#` as a legacy channel link. It isn't. `/v/` is the live contact-share (vCard) format, routed to `ContactURLHandler.handleContactUrl` and used by `AddContactIntent`, `ShareContactQRDialog`, and `MessageText`. See the current routing in `MeshtasticApp.swift`:

```swift
if ...contains("meshtastic.org/v/#") {
    ContactURLHandler.handleContactUrl(...)   // contacts
} else if ...contains("meshtastic.org/e/") {
    handleChannelLinkURL(...)                  // channels
}
```

If `MeshtasticChannelURL` claims `/v/`, it will intercept contact links and try to decode them as a `ChannelSet`, which breaks the contacts flow. The channel parser must explicitly **not** match `/v/`.

### 2. Drop the legacy-channel parsing
There's no `/c/` (the actual old channel pattern) anywhere in the tree — the app can't produce it and nothing consumes it, so parsing legacy channel formats is dead code. Please scope the channel parser to the formats we actually use: `meshtastic.org/e/#…` and the `meshtastic://` scheme. That also keeps `canHandle()` from over-claiming URLs that belong to other flows.

### 3. Please drop the entitlement / associated-domains change from this PR
Path scoping is enforced by the server-side `apple-app-site-association`, not by the entitlement. The live AASA (served from both `meshtastic.org` and `www.meshtastic.org`) already scopes things correctly:

```
Allowed:  /d/*, /e/*, /v/*   (+ uppercase)
Excluded: NOT /docs/*, NOT /downloads/*, NOT /showcase/*
```

So docs/downloads/showcase are already excluded and won't route to the app regardless of the entitlement string. Given that:
- The `applinks:meshtastic.org/e/*` → bare `applinks:meshtastic.org` change is cosmetic — the entitlement format is host-only, so the path suffix was never doing the scoping anyway.
- The only functional change here is adding `www.meshtastic.org` as an associated host, which is a deliberate "we now support www universal links" decision, not channel-QR hardening.

Please remove the entitlement change from this PR. If there's a real bug behind it (universal links not opening, or we genuinely need `www` support), split it into its own PR that states the reason — it shouldn't ride along silently here.

### 4. Minor: scope creep
The DEBUG perf-seed / `simulateConnectedDevice` additions are unrelated to channel QR hardening. Not a blocker since they're `#if DEBUG`, but they'd be cleaner as a separate PR.

Happy to re-review quickly once the parser is scoped to `/e/` + scheme and the `/v/` handling is removed.

RCGV1:
Addressed the review feedback in `1bc5669b`:

- Scoped `MeshtasticChannelURL` to channel links only: `meshtastic.org/e/...` and explicit `meshtastic://` channel routes.
- Added tests proving `/v/` contact URLs are not claimed by channel handling.
- Removed the broad/nested path handling (`/channel/e/...`) so the parser no longer over-claims unrelated Meshtastic URLs.
- Reverted the associated-domains entitlement change from this PR.
- Removed the DEBUG simulated connected-node/perf-seed scope creep from this PR.
- Updated the share QR placeholder to the canonical `https://meshtastic.org/e/#` form.

Validation:

- `git diff --check` passes.
- `xcodebuild test -workspace Meshtastic.xcworkspace -scheme Meshtastic -destination 'platform=iOS Simulator,id=DE3343B3-1E9D-4C2D-9AFE-B50AC465BEBA' -derivedDataPath ~/Library/Developer/Xcode/DerivedData/Meshtastic-ChannelQR-Hardening -only-testing:MeshtasticTests/MeshtasticChannelURLTests -quiet` exited successfully.

Note: the test run emitted existing project warnings (SwiftLint missing, Swift 6/SwiftData warnings, duplicate build phase warnings), but the focused tests passed.

garthvh:
Resolved the merge conflicts with main and pushed — the PR is mergeable again. Resolution notes:

- `MessageText.swift`: took main's `cornerBadges` refactor (the XEdDSA signing-shield badge row from #1993); this PR's `handleURL` rework merged cleanly alongside it.
- `project.pbxproj`: union of both sides' test-file registrations (this PR's `MeshtasticChannelURLTests`/`ChannelSetSaveTests` plus main's newer entries); `plutil -lint` clean.
- `Meshtastic/Resources/docs/index.json`: regenerated with `scripts/build-docs.sh` from the merged markdown sources.

Both suites pass against current main: Meshtastic channel URLs (9) and Channel Set Save (7), 16 tests total.

garthvh (approved):

PR #1986, opened 2026-06-23
fix: validate static IPv4 fields in Network config before save

bruschill:
## What changed?

Static IPv4 fields in the Network config (IP / Gateway / Subnet / DNS) are now validated before they can be saved:

- A new `isValidIPv4Field` helper checks for a strict dotted quad — each octet must be 1–3 ASCII digits in the range 0–255, which also rejects sign characters (`+1`) and whitespace that a bare `UInt32` parse would otherwise accept.
- A field holding a malformed address tints **red**, and an inline hint appears under the section.
- The **Save** button is disabled while any static field holds a malformed address (via a pure, unit-testable `isStaticConfigValid(addressMode:ip:gateway:subnet:dns:)` gate).
- A **blank** field stays valid and means "unset" (written as `0.0.0.0`). This matches how the firmware round-trips an unconfigured field (`uint32ToIpString(0)` is `""`), so a node already in Static mode with zeroed fields loads cleanly and unrelated edits remain saveable — only a non-empty typo blocks the save.

## Why did it change?

This is the deferred follow-up from #1965 (the IPv4 byte-order fix). `ipStringToUInt32` silently maps any malformed/incomplete input (e.g. `192.168.1` or `192.168.1.300`) to `0`, and the Network config wrote that result straight to firmware. The result: a typo in a static IP/gateway/subnet/DNS field saved as `0.0.0.0` with no feedback, leaving the device with a broken static network configuration. This change surfaces the bad field and blocks the save instead of writing garbage.

## How is this tested?

Extended `MeshtasticTests/NetworkConfigIPConversionTests.swift` (Swift Testing). All 18 tests pass; run via the workspace on an iPhone 16 / iOS 18 simulator (matching CI):

- `isValidIPv4Field`: empty (valid/unset), 5 well-formed, 11 malformed (too few/many octets, out-of-range, empty octets, non-numeric, `+1`, embedded whitespace, overlong `0000`).
- `isStaticConfigValid` gate: DHCP bypass, well-formed, all-blank, partially-blank (all valid), and malformed IP/gateway/DNS (each blocks save). The gate was extracted to a pure static function precisely so it can be tested without a live SwiftUI `@State` context.

```
  -destination 'platform=iOS Simulator,name=iPhone 16,OS=18.x' \
  -only-testing:MeshtasticTests/NetworkConfigIPConversionTests
# ** TEST SUCCEEDED ** — Test run with 18 tests in 1 suite passed
```

## Checklist

- [x] I have verified whether these changes require updates to the in-app documentation under `docs/user/` or `docs/developer/`, and updated accordingly. _No doc update needed — this only adds validation/feedback to existing Network config fields (no new fields or settings). Requesting the **`skip-docs-check`** label (external-contributor PR can't self-apply it)._

## Summary by CodeRabbit

* **New Features**
  * Added inline validation and user feedback for static IPv4 network settings.
  * Static IP, gateway, and subnet now highlight invalid required input; DNS highlights valid/invalid optional input.

* **Bug Fixes**
  * Prevent saving when required static fields contain malformed or blank IPv4 values.
  * Saving remains available in DHCP mode, even if static fields are not valid.

* **Tests**
  * Added automated unit tests for IPv4 validation and save-gating logic in both DHCP and static modes.


garthvh (changes requested):
Thanks for tackling this — surfacing the silent `0.0.0.0` corruption is a real fix, the scoping to static mode (`guard addressMode == 1 else { return true }`) is correct so DHCP saves are unaffected, and I appreciate the test coverage.

One change before merge: **blank IP/gateway/subnet should not be savable in static mode.**

Right now blank fields are treated as valid in static mode (and `staticModeWithBlankFieldsIsValid` confirms all four can be empty and still save). That lets a user select Static, leave IP/gateway/subnet blank, and Save — which silently writes `0.0.0.0` for them. That's the same class of silent-broken-config this PR is meant to eliminate, just moved from "typo" to "incomplete."

Could you require IP, gateway, and subnet to be non-blank when `addressMode == 1`? DNS being blank is fine to keep optional. That's an extra condition in `isStaticConfigValid` plus updating `staticModeWithBlankFieldsIsValid` to assert it's now invalid (and a complementary test where all required fields are filled).

Format validation here is solid — this last bit just makes it completeness-validation too, so a static config can't be saved in a non-functional state.

garthvh:
Addressed the requested change — blank fields are no longer savable in static mode:

- `isStaticConfigValid` now requires **IP, gateway, and subnet to be non-blank** when `addressMode == 1`, in addition to the format checks. DNS stays optional: blank means "unset" and writes 0.0.0.0, which is fine for DNS.
- The three required fields tint red while blank (via a new `isRequiredIPv4FieldValid` tint helper) so it's visible which fields are blocking Save, placeholders now show realistic examples (`192.168.1.10`, `192.168.1.1`) instead of `0.0.0.0`, and DNS's placeholder says "Optional". The inline hint reads: "IP, gateway, and subnet are required and must be valid IPv4 addresses (e.g. 192.168.1.10). DNS may be left blank."
- Tests updated per the review: `staticModeWithBlankFieldsIsValid` is now `staticModeWithBlankFieldsBlocksSave` asserting invalid, plus per-field blank-blocks-save tests for IP/gateway/subnet, a blank-DNS-still-valid test, and a tint-helper test. The existing malformed-gateway/DNS tests also had their other fields filled with valid values so each isolates exactly one failure cause. 22 tests, all passing.

One behavior note: a node already in static mode with zeroed fields now loads with Save blocked until the required fields are filled — per the review, that incomplete state is the same class of silently-broken config, so it's intentional that it can't be re-saved as-is.

Also applied the `skip-docs-check` label requested in the PR description.

garthvh (approved):

Issue #1970, opened 2026-06-19
🐞 [Bug]: GeoJSONStyledFeature causes MKOverlay creation failure for Point features


EduardoGBarron:
### Before submitting

- [x] I have searched existing issues to make sure this bug hasn't already been reported
- [x] I have updated to the latest version of the software to verify the issue still exists

### Firmware Version

2.7.13 (2045)

### What did you do?

Loaded custom GeoJSON overlay files in Meshtastic iOS/iPadOS 2.7.12.

I tested several GeoJSON FeatureCollections containing only Point geometries:

* 1 Point
* 10 Points
* 50 Points
* 152 Points

I also tested variations with and without:

* emoji characters in the name property
* marker-color
* marker-size
* additional properties

Finally, I created a minimal GeoJSON containing only:

* FeatureCollection
* Point geometries
* a single name property

All files rendered correctly on the map.

### Expected Behavior

Point features should render on the map without generating GeoJSON overlay errors in the application log.

If Point geometries are not intended to be processed as MKOverlay objects, they should be handled separately or ignored by overlay creation code without logging errors.

### Current Behavior

The application log is flooded with repeated messages:

GeoJSONStyledFeature: Failed to create overlay - no valid MKOverlay geometry

Despite these errors, all Point features render correctly on the map.

The error count scales linearly with the number of Point features:

* 1 Point → 1 error
* 10 Points → 20 errors
* 50 Points → 100 errors
* 152 Points → 304 errors

This pattern appears consistent across multiple test files and remains present even with minimal GeoJSON containing only Point geometries and a name property.

### Participation

- [ ] I am willing to submit a pull request for this issue.

### Additional comments

The GeoJSON files appear to be valid and render correctly.

Based on testing, the issue does not seem related to:

* emoji or Unicode characters
* marker-color
* marker-size
* additional properties

A possible explanation is that Point geometries are being processed by an MKOverlay creation path before being rendered through a separate Point/annotation rendering path. However, this is only a hypothesis based on observed behavior and error scaling; I have not inspected the source code.

I can provide the test GeoJSON files and additional logs if useful.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

EduardoGBarron:
> ## 🤖 Bug Report Received
> 
> Thank you for this detailed bug report! A maintainer will review it and investigate the root cause.
> 
> If you can provide any of the following it will speed up the investigation:
> 
[...]

PR #1987, opened 2026-06-24
fix: render GeoJSON Multi* geometries on the mesh map

bruschill:
## What changed?

GeoJSON `MultiPoint`, `MultiLineString`, and `MultiPolygon` features now render on the mesh map. Previously they produced no (or only partial) map content:

- `MeshMapContent.overlayContent` only branched on `Point` / `LineString` / `Polygon`, so Multi* geometry types matched no branch and drew nothing.
- `GeoJSONStyledFeature` kept only `mkFeature.geometry.first`, so even where a branch existed, every sub-geometry past the first was silently dropped.

The fix:

- `GeoJSONStyledFeature` now collects **all** decoded sub-geometries into `precomputedOverlays`, flattening any `MKMultiPolygon` / `MKMultiPolyline` wrapper into simple `MKPolygon` / `MKPolyline` so the render layer handles one uniform list.
- `overlayContent` renders the overlay layer by switching on overlay type (`MKPolyline` → `MapPolyline`, `MKPolygon` → `MapPolygon`), so `LineString`/`Polygon` and their Multi* variants are drawn the same way.
- A new `markerCoordinates` (precomputed at init, like the overlays) draws annotations for both `Point` and `MultiPoint`.
- Tidied an unused-binding warning in `MapDataManager.validateFile`.

## Why did it change?

Valid, common GeoJSON that uses Multi* geometries silently produced no/partial map content. The gap was surfaced during the #1970 review (the Point overlay log-spam fix); after #1970 quieted the spurious Point overlay error, the silent drop became even easier to miss. This makes the map render the geometry types that the GeoJSON spec — and typical exported overlays — routinely contain.

## How is this tested?

Added model-layer tests in `MeshtasticTests/GeoJSONOverlayConfigTests.swift` (Swift Testing). All GeoJSON tests pass (64 total); run via the workspace on an iPhone 16 / iOS 18 simulator (matching CI):

- **Overlay counts per geometry**: single Polygon → 1, MultiPolygon → 2 (the direct regression guard against the old `.first` truncation), MultiLineString → 3, LineString → 1, and Point / MultiPoint → 0 overlays.
- **Marker coordinates**: Point → 1, MultiPoint → N, malformed sub-coords skipped, Polygon → none; plus `toCoordinates()` parsing.
- **Empty/malformed input**: empty MultiPolygon and malformed coordinates → empty overlay set (the error/empty-result branches).

```
  -destination 'platform=iOS Simulator,name=iPhone 16,OS=18.x' \
  -only-testing:MeshtasticTests/GeoJSONStyledFeatureOverlayTests \
  -only-testing:MeshtasticTests/GeoJSONFeatureMarkerCoordinatesTests
# ** TEST SUCCEEDED **
```

Note: a SwiftUI `Map` snapshot test was evaluated and deliberately not included — MapKit draws overlays asynchronously, so an off-screen snapshot captures a blank base map and can't assert overlay rendering. The deterministic guard lives at the model layer (overlay/marker counts) instead.

Known follow-up (out of scope): `GeometryCollection` member points still aren't drawn as markers.

## Checklist

- [x] I have verified whether these changes require updates to the in-app documentation under `docs/user/` or `docs/developer/`, and updated accordingly. _No doc update needed — this fixes rendering of existing GeoJSON overlay support; no new user-facing settings or views. Requesting the **`skip-docs-check`** label (external-contributor PR can't self-apply it)._

### DIFF
--- a/Meshtastic/Helpers/GeoJSONOverlayConfig.swift
+++ b/Meshtastic/Helpers/GeoJSONOverlayConfig.swift
@@ -165,6 +165,19 @@ struct GeoJSONFeature: Codable {
 		default: return 4.0
 		}
 	}
+
+	/// Coordinates to draw as map markers. A Point yields one coordinate; a MultiPoint yields one
+	/// per sub-coordinate. Empty for every other geometry type (those render as overlays instead).
+	var markerCoordinates: [CLLocationCoordinate2D] {
+		switch geometry.type.lowercased() {
+		case "point":
+			return geometry.coordinates.toCoordinate().map { [$0] } ?? []
+		case "multipoint":
+			return geometry.coordinates.toCoordinates()
+		default:
+			return []
+		}
+	}
 }
 
 // MARK: - Styled Feature Wrapper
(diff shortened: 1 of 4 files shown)


garthvh:
Heads-up on an incoming collision with the `feat/pmtiles-mapkit-shim` branch — this is a good fix (the `.first` sub-geometry drop is a real bug we currently ship), but it'll need reconciliation, so flagging early.

The pmtiles branch reworks the map render path and will conflict with this PR in three places:

1. **`createOverlay()` removal.** This PR removes `createOverlay()` and replaces `precomputedOverlay` (singular) with `precomputedOverlays: [MKOverlay]`. The pmtiles branch adds a new consumer, `MeshMapMK.swift`, that calls `styled.createOverlay()` — so after both land that's a compile break, not just a merge marker. Whoever merges second needs to update `MeshMapMK` to iterate `precomputedOverlays`.

2. **`MeshMapContent.swift`.** This PR edits the rendering loops there, but the pmtiles branch moves overlay rendering out of `MeshMapContent` into `MeshMapMK` + `GeoJSONOverlayManager`, so some of the edits here target code paths that branch has removed.

3. **`GeoJSONOverlayConfigTests.swift`.** Both this PR and the pmtiles branch add this file → add/add conflict; the two test sets will need to be merged.

No change requested on the substance — plan is to land this on `main` first, then rebase pmtiles and port the plural-overlay API + this fix into the relocated render path. Just didn't want us to collide silently.

garthvh:
Superseded by #1989.

The Map tab MKMapView migration (#1989) removed the SwiftUI `MeshMapContent` renderer that this PR patched, and its new architecture already renders `MultiPolygon` / `MultiLineString` correctly via `MKGeoJSONDecoder` + the `ClusterMapView` multi-geometry renderers.

The one remaining gap this PR addressed — `MultiPoint` markers (which now render nothing, since `toCoordinate()` returns nil for the nested coordinate array) — is carried forward in #2000, ported onto the new `MeshMapMK` overlay path.

Closing as superseded. 🙏

PR #1990, opened 2026-06-26
feat: 2.8 protobuf sync + LoRa region/preset compatibility map

thebentern:
## What changed?

Syncs the vendored protobufs to 2.8 and implements the LoRa **region → preset compatibility** feature, with the new 2.8 regions and presets gated behind connected-firmware support.

### Protobuf sync (fix the build)

- Regenerate the vendored `.pb.swift` from the 2.8 `protobufs` submodule (plus the new `deviceonly_legacy` / `mesh_beacon` files).
- Handle the new `meshBeaconApp` PortNum (37) in the packet `switch`.
- Realign `ModuleConfig.TrafficManagementConfig` to its redesigned 2.8 schema — the per-feature boolean flags plus `positionPrecisionBits` and the hop-management trio were dropped in favor of five "non-zero implies enabled" integers. Updates the proto↔entity mapping (`UpdateSwiftData`) and the config view (removes the defunct precision-bits and hop-management UI). The SwiftData entity is left intact, so **no migration** is required.

### LoRa region → preset map

Implements the firmware [client spec](https://github.com/meshtastic/firmware/blob/a84a3a984e8682dd8328fe026ed4855acc61a3c6/docs/lora_region_preset_compatibility_client_spec.md):

- Decode `FromRadio.region_presets` into a per-region lookup (`LoRaRegionPresetMap.decoded()` → `RegionPresetInfo`), dispatched from `AccessoryManager.processFromRadio` and stored on the connection model; reset on disconnect.
- `LoRaConfig` constrains the preset picker to the selected region's legal set, switches to the region's `default_preset` when the current one is illegal, and warns on licensed (ham) bands (coordinated with the operator's `is_licensed` flag). Falls back to the full list when the map/region is absent — never an empty picker, and only consulted on 2.8 firmware.

### Firmware gating

- The new 2.8 regions (ham 2m / 70cm / 1.25m, EU 866 / narrow) and presets (Lite / Narrow / Tiny) are **hidden on radios reporting firmware < 2.8.0** via `RegionCodes` / `ModemPresets` `selectable(supports2_8:)`. On 2.7.x and earlier they don't appear in the pickers.

### Docs

- New developer page `docs/developer/lora-region-presets.md`; updated LoRa + Traffic Management sections in `docs/user/settings.md`; what's-new entries; regenerated the in-app documentation bundle (`Meshtastic/Resources/docs/`, 7.8 MB / 10 MB).

### Tests

- `LoRaRegionPresetMapTests` (decodes the spec's reference table; covers out-of-range / absent-region / licensing cases) and `LoRaFirmwareGatingTests`. Updated the `RegionCodes` / `ModemPresets` count assertions.

## Testing

- `xcodebuild build-for-testing` (iOS Simulator) — **TEST BUILD SUCCEEDED**.
- Targeted run — 38 LoRa/enum tests plus `DocBundleTests` pass.


garthvh (approved):
Looks good

PR #2009, opened 2026-06-29
fix(connect): confirm before powering off a node from the connection screen (#2006)

### LINKED ISSUE #2006: 🐞 [Bug]: iOS needs confirmation dialog for the "Power Off" button from the connection screen
### Before submitting

### Firmware Version

2.7.15 (iOS app version)

### What did you do?

When you go to the connection dialog and long press on the node that you're connected to in order to disconnect, a dialog pops up with two options: 1) Disconnect 2) Power Off.  There is no confirmation dialog when pressing the power off, making it extremely easy to accidently power off a node rather than disconnect.  

This is especially important when working with roof or remote nodes.  There is no way to recover from a mistake like this other than climbing up to the remote node, weather it be high on a tower, a remote location or on a roof. 

There **is** a confirmation dialog when powering off a node from the node's entry in the node list.  This functionality should be added to the connection dialog to avoid mistakenly shutting down a node.

### Expected Behavior

When on the connection list and doing a long press to bring up the dialog for disconnecting, there should be a confirmation dialog if the Power Off option is selected.

### Current Behavior

There is no confirmation dialog and the node shuts down immediately, causing a problem if the option was selected by mistake or a inaccurately placed finger.

### Participation

### Additional comments

_No response_

### Code of Conduct

bruschill:
## What changed?

Adds an "Are you sure?" confirmation dialog before the **Power Off** action in the connection screen's long-press context menu. Previously, long-pressing the connected node and tapping "Power Off" shut the node down immediately with no confirmation.

- The Power Off button sets a confirm flag instead of shutting down directly.
- A `confirmationDialog` ("Are you sure?" → destructive "Shutdown Node?") performs the actual `sendShutdown`, mirroring the existing flow in `NodeDetail`, with `titleVisibility: .visible` so the message shows.
- The dialog is attached to the always-present root `VStack` rather than the connected-device subtree, so it survives a connection state change between the long-press and confirm.
- The shutdown target is resolved at confirm time via `Connect.shutdownTarget(for:)`, which gates on `modelContext != nil`. This avoids a crash: capturing/reading a cached SwiftData `@Model` that has faulted (after a disconnect/reconnect or node switch while the dialog is up) would trap; resolving through the guard yields a safe, logged no-op instead.

## Why did it change?

Per issue #2006: the node list already confirms before powering off, but the connection screen did not. An accidental tap could power off a roof, tower, or otherwise remote node, with no way to recover short of physically reaching the hardware. This makes the connection screen consistent with the node list and prevents accidental shutdowns.

Fixes #2006.

## How is this tested?

- Extracted `Connect.liveNode(_:)` (the `modelContext != nil` guard) and `Connect.shutdownTarget(for:)` into static, value-in/value-out helpers and added unit tests in `ConnectViewTests.swift`: nil, detached, live-with-user, live-without-user, the guard itself, and an inserted-then-deleted node that reproduces the real detach path and confirms the guard skips it (so `.user` is never read on a faulted model). Tests use a fresh `ModelContext` (no save → no leak) and `withExtendedLifetime` to keep the context alive.
- The dialog *presentation* itself isn't unit-testable here (no ViewInspector, `AccessoryManager` is concrete, confirmation dialogs don't render into snapshots), so that part is covered by manual verification.
- Built and ran the suite against the iOS 18 simulator; build succeeds and all tests pass.

_N/A — adds a standard system confirmation dialog._

## Checklist

- [x] I have verified whether these changes require updates to the in-app documentation under `docs/user/` or `docs/developer/`, and updated accordingly. Updated `docs/user/bluetooth.md` (new "Powering Off a Radio" section + tip) and regenerated the bundled HTML.

## Summary by CodeRabbit

* **New Features**
  * Added “Power Off” for connected radios, now requiring confirmation before sending the shutdown request.

* **Bug Fixes**
  * Safer shutdown handling that avoids acting on detached/invalid nodes and prevents mismatched identities from triggering incorrect shutdowns.

* **Documentation**
  * Updated Bluetooth guides and tips with a new “Powering Off a Radio” section, clarifying **Disconnect** vs **Power Off** and the long-press behavior (including Live Activity).

* **Tests**
  * Added Connect view tests covering creation, shutdown target resolution, and live node lookup behavior.


garthvh (approved):
Good fix for #2006 — mirrors the existing NodeDetail confirmation flow, and resolving the shutdown target through the modelContext guard at confirm time avoids the faulted-model trap if the connection changes while the dialog is up.

PR #1682, opened 2026-04-22
Working Changes 2.7.11

garthvh:
- **Onboarding cleanup** — merged Background Activity step into Location, reordered highlights, fixed button padding, auto-advance after permission request, defer Siri auth to onboarding screen
- **BLE transport** — `guard centralManager != nil` safety checks, log message cleanup
- **Invalid UTF-8 fix** (#1687) — catches bad UTF-8 in the BLE drain loop instead of triggering a reconnect loop; new `InvalidUTF8Tests` suite
- **OTA firmware update** (#1522) — ESP32 BLE OTA, ESP32 Wi-Fi OTA, NRF DFU, and UF2 mass-storage update paths
- **BLE/TCP stream fixes** (#1681) — finish stale continuations before creating new ones, guard `onTermination` against nil self, only reconnect on `.cancelled`, remove duplicate `connectionDidDisconnect` call
- **Mac tapbacks + distance filter cleanup** — open Character Palette on Mac Catalyst for emoji input; distance filter now uses `currentPreciseLocation` (≤100 m accuracy, not stale) across node list, user list, and map; toggle disabled with explanatory caption when no fix is available; map reloads when distance setting changes

PR #2011, opened 2026-06-29
fix(channels): don't report QR channel-set save as failed after device reboot (#2010)

### LINKED ISSUE #2010: 🐞 [Bug]: iOS app fails to apply QR-generated config — 'failed to save' error on valid config.
### Before submitting

### Firmware Version

2.7.12 (1831)

### What did you do?

New RAK4631, Using IOS 2.7.12 - scanning QR code for local mesh. Applying the config fails. 

Image

Image

### Expected Behavior

Android users _tell_ me that same QR config works to configure their devices. 
The IOS app should not error "Failed to shave channel configuration"

### Current Behavior

B8B8 asserts that 0 means default, therefore when the iOS app attempts to apply TX = 0, Ch# = 0, BW = 0, CR = 0, SF = 0. it SHOULD apply the defaults (for MediumFast) EVEN THOUGH the QR code has these values explicitly set..... TX = 30dbm, Ch# = 45, BW = 250, CR = 5, SF = 11

Perhaps this is a collision with manual values and the Use Preset selector?

either way the app throws an error: Failed to save channel configuration

### Participation

- [x] I am willing to submit a pull request for this issue.

### Additional comments

I suppose i could PR this, but I dont know the codebse and wouldn't know where to begin.

### Code of Conduct

bruschill:
## What changed?

Fixes #2010. Applying a QR-generated channel set no longer reports "Failed to save channel configuration" when the configuration is actually applied successfully.

- Treat the post-reboot `wantConfig` disconnect as success in `AccessoryManager.saveChannelSet`. A substantial LoRa change reboots the radio and drops the connection; the follow-up `wantConfig` is expected to fail, and that failure is no longer surfaced as a save error. The tolerance only applies when a reboot-causing LoRa config was actually transmitted — otherwise `wantConfig` errors still surface.
- Removed the duplicated LoRa-config send block (a copy-paste from #1682 that transmitted the config twice).
- Clear the local channels only **after** the channel-set link decodes successfully, and throw on a malformed link. Previously a bad/undecodable link wiped the local channel list and returned a false success without ever contacting the device.
- Skip the LoRa send when the channel set carries no LoRa config, so a channels-only share link no longer pushes a default-initialized (`region UNSET`) config that would knock the radio off its frequency.

## Why did it change?

This is a regression introduced by the BLEManager → AccessoryManager transport refactor. The legacy `BLEManager.saveChannelSet` was non-throwing: writes were fire-and-forget (guarded by `peripheral.state == .connected`) and it returned `true` immediately after kicking off a non-throwing `sendWantConfig()`. A reboot-induced disconnect could therefore never be reported as a failure.

The new path uses an `async throws` `send()` that throws the instant the link is gone, plus an awaited, throwing `sendWantConfig()` that blocks on a continuation. When the device reboots after the LoRa change, those throw, and `SaveChannelQRCode` flattens any error into "Failed to save channel configuration" — even though the channels and LoRa config were already delivered. Android applies the same QR fine, which matches: the config is valid and does get applied.

## How is this tested?

Added `MeshtasticTests/ChannelSetSaveTests.swift` — a `MockChannelSetConnection` actor conforming to `Connection` drives `saveChannelSet` without a real transport and simulates the post-reboot disconnect by throwing on the `wantConfig` send. Four tests, each verified to **fail on the pre-fix code and pass with the fix**:

- A QR channel set save survives the post-reboot disconnect without surfacing an error (#2010).
- The LoRa config is transmitted exactly once (guards the #1682 duplicate).
- A malformed channel-set link throws instead of silently succeeding, and nothing is put on the wire.
- A channels-only replace (no embedded LoRa config) does not transmit a LoRa config.

Run with: `-only-testing:MeshtasticTests/ChannelSetSaveTests` (iPhone 16 / iOS 18 simulator). All pass.

N/A — behavior/error-handling fix; no UI changes.

## Checklist

- [x] I have verified whether these changes require updates to the in-app documentation under `docs/user/` or `docs/developer/`, and updated accordingly. No doc update is needed (no view/behavior is documented for this path) — requesting the **`skip-docs-check`** label.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved channel-set saving so existing channels are only cleared after the incoming data is successfully validated.
  * Made saving more reliable when a device may briefly disconnect after a configuration update.
  * Prevented duplicate configuration sends during full channel replacement.
  * Invalid channel-set data now fails cleanly without changing saved channels.

* **Tests**
  * Added regression coverage for channel-set saving, including reboot/disconnect handling, malformed data, and configuration send behavior.

### DIFF
--- a/Meshtastic.xcodeproj/project.pbxproj
+++ b/Meshtastic.xcodeproj/project.pbxproj
@@ -310,6 +310,7 @@
 		DD00003ASUPPORTLVL0000001 /* SupportLevel.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD00003ASUPPORTLVL0000000 /* SupportLevel.swift */; };
 		DD00003C000SNAP00000001 /* SnapshotTesting in Frameworks */ = {isa = PBXBuildFile; productRef = DD00003B000SNAP00000000 /* SnapshotTesting */; };
 		DD0000DEVLNKTST000000001 /* DeviceLinkTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD0000DEVLNKTST000000000 /* DeviceLinkTests.swift */; };
+		DD0000CHSETST0000000001 /* ChannelSetSaveTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD0000CHSETST0000000000 /* ChannelSetSaveTests.swift */; };
 		DD007BAE2AA4E91200F5FA12 /* MyInfoEntityExtension.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD007BAD2AA4E91200F5FA12 /* MyInfoEntityExtension.swift */; };
 		DD007BB02AA5981000F5FA12 /* NodeInfoEntityExtension.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD007BAF2AA5981000F5FA12 /* NodeInfoEntityExtension.swift */; };
 		DD09240001E7FAD600E70001 /* MapLegend.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD09240002E7FAD600E70001 /* MapLegend.swift */; };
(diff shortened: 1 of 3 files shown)


RCGV1:
This may have conflicts with my other PR

RCGV1:
https://github.com/meshtastic/Meshtastic-Apple/pull/1983

bruschill:
Superseded by RCGV1/Meshtastic-Apple-Fork#1, which stacks this #2010 fix on top of #1983 (the channel-QR hardening refactor) instead of conflicting with it against `main`. #1983 already covers the decode-before-clear, single LoRa send, and no-UNSET-config behavior; the remaining piece — tolerating the post-reboot `wantConfig` disconnect — moves there. If #1983 doesn't land, I'll reopen/retarget this.

bruschill:
@RCGV1 opened a PR with my work against your changes 👍🏻

PR #2018, opened 2026-06-30
fix(map): stop Map tab resetting to the South Atlantic (0,0) on launch (#2016)

### LINKED ISSUE #2016: 🐞 [Bug]: iOS App, map resets to South Atlantic Ocean every time I open it
### Before submitting

### Firmware Version

2.7.15 (iOS App)

### What did you do?

Every time I open the app, the Map tab positions in the South Atlantic Ocean instead of my local region as the previous version did. I need to pan the map, then expand it to see my local region. Next time I open the app it's reset again.

### Expected Behavior

Map tab should position and zoom into local region as previous behavior.

### Current Behavior

Every time I open the app, the Map tab positions in the South Atlantic Ocean instead of my local region as the previous version did. I need to pan the map, then expand it to see my local region. Next time I open the app it's reset again.

### Participation

### Additional comments

_No response_

### Code of Conduct

bruschill:
## What changed?

Fixes #2016 — the Map tab opening centered on (0,0) (Null Island / the South Atlantic) on every launch, requiring the user to manually pan/zoom back to their region each time.

Three related fixes to the `MeshMapMK` / `ClusterMapView` map introduced in the MKMapView migration:

1. **Persist & restore the last region.** New `@AppStorage` keys store the user's viewport and `frameInitialRegionIfNeeded` restores it first, so the map reopens where it was left. Saving happens at lifecycle commit points (leaving the tab, backgrounding) rather than on every region callback, to keep UserDefaults off the pan/zoom hot path.
2. **Self-healing one-shot frame.** Camera commands stage as `pendingCameraCommand` and only commit once the map has a non-zero layout size, backstopped by `mapViewDidChangeVisibleRegion` and a small `LayoutObservingMapView` that forwards `layoutSubviews`.
3. **Follow-the-user floor.** When there's nothing to frame on (no saved region, no GPS, no connected device, no node positions), the map follows the user's location instead of sitting at (0,0). It's asserted at most once so it never fights a manual pan.

Region-persistence rules are factored into a small, testable `MeshMapRegionPersistence` helper.

## Why did it change?

The migrated MKMapView Map tab positioned its camera **exactly once** via a one-shot `ClusterMapCameraCommand`, with no persistence and no retry. If that single frame didn't land — e.g. the `setRegion` was applied before the map had a layout size (it was then silently dropped yet permanently marked applied), or there was no GPS/device/node coordinate available at that instant — the map fell back to MKMapView's default unconfigured region, whose center is (0,0): the South Atlantic. The pre-migration SwiftUI map framed continuously, so this was a regression.

## How is this tested?

- Added **10 regression tests** (`MeshMapRegionPersistenceTests`) pinning the core invariant: a (0,0)/invalid region can never be persisted or restored, including self-healing a (0,0) saved by a pre-fix build, plus valid-region round-trip and not-yet-saved cases. All pass on the iOS 18 simulator.
- App + test targets build clean.
- Reviewed the diff with a high-effort code review and addressed all findings (follow-fallback fighting manual pans, follow-mode pans not being persisted, hot-path UserDefaults churn, pending-command stranding).

N/A — behavioral fix to initial map framing.

## Checklist

- [ ] I have verified whether these changes require updates to the in-app documentation under `docs/user/` or `docs/developer/`, and updated accordingly. No user-facing doc page documents map auto-framing behavior; requesting the **`skip-docs-check`** label.

## Summary by CodeRabbit

* **New Features**
  * The map now remembers your last viewed area and restores it when you return.
  * If no saved region is available yet, the map can follow your location until it has enough data to frame.
  * Initial map framing is improved: it waits for layout to complete before applying the one-time view update.
* **Bug Fixes**
  * Improved handling of invalid saved map regions, including edge cases near (0,0) and other invalid/unsafe bounds.
  * Added safeguards to avoid unexpected viewport changes after manual interaction.
* **Tests**
  * Added coverage for region persistence validation and restoration edge cases.

### DIFF
--- a/Meshtastic/Views/Nodes/Helpers/Map/ClusterMapView.swift
+++ b/Meshtastic/Views/Nodes/Helpers/Map/ClusterMapView.swift
@@ -168,6 +168,9 @@ struct ClusterMapView<Item: Identifiable, Pin: View, Cluster: View>: UIViewRepre
 	/// One-shot programmatic camera move (e.g. the initial frame), applied exactly once per distinct
 	/// id. The ONLY thing that moves the camera programmatically; after it fires the user is in control.
 	let cameraCommand: ClusterMapCameraCommand?
+	/// Floor fallback: while no camera command has framed the map yet, follow the user's location
+	/// instead of sitting at MKMapView's default (0,0) region. A camera command supersedes this.
+	let followsUserLocationUntilFramed: Bool
 	/// When true, annotations share a `clusteringIdentifier` so MapKit collapses nearby pins.
 	let clustering: Bool
 	/// Builds the SwiftUI view for one item's pin.
@@ -204,6 +207,7 @@ struct ClusterMapView<Item: Identifiable, Pin: View, Cluster: View>: UIViewRepre
 		coordinate: @escaping (Item) -> CLLocationCoordinate2D,
 		region: Binding<MKCoordinateRegion?>? = nil,
 		cameraCommand: ClusterMapCameraCommand? = nil,
+		followsUserLocationUntilFramed: Bool = false,
 		clustering: Bool = true,
 		onSelect: ((Item) -> Void)? = nil,
 		configuration: ClusterMapConfiguration = .init(),
@@ -221,6 +225,7 @@ struct ClusterMapView<Item: Identifiable, Pin: View, Cluster: View>: UIViewRepre
 		self.coordinate = coordinate
 		self.region = region
 		self.cameraCommand = cameraCommand
+		self.followsUserLocationUntilFramed = followsUserLocationUntilFramed
 		self.clustering = clustering
 		self.onSelect = onSelect
 		self.configuration = configuration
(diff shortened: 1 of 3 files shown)


garthvh:
Thanks for digging into this — the bug is real, but this bundles three new mechanisms into the camera path we've just spent a couple of PRs stabilizing (the update re-entrancy guard, the external-region guard). Closing in favor of a much smaller fix; details below so the analysis isn't lost.

## Current design (main)

- **One-shot frame:** `frameInitialRegionIfNeeded()` centers on phone GPS → connected device position → node centroid, guarded by `didInitialFrame`. If none of the three exists yet, it returns **without consuming the one-shot** and retries on the next position refresh.
- After it fires once, **the user owns the camera** — never re-framed until the view dies (force close).

## Where (0,0) actually comes from

Nobody sets Null Island. If the user has location denied, no device connected yet, and no node positions, the frame never fires — and an `MKMapView` with no region set shows its **default world view, centered near (0,0)** — which on a phone viewport reads as "the South Atlantic." And since nothing is persisted, every launch starts there again until data shows up (or never, for a GPS-denied user with position-less nodes).

## The path forward

Of the three mechanisms here, **region persistence alone fixes the reported bug**: anyone who has ever panned the map reopens where they left it, with no dependence on GPS / device / node data at launch. That's roughly one `@AppStorage` region + restoring it as the first choice in `frameInitialRegionIfNeeded()` — a few lines, no new camera machinery.

The pending-camera-command / `LayoutObservingMapView` plumbing and the follow-the-user floor may each have merit, but they belong in their own PRs with a demonstrated repro, so the hot camera path stays simple and reviewable.

A slimmed persistence-only PR would be very welcome — happy to review it quickly.

PR #2019, opened 2026-06-30
fix(messages): refresh ack status without leaving the conversation (#2017)

### LINKED ISSUE #2017: 🐞 [Bug]: iOS app - Message status does not update
### Before submitting

### Firmware Version

2.7.15

### What did you do?

Send a message in the channel page: Channels tab > Primary Channel > Type message > Press arrow

### Expected Behavior

Message should say "Waiting to be acknowledged" followed shortly by "Acknowledged" when picked up by another node.

### Current Behavior

Message says "Waiting to be acknowledged" indefinitely unless I back out of channel and back into channel, then will say "Acknowledged". Doesn't seem to update on its own any more.

### Participation

### Additional comments

Previous version worked fine.

### Code of Conduct

bruschill:
## What changed?

The channel and direct-message lists now reload when an outgoing message's ACK resolves, so its status updates in place instead of only after the conversation view is rebuilt.

Both `ChannelMessageList` and `UserMessageList` snapshot their rows into a `@State` array and only re-fetch when a lightweight *change token* differs. That token keyed solely on the newest-message cursor `(timestamp, messageId)` and the total message count. An incoming ACK changes neither — it just flips `receivedACK` / `ackError` on an existing row — so the poll-based refresh skipped the reload.

This folds the resolved-ACK state into each list's change token as **separate `deliveredAckCount` / `erroredAckCount` tallies** (via a shared `static resolvedAckCounts(...)` helper) so the existing 5-second poll reloads the moment any message leaves the waiting state. The tallies are kept distinct rather than summed so an `errored → delivered` transition — which nets zero on a sum — still moves the token.

## Why did it change?

Fixes #2017: a sent message stayed on "Waiting to be acknowledged. . ." indefinitely and only updated to its real status after backing out of the channel and re-entering.

The ACK itself was received and persisted correctly — this is purely a UI-refresh gap. It's a regression from two compounding performance changes already on `main`:

- **#1901** replaced the reactive message list with a manual `@State` snapshot + change-token poll. The token didn't include ACK state.
- **#1914** moved packet handling onto a background `@ModelActor` and switched the routing-ACK save to a debounced write. The ACK mutates a background-context instance, so the view's main-context `@Bindable` row never observes it directly.

With the reactive binding gone *and* the ACK applied on a separate context, nothing triggered a reload until the view was recreated.

## How is this tested?

New `MessageAckStatusRefreshTests` suite (9 tests, passing on iPhone 16 / iOS 18.2): covers delivery-ack, routing-error, **errored→delivered**, retry-reset, incoming-message-not-counted, and detection-sensor exclusion across both the channel and DM paths. One test also asserts the *legacy* token signals stay unchanged on an ACK, locking in the root cause. Tests call the production `resolvedAckCounts` helper directly so the predicates can't silently drift.

Reviewed with `/code-review` at high effort, which caught the sum-vs-tallies correctness issue (fixed before this PR).

> [!NOTE]
> SwiftData's `#Predicate` macro can't type-check a compound `||` (or a fourth `&&` term) in this predicate shape without exceeding its budget, so the count is split into two single-term `fetchCount`s. `isEmoji` is intentionally not filtered for the same reason — an acked tapback costs one extra benign reload, never wrong data. Both are documented inline.

N/A — no visual change; the existing "Waiting to be acknowledged. . ." / "Acknowledged" text simply updates without re-entering the view.

## Checklist

- [x] I have verified whether these changes require updates to the in-app documentation under `docs/user/` or `docs/developer/`, and updated accordingly. No doc update needed — internal refresh logic only; please add the **`skip-docs-check`** label.

## Summary by CodeRabbit

* **Bug Fixes**
  * Message and conversation lists now refresh reliably when delivery acknowledgment status changes, even if the total message count and latest cursor remain unchanged.
  * Improves handling of both successful and errored delivery transitions (including retries) so the UI doesn’t get stuck or go stale.
  * Ensures incoming broadcast-style messages and detection-sensor direct messages don’t incorrectly affect refresh behavior.

* **Tests**
  * Added regression tests covering ACK-resolution–driven refresh behavior for both channel messages and direct messages.

### DIFF
--- a/Meshtastic.xcodeproj/project.pbxproj
+++ b/Meshtastic.xcodeproj/project.pbxproj
@@ -304,6 +304,7 @@
 		DD000029APSTFWTST0000001 /* AppSettingsAndFirmwareTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD000029APSTFWTST0000000 /* AppSettingsAndFirmwareTests.swift */; };
 		DD00002AINTREMTST00000001 /* IntentAndRemainingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD00002AINTREMTST00000000 /* IntentAndRemainingTests.swift */; };
 		DD00002BMSHPKTELTST000001 /* MeshPacketsAndTelemetryTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD00002BMSHPKTELTST000000 /* MeshPacketsAndTelemetryTests.swift */; };
+		DD00002BACKSTATUSRT000001 /* MessageAckStatusRefreshTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD00002BACKSTATUSRT000000 /* MessageAckStatusRefreshTests.swift */; };
 		DD00002DSNPVWTST000000001 /* SwiftUIViewSnapshotTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD00002DSNPVWTST000000000 /* SwiftUIViewSnapshotTests.swift */; };
 		DD00002ESDMIGTST000000001 /* SwiftDataMigrationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD00002ESDMIGTST000000000 /* SwiftDataMigrationTests.swift */; };
 		DD00002FSHRDTSTCNT000001 /* SharedTestContainer.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD00002FSHRDTSTCNT000000 /* SharedTestContainer.swift */; };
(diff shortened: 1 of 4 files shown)


garthvh (approved):
Right shape for the fix — works with the existing poll+token design instead of reintroducing cross-context observation, and the added cost is two COUNT queries per 5s for the visible conversation only (no object materialization), which is safe at large-event scale.

On the CodeRabbit offsetting-tallies threads: the per-message fingerprint it suggests would materialize every row on every poll — a real perf cost — to close a window that requires two opposite ACK transitions in the same conversation within one 5s tick, is astronomically rare in practice, and self-heals on the next token-changing event. The count-only trade-off is the right one; the token comment documents the state machine.

PR #2032, opened 2026-07-04
Fix ESP32 BLE OTA terminal OK handling

RCGV1:
## Summary
- Handle ESP32 BLE OTA devices that ACK the final data chunk, then send the terminal OK after verification.
- Keep premature OK and ERR responses as failures.
- Move ACK/OK/ERR parsing into a small ESP32OTAProtocol helper with focused tests.
- Update the firmware update docs and regenerated in-app docs bundle for the revised completion behavior.

## Repro
Some ESP32 BLE OTA status streams can end the data phase with ACK for the final chunk, then send OK only after firmware verification. Before this change, the final ACK advanced offset to the firmware size, the upload loop exited, and the view model reported `Stream ended without OK` without waiting for that terminal OK.

## Validation
- `MeshtasticTests/ESP32BLEOTATerminalResponseTests`: 9 passed.
- Simulator build for scheme `Meshtastic` on iPhone 17 succeeded.
- GitHub checks: docs coverage, SwiftLint, CodeRabbit, and CLA passed.


garthvh (changes requested):
Requesting changes.

To set expectations up front: we're not looking to take "hardening" PRs into this app. We don't want broad reliability sweeps over the OTA flows — every change to a flashing path needs to be tied to a specific, reproducible failure we can point at, not a general "make it more robust" pass. A PR framed as hardening is very hard to review and inevitably mixes a real fix with speculative churn, which is what's happening here. Please drop the hardening framing entirely.

Concretely, this needs to be reworked before it can move forward:

**Scope — split it up.** This is three unrelated changes under one label:

- The `FirmwareUpdateNotifier` / `FirmwareUpdateNotificationPolicy` / Settings work is a new user-facing feature (a "your firmware is behind" prompt), not OTA reliability, and it's the majority of the diff. That's a product decision that has to be discussed and evaluated on its own — pull it into a separate PR.
- The ESP32 BLE OTA changes and the nRF DFU changes are different flows and should be separate, independently justified PRs.

**Description — it over-claims.** Rewrite the summary to state exactly what changed and why, each item tied to a concrete repro:

- The nRF summary says this moves to a "nonzero packet receipt notification interval," but the interval was already 8 — the actual change is 8 → 4. Either justify why 4 beats the value that already works, with the specific failure it fixes, or drop it. Lowering a working PRN changes transfer behavior and shouldn't move on a hunch.
- Hardware-bench results we can't reproduce or see aren't something anyone here can review against. Keep the description to changes that ship with a repro a reviewer can actually follow.

**Code concern:**

- `NRFDFUFailureClassifier` matches on raw substrings of the DFU library's log text ("operation failed", "100306", "receive firmware image"). That silently breaks the moment the library rewords a message. If a friendlier error is worth having, key it off a stable error code/type, not log strings — otherwise drop it.

The one piece that looks genuinely worthwhile is the ESP32 BLE terminal-response handling (waiting for the final OK after the last ACK) and moving the ACK/OK/ERR decision into a testable helper. Split that out on its own, tied to the specific transfer failure it fixes, and it's reviewable.

Thanks.

garthvh (dismissed):
Duplicate of the review above (accidental double-submit) — dismissed. See the other review for the actual feedback.

garthvh (approved):
Approved after review.

**Logic:** Correctly fixes the ESP32 BLE OTA case where the device ACKs the final data chunk and sends the terminal `OK` only after firmware verification — previously the loop exited on the final ACK and reported "Stream ended without OK". ACK/OK/ERR parsing is cleanly extracted into a tested `ESP32OTAProtocol` helper; premature `OK` and `ERR` still fail; the terminal-`OK` wait is bounded by a 30s timeout and handles disconnect. Covered by the added tests.

**Conflict resolution:** This branch was 137 commits behind `main` and conflicted only in the generated `Meshtastic/Resources/docs/index.json`. I merged current `main` and regenerated the docs bundle via `scripts/build-docs.sh` (cmark-gfm 0.29.0.gfm.13) rather than hand-merging the frequency-ordered keyword arrays — regeneration touched only `index.json` (no HTML diverged), JSON validated, no `charCount:0`. Net diff vs `main` is exactly the 7 intended PR files.

Note: not build-verified locally (no build/test CI in the repo; local build gated by watchOS runtime). Relying on the author's passing ESP32 protocol tests + the conflict-free auto-merge of the logic files.

PR #2050, opened 2026-07-06
fix: muted ("Hide Alerts") channels & DMs must not badge the app icon or Messages tab

rancur:
## Summary

A channel or direct-message conversation set to **"Hide Alerts" (muted)** was still badging both the **home-screen app icon** and the **in-app Messages tab** when a new message arrived. Expected iOS "Hide Alerts" semantics are that a muted conversation is completely silent — no notification **and** no badge.

The incoming-message → notification path already respects mute (`MeshPackets.textMessageAppPacket` gates the DM notification on `fromUser.mute` and the channel notification on the channel's `mute`), so notifications were already suppressed. The **unread/badge count tallies were not** filtering muted conversations, so muted channels/DMs kept incrementing the counts that drive **every** badge.

## Root cause

Both badges are fed from `AppState`:
- **App-icon badge** — `AppState` mirrors `unreadChannelMessages + unreadDirectMessages` into `UNUserNotificationCenter.setBadgeCount(...)`.
- **In-app Messages tab badges** — `ContentView` / `Messages` bind `.badge(...)` to `appState.totalUnreadMessages` / `unreadChannelMessages` / `unreadDirectMessages`.

Those two `AppState` counts are computed in four places, **none of which excluded muted conversations**:

1. `AppState.refreshBadgeCount(context:)` — master refresh on app-active / after bulk read.
2. `MyInfoEntity.unreadMessages(context:)` — channel-total helper (feeds the MQTT-path channel count).
3. `ChannelEntity.unreadMessages(context:)` — per-channel unread.
4. `UserEntity.unreadMessages(context:skipLastMessageCheck:)` — per-DM / connected-user DM aggregate (feeds `UserMessageList`, the MQTT path, and the incremental update in `textMessageAppPacket`).

## Fix

Exclude muted conversations from all four count sources, leaving **unmuted** counts unchanged:

- **`AppState.refreshBadgeCount`** — build the set of muted channel indexes and drop unread whose `channel` is in it; drop DMs whose counterpart (`fromUser`) is muted.
- **`ChannelEntity.unreadMessages`** — a muted channel returns `0`.
- **`MyInfoEntity.unreadMessages`** — exclude messages on muted channel indexes.
- **`UserEntity.unreadMessages`** — a muted conversation returns `0`; the connected-user aggregate additionally subtracts each muted counterpart's unread contribution so muted DMs never badge.

The mute join is filtered in Swift rather than pushed into the `#Predicate`, matching the existing pattern in these files (the code comments already note that comparing an optional relationship in a `#Predicate` crashes SwiftData on iOS 26). The notification-post path was already correct and is left untouched.

A muted channel/DM's messages are still stored and still visible in the thread (and still marked read normally) — they just no longer contribute to any badge, exactly like iOS "Hide Alerts".

## Testing

**Statically verified (headless, no simulator available in this environment):**
- Traced every badge/unread source. Both the app-icon badge and both in-app tab badges resolve to `AppState.unreadChannelMessages` / `unreadDirectMessages`, which are only ever set from the four functions patched here — so filtering at these sites covers both badges.
- Confirmed a muted channel/DM now contributes 0 at each site, and that unmuted channels/DMs are untouched (the added predicate/filter only removes muted rows).
- All changed Swift files pass `swiftc -parse` cleanly.
- Added a unit test (`ChannelEntityTests.testMutedChannelReportsZeroUnread`) asserting a muted channel reports 0 unread, an unmuted channel still counts, and un-muting restores the count.

**Not verified here (no Xcode.app / iOS SDK in this environment — only Command Line Tools), please confirm in a simulator/on-device:**
- Full `xcodebuild`/type-check compile of the app target and running the new unit test.
- End-to-end behavior: mute a channel via "Hide Alerts", receive a message on it → **no** app-icon badge, **no** Messages-tab badge, **no** notification; then confirm an **unmuted** channel still badges + notifies normally. Same check for a muted vs unmuted DM.

I was honest above about the compile/runtime steps I could not run headlessly — happy to adjust if a maintainer's simulator run surfaces anything.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed badge and notification badge counts to exclude muted conversations (muted channels and muted direct messages).
  * In-app unread indicators now correctly reflect unread messages even for muted channels.
  * Unread direct-message counts are adjusted to ignore muted counterpart accounts and ensure counts don’t go negative.
  * Badge counts update immediately when a channel is muted or unmuted.

### DIFF
--- a/Meshtastic/AppState.swift
+++ b/Meshtastic/AppState.swift
@@ -46,8 +46,18 @@ class AppState: ObservableObject {
 			}
 		)
 		let unread = (try? context.fetch(unreadDescriptor)) ?? []
-		let channelCount = unread.filter { $0.toUser == nil }.count
-		let dmCount = unread.filter { $0.toUser != nil && !$0.admin }.count
+		// A muted ("Hide Alerts") channel or DM conversation is silent — exclude its unread
+		// messages from every badge count (app icon + in-app Messages tab). A channel message is
+		// muted when its channel index is muted; a DM is muted when the other party (fromUser) is
+		// muted. Fetch the muted channel indexes once and filter in Swift to match the existing
+		// pattern (comparing an optional relationship to nil in a #Predicate crashes SwiftData on
+		// iOS 26, so the mute join stays out of the predicate too).
+		let mutedChannelDescriptor = FetchDescriptor<ChannelEntity>(
+			predicate: #Predicate<ChannelEntity> { $0.mute == true }
+		)
+		let mutedChannelIndexes = Set(((try? context.fetch(mutedChannelDescriptor)) ?? []).map { $0.index })
+		let channelCount = unread.filter { $0.toUser == nil && !mutedChannelIndexes.contains($0.channel) }.count
+		let dmCount = unread.filter { $0.toUser != nil && !$0.admin && !($0.fromUser?.mute ?? false) }.count
 		if unreadChannelMessages != channelCount {
 			unreadChannelMessages = channelCount
 		}
--- a/Meshtastic/Extensions/SwiftData/ChannelEntityExtension.swift
+++ b/Meshtastic/Extensions/SwiftData/ChannelEntityExtension.swift
@@ -43,6 +43,9 @@ extension ChannelEntity {
 
 	@MainActor
 	func unreadMessages(context: ModelContext) -> Int {
+		// A muted ("Hide Alerts") channel is silent: it never contributes to the unread/badge
+		// counts that drive the app icon badge and the in-app Messages tab badge.
+		guard !self.mute else { return 0 }
 		let channelIndex = self.index
 		let descriptor = FetchDescriptor<MessageEntity>(
 			predicate: #Predicate<MessageEntity> { msg in
--- a/Meshtastic/Extensions/SwiftData/ChannelEntityExtension.swift
+++ b/Meshtastic/Extensions/SwiftData/ChannelEntityExtension.swift
@@ -43,9 +43,12 @@ extension ChannelEntity {
 
 	@MainActor
 	func unreadMessages(context: ModelContext) -> Int {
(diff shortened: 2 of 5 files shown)


rancur:
@coderabbitai Good catch — addressed in 8b2d868.

The mute check has been removed from the shared `ChannelEntity.unreadMessages` helper, which drives the in-app `ChannelList` unread indicator. That helper is now deliberately mute-agnostic and reports the channel's true unread count, so a muted ("Hide Alerts") channel still shows unread **inside the app**.

The badge itself still excludes muted channels: that filtering lives entirely in the badge-aggregation path — `AppState.refreshBadgeCount` and `MyInfoEntity.unreadMessages`, the two sources that feed `appState.unreadChannelMessages`. `ChannelEntity.unreadMessages` never fed the badge, so dropping its mute check doesn't regress the badge fix. (`MyInfoEntity`/`UserEntity.unreadMessages` are badge-only aggregates, not used by any in-app per-conversation indicator, so they're left as-is.)

Tests: `testMutedChannelReportsZeroUnread` asserted the now-incorrect behavior, so it's replaced by:
- `testMutedChannelStillReportsUnreadInApp` — muted channel still reports its real unread via the helper (in-app survives).
- `testBadgeCountExcludesMutedChannel` — `AppState.refreshBadgeCount` drops the muted channel's unread, asserted as a mute-toggle delta so it's robust against the shared in-memory test container.

All changed files pass `swiftc -parse`; there's no Xcode/simulator in this environment (Command Line Tools only), so the full type-check + running the unit tests relies on CI.

(Re: the Swift Testing nitpick on the test file — left as XCTest to stay consistent with the two existing XCTest cases in this same file; happy to migrate the file to Swift Testing separately if preferred.)

garthvh:
This correctly mirrors iMessage already

PR #2055, opened 2026-07-07
perf: move node-list filtering off the main thread

bruschill:
## What changed?

Moves node-list filtering off the main thread. `FilteredNodeList` previously ran an unconditional 350 ms poll that recomputed the filtered/sorted list on the **main actor** — an O(N) scan doing 6 relationship traversals (each a potential SwiftData fault) plus locale-aware string compares per node, ~3×/sec, forever.

The poll is replaced with a change-driven, coalesced recompute:

- The scan (`computeNodeOrder`) runs **off the main actor** on a `Task.detached` with a fresh, confined `ModelContext`, returning only Sendable `[Int64]` node nums.
- Those nums are mapped back onto the live `@Query` objects and partitioned into connected → favorites → regular **on the main actor from current state**, so no background-context model ever reaches the `List` (preserving the existing SIGTRAP hardening against faulting pruned rows).
- A single long-lived consumer drains a `bufferingNewest(1)` stream: filter/search edits and node-set changes fire an **instant** recompute; a reconnect nodeDB dump collapses to one pending pass (300 ms throttle → ~3/sec cap). A 1.5 s **off-main** backstop refreshes the inputs nothing notifies us about (GPS location with the distance filter, the online-window wall-clock crossing, non-top `lastHeard` reorders, favorite toggles).
- One `makeNodeListPredicate` source feeds both the `@Query` and the off-main fetch, so the two node sets can't drift.
- First paint shows a loading indicator instead of flashing an empty "Nodes (0)" list during the initial off-main fetch.

## Why did it change?

A user reported multi-second keypress lag and **dropped** keystrokes when searching the node list on a large mesh, worsening the longer the app ran. That signature — dropped input events — is the main thread being seized synchronously while UIKit tries to service the keyboard.

Profiling (iPhone 16 Pro Max simulator, seeded DBs) confirmed the poll's per-tick scan on the main actor:

| Nodes | Cold first-keystroke tick | Warm steady-state tick |
|------:|--------------------------:|-----------------------:|
| 1,000 | 113 ms | 11 ms |
| 5,000 | **514 ms** | 52 ms |

The scan is cleanly O(N) (~10.6 µs/node warm, ~100 µs/node cold from faulting `user`), and with `purgeStaleNodeDays` defaulting to 0 the node set grows unbounded — so it strictly worsens over time, on a device 3–4× slower than the sim. Running the identical scan on a `Task.detached` removes it from the main thread entirely; typing stays responsive regardless of the scan's absolute cost.

## How is this tested?

- Built (Debug) and run on the iPhone 16 Pro Max simulator against a 5,000-node seeded store (75k telemetry / 25k position / ~400 message rows). Node list renders correctly (4,948 after the 52 ignored are filtered), rows fully populated, no SIGTRAP.
- Compiles under Swift 6 strict concurrency (Sendable snapshot in, `[Int64]` out; the confined `ModelContext` never escapes the synchronous `computeNodeOrder`).
- Verified filter-match parity: the refactored predicate + post-predicate produce the same result count as the previous in-memory path.
- Reviewed for the change-detection regressions inherent in replacing a brute-force poll (device location, online-window crossing, favorite re-sectioning, non-top reorders) and addressed them via the 1.5 s off-main backstop + on-main partition.

Note: `relationshipKeyPathsForPrefetching = [\.user]` was evaluated as a way to cut the cold fault cost but measured no effect on this platform (SwiftData to-one prefetch appears ineffective), so it is not included. The durable cold-latency win — denormalizing a lowercased `searchIndex` and pushing search into the `#Predicate` so SQLite filters — is a separate follow-up (needs a migration).

_Node list rendering 4,948 nodes from a 5,000-node seeded DB via the new off-main path (available on request)._

## Checklist

- [ ] I have verified whether these changes require updates to the in-app documentation under `docs/user/` or `docs/developer/`. This is an internal performance refactor of the node list with no user-facing behavior change, so no doc update is needed — requesting the **`skip-docs-check`** label.

## Summary by CodeRabbit

* **New Features**
  * Node lists now refresh using an off-main, change-driven pipeline for smoother filtering and ordering.
  * A brief loading indicator may appear while the first refreshed results are computed.

* **Bug Fixes**
  * Improved filter consistency so visible nodes more reliably match the selected criteria, including online and distance-related effects.
  * Reduced unnecessary polling while keeping results timely with an automatic backstop refresh.

* **Tests**
  * Added unit tests covering the node list filtering and ordering behavior, including favorite/search/online semantics.

### DIFF
--- a/Meshtastic.xcodeproj/project.pbxproj
+++ b/Meshtastic.xcodeproj/project.pbxproj
@@ -8,6 +8,7 @@
 
 /* Begin PBXBuildFile section */
 		0516B3FE2F68892000D0FC40 /* NodeFilterParametersTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 05DCA1AE2F646B3B00D0724C /* NodeFilterParametersTests.swift */; };
+		DDFF00000000000000000001 /* NodeListFilteringTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = CCFF00000000000000000001 /* NodeListFilteringTests.swift */; };
 		DD0000A10EDD5A00000000B1 /* XEdDSASigningTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD0000A20EDD5A00000000B2 /* XEdDSASigningTests.swift */; };
 		05A1B2C32F70000100D0FC40 /* SettingsNodeSortTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 05A1B2C42F70000100D0FC40 /* SettingsNodeSortTests.swift */; };
 		0B9642EC0DB1B971CB26E8AE /* RadarSweepView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4E06A9BCC789BD452DFA9CFB /* RadarSweepView.swift */; };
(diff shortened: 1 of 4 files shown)


garthvh:
Closing for now — Datadog RUM for 2.7.16 shows no crash or memory-pressure signal that supports prioritizing this change. The main-thread-hang/dropped-keystroke issue itself isn't observable in our current telemetry either (long-task/app-hang tracking is disabled), so we can't confirm how widespread it is beyond the one report.

This also isn't an iOS 18 fix: the SwiftData crash signal in this area skews toward iOS 26.x, and iOS 18 devices are a smaller share of those crashes than of the install base overall.

Happy to revisit with real telemetry (e.g. app-hang tracking) or a reproducible iOS 18 report.

PR #2062, opened 2026-07-09
Consolidate FromRadio decode and UTF-8 validation across transports

thebentern:
## Summary

Each transport decoded `FromRadio` frames on its own, with inconsistent handling when a protobuf string field (e.g. a node's `long_name`) contains an invalid UTF-8 sequence that SwiftProtobuf rejects at decode time:

- **TCP** tore down the connection (`disconnect(shouldReconnect: true)`) on *any* decode miss, so a single unparseable string field would drop an otherwise healthy stream.
- **BLE** already skipped invalid-UTF-8 frames, but via bespoke inline handling.
- **Serial** skipped too, but logged a generic "failed to deserialize" message.

This consolidates all three onto one decode + validation path, `FromRadioDecoder.classify(_:)`, which returns exactly one outcome:

- `.decoded(FromRadio)` — yield to the `AccessoryManager`
- `.skipInvalidUTF8(Error)` — a string field failed UTF-8 validation: skip the frame, keep reading
- `.failed(Error)` — genuine framing/wire corruption: BLE/TCP disconnect + reconnect; Serial skips

So a malformed string encoding is handled identically across transports, and the skip-vs-reconnect policy lives in one place instead of drifting per transport.

## Changes

- Add `FromRadioDecoder` + `FromRadioDecodeOutcome` in `Connection.swift`.
- Route `BLEConnection`, `TCPConnection`, and `SerialConnection` through the shared classifier.
- TCP no longer drops the connection over an invalid-UTF-8 field; genuine decode/framing failures still reconnect.
- Register `InvalidUTF8Tests` in the `MeshtasticTests` target — it was committed but never linked into the target, so it wasn't running — and add classifier coverage for the skip / decode / fail outcomes.

## Testing

```
  -destination 'platform=iOS Simulator,name=iPhone 17' \
  -only-testing:MeshtasticTests/InvalidUTF8Tests
```

All 8 tests pass (`** TEST SUCCEEDED **`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated incoming frame decoding to treat protobuf invalid UTF-8 in string fields as non-fatal, skipping those frames instead of disconnecting.
  * Kept normal message delivery uninterrupted while continuing to disconnect/reconnect on genuinely corrupted or unreadable payloads.
  * Applied consistent decoding behavior across Bluetooth Low Energy, Serial, and TCP transports.

* **Tests**
  * Added/expanded tests ensuring frames are classified as skipped for invalid UTF-8, decoded for valid payloads, and failed for malformed protobuf inputs.

* **Documentation**
  * Documented the shared inbound frame decoding outcomes and handling semantics.

### DIFF
--- a/Meshtastic.xcodeproj/project.pbxproj
+++ b/Meshtastic.xcodeproj/project.pbxproj
@@ -9,6 +9,7 @@
 /* Begin PBXBuildFile section */
 		0516B3FE2F68892000D0FC40 /* NodeFilterParametersTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 05DCA1AE2F646B3B00D0724C /* NodeFilterParametersTests.swift */; };
 		DD0000A10EDD5A00000000B1 /* XEdDSASigningTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD0000A20EDD5A00000000B2 /* XEdDSASigningTests.swift */; };
+		DD0000C10EDD5A00000000C1 /* InvalidUTF8Tests.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD0000C20EDD5A00000000C2 /* InvalidUTF8Tests.swift */; };
 		05A1B2C32F70000100D0FC40 /* SettingsNodeSortTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 05A1B2C42F70000100D0FC40 /* SettingsNodeSortTests.swift */; };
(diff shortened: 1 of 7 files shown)


thebentern:
This is a follow-up to my earlier fix: #1687

thebentern:
Good call — documented the shared `FromRadioDecoder.classify` funnel and the invalid-UTF-8 skip vs. reconnect semantics in `docs/developer/transport.md` (8d77edb).

PR #2086, opened 2026-07-13
feat(map): let a waypoint be sent as a DM or to a specific channel

joeyleake:
## What changed?

Adds a **"Send to"** picker to the waypoint editor. A waypoint can now be sent as a broadcast on the primary channel (shown by its real configured name, e.g. "LongFast"), broadcast on any secondary channel, or sent as a private direct message to a specific node — matching the destinations already supported by the Meshtastic API.

Key pieces:
- `WaypointDestination` enum + a `destination` property persisted on `WaypointEntity`, so re-opening an existing waypoint for editing, or resending it (e.g. "delete for everyone"), always routes back to wherever it was originally exchanged instead of drifting to broadcast.
- `waypointDestination(packet:myNodeNum:)` disambiguates a DM we sent from a DM sent to us (a waypoint DM'd *to* us has `packet.to` set to our own node — reading it unconditionally would misroute follow-ups back to ourselves).
- DM routing reuses the same PKC-vs-fallback check regular direct messages already use, extracted into a shared `applyPKCRoutingIfNeeded` helper so the two send paths can't drift apart.
- New `WaypointRecipientPicker` view: channels first (deduped/index-sorted, sharing its logic with the Channels settings screen so the two never show different names for the same channel), then a filterable node list — filterable since node lists can run into the thousands on an MQTT-bridged mesh.

## Why did it change?

Waypoints previously always broadcast on the primary channel, unlike text messages, which can already be sent to any channel or DM'd to a specific node. This closes that gap.

## How is this tested?

- New Swift Testing suites in `WaypointDestinationTests.swift`: packet-to-destination derivation (including the sent-vs-received DM disambiguation), `WaypointEntity.destination` round-trip, recipient/channel label fallbacks, and recipient-picker filtering.
- Verified a full build + full test suite run on a real Xcode/iOS Simulator CI job; all new tests pass. (A handful of pre-existing, unrelated failures — snapshot-image drift and flaky async-timer tests — reproduce identically against a clean `main` baseline, confirmed separately, so they're not from this change.)
- Manually sideloaded onto a physical iPhone and exercised both the send-to-channel and send-to-node-DM flows end to end.

## Screenshots/Videos
wp_00

TEMP_waypointRecipientPicker

## Checklist

- [x] I have verified whether these changes require updates to the in-app documentation under `docs/user/` or `docs/developer/`, and updated accordingly. (`docs/user/map.md` updated.)

## Summary by CodeRabbit

* **New Features**
  * Added “Send to” options for waypoints, supporting channels or specific nodes.
  * Added recipient selection with searchable node and channel lists.
  * Preserved waypoint destinations when editing, deleting, importing, or receiving waypoints.
  * Improved channel names and recipient labels with clearer fallback handling.

* **Bug Fixes**
  * Prevented waypoint processing when no active device is connected.

* **Documentation**
  * Updated waypoint guidance for choosing recipients and managing edits or deletions.

### DIFF
--- a/Meshtastic.xcodeproj/project.pbxproj
+++ b/Meshtastic.xcodeproj/project.pbxproj
@@ -382,6 +382,9 @@
 		DD457188293C7E63000C49FB /* BLESignalStrengthIndicator.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD457187293C7E63000C49FB /* BLESignalStrengthIndicator.swift */; };
 		DD4640202AFF10F4002A5ECB /* WaypointForm.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD46401F2AFF10F4002A5ECB /* WaypointForm.swift */; };
 		AABBCCDD0000000000000005 /* GeofenceBoundsSelectorView.swift in Sources */ = {isa = PBXBuildFile; fileRef = AABBCCDD0000000000000006 /* GeofenceBoundsSelectorView.swift */; };
+		AABBCCDD0000000000000007 /* WaypointDestination.swift in Sources */ = {isa = PBXBuildFile; fileRef = AABBCCDD0000000000000008 /* WaypointDestination.swift */; };
+		AABBCCDD0000000000000009 /* WaypointRecipientPicker.swift in Sources */ = {isa = PBXBuildFile; fileRef = AABBCCDD000000000000000A /* WaypointRecipientPicker.swift */; };
+		AABBCCDD000000000000000B /* WaypointDestinationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = AABBCCDD000000000000000C /* WaypointDestinationTests.swift */; };
 		DD4756AB2E9F1A0000029299 /* SecurityVersionNag.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD4756AA2E9F1A0000029299 /* SecurityVersionNag.swift */; };
 		DD47E3D626F17ED900029299 /* CircleText.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD47E3D526F17ED900029299 /* CircleText.swift */; };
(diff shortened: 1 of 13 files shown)


joeyleake:
@coderabbitai full review

joeyleake:
Thanks for the review! Fixed in cf9c2b74:
- `AccessoryManager.swift`: `UInt32(connectedNodeNum)` → `UInt32(truncatingIfNeeded:)`, matching the convention used elsewhere in this PR.
- `AccessoryManager+ToRadio.swift`: same fix for `sendWaypoint`'s channel-index conversion — good catch on the chain reaction from the earlier `truncatingIfNeeded` fix on the receive side, which could otherwise plant a negative `destinationChannel` that traps on a later resend.
- `Channels.swift`: `normalizeDuplicateChannelsIfNeeded()` now uses `dedupedChannels(for:)` instead of its own copy.
- Added test coverage for `channelDisplayName`'s "Custom" and unresolved-preset-fallback branches with a node present.

Skipping the search-filter debounce suggestion for now — it's a real but currently unmeasured concern, and adding debounce/cancellation logic is more complexity than seems warranted for a "trivial"-rated nitpick. Happy to revisit if it turns out to matter in practice.

RCGV1:
This needs a better UX instead of a big list of channels and nodes

joeyleake:
> This needs a better UX instead of a big list of channels and nodes

Thanks for the feedback. I pushed a change that labels the two groups ("Channels" / "Direct Messages") so the split is clear at a glance instead of just whitespace. I updated the initial post with a screenshot.

On the broader structure: I went with a single searchable list rather than a Messages-style category-first flow on purpose. The Messages tab's Channels/Direct Messages split makes sense there because it's a persistent surface you browse repeatedly. This picker is the opposite, a one-time "pick a destination and go" choice made once per waypoint, and you almost always already know whether you're sending to a channel or a specific person before opening it. The search box handles the "big list" concern directly so even on a large mesh you're never actually scrolling a big list to find someone.

Happy to iterate further if there's something specific still feeling off, genuinely want to get this right.

garthvh:
The firmware should be updated to use the new channel position logic to determine the channel

RCGV1:
@garthvh This should be reopened, waypoints have nothing to do with where position is sent. Waypoints are like any other chat message.

PR #2114, opened 2026-07-18
ci(sync-device-svgs): refresh the Resources/images catalog the app actually loads

bruschill:
## What changed?

Retargets the nightly **Sync Device SVGs** workflow (`.github/workflows/sync_device_svgs.yml`) so it refreshes the device catalog the app actually loads.

- Replaces the old "clone `meshtastic/web-flasher` → svgo → write imagesets into `Meshtastic/Assets.xcassets`" steps with a single call to the repo's own **`scripts/download_images.py --force`**, writing to **`Meshtastic/Resources/images/`** + **`Meshtastic/Resources/DeviceHardware.json`** — the same script and output location the Xcode **"Download Node SVG Images"** build phase uses.
- Adds explicit **`permissions: contents: write`** so the scheduled job can actually commit/push the refresh.
- Stages new/updated/pruned device SVGs via `git add -A`, and detects changes (including brand-new untracked SVGs) with a path-scoped `git status --porcelain`.

Net: +37 / −127 (removes the now-unnecessary web-flasher clone, svgo install, and xcassets generation).

## Why did it change?

The app loads device images at runtime from the **bundled `images` subdirectory** and **`DeviceHardware.json`** (`MeshtasticAPI.swift` — `Bundle.main.url(forResource:withExtension:subdirectory:"images")` and the bundled hardware catalog), maintained by `scripts/download_images.py`. The device imagesets under `Assets.xcassets` are **no longer the runtime source** (the remaining imagesets there are AppIcons). So the nightly job was updating a legacy location and **not** keeping the real catalog current — it drifted until someone happened to run a clean build (which force-refreshes the catalog) and committed the result.

Concretely, this catalog is currently behind upstream: a new **RAK6421** device is present in the API but missing from the committed snapshot. Pointing the automation at the correct script + path means the bundled catalog stays in sync with `https://api.meshtastic.org/resource/deviceHardware` automatically, and contributors stop seeing spurious `DeviceHardware.json` / `image_manifest.json` diffs after clean builds.

## How is this tested?

- `python3 -c "import yaml; yaml.safe_load(...)"` — workflow YAML validates.
- `scripts/download_images.py` is the exact script already run by the Xcode build phase (unchanged), so the workflow reproduces the same output the app bundle expects; running it locally with these args produces the expected catalog refresh (e.g. adds `rak6421.svg` + its manifest/JSON entries).
- The job is `schedule` + `workflow_dispatch` only (no untrusted/fork input), and `contents: write` is the minimum permission required for the auto-commit. It can be dry-run via **workflow_dispatch** after merge to confirm the commit/push path.

N/A — CI workflow change only.

## Checklist

- [x] I have verified whether these changes require updates to the in-app documentation under `docs/user/` or `docs/developer/`. This is a CI-only change with no `Meshtastic/Views/**` or `@Model` changes, so no doc update is required — adding the **`skip-docs-check`** label.

## Summary by CodeRabbit

* **Chores**
  * Reworked the automated device resource synchronization process.
  * Device images and the hardware catalog are now updated and pushed directly when changes are detected.
  * Improved workflow logging to clearly report whether device resources were updated.

### DIFF
--- a/.github/workflows/sync_device_svgs.yml
+++ b/.github/workflows/sync_device_svgs.yml
@@ -7,155 +7,65 @@ on:
   workflow_dispatch:
     # Allow manual triggering
 
+# Needed so the scheduled run can commit the refreshed catalog back to the repo.
+permissions:
+  contents: write
+
 jobs:
   sync-device-svgs:
     runs-on: ubuntu-latest
-    
+
     steps:
     - name: Checkout repository
       uses: actions/checkout@v4
       with:
         token: ${{ secrets.GITHUB_TOKEN }}
-        
-    - name: Setup Node.js
-      uses: actions/setup-node@v4
+
+    - name: Set up Python
+      uses: actions/setup-python@v5
       with:
-        node-version: '18'
-        
-    - name: Install dependencies
+        python-version: '3.11'
+
+    - name: Sync device images and hardware catalog
+      # Reuse the exact script the Xcode "Download Node SVG Images" build phase runs,
+      # writing to the bundled resource location the app actually loads at runtime
+      # (the Bundle "images" subdirectory + DeviceHardware.json — see MeshtasticAPI).
+      # --force ignores the cached API hash so the committed snapshot always matches
+      # the upstream device hardware API instead of drifting until someone clean-builds.
       run: |
-        npm install -g svgo
-        
-    - name: Download and process SVGs
-      run: |
-        #!/bin/bash
-        set -e
-        
-        # Create temporary directory
-        mkdir -p temp_svgs
-        cd temp_svgs
-        
-        # Clone web-flasher repo (shallow clone for speed)
-        git clone --depth 1 https://github.com/meshtastic/web-flasher.git
-        
-        # Navigate to SVG directory
-        cd web-flasher/public/img/devices
-        
-        # Create output directory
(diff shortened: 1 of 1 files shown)


bruschill:
Thanks @coderabbitai — addressed both findings in 8fe78443:

**1. Detached HEAD push (major).** `actions/checkout` leaves a detached HEAD, so the bare `git push` would fail. Now pushes HEAD explicitly to the branch:
```
git push origin "HEAD:${TARGET_BRANCH}"
```

**2. Expression expansion inside bash (nitpick).** Both `github.ref_name` (for the push) and the `has_changes` output (Summary step) are now read through step-level `env:` vars instead of inline `${{ }}`, so no GitHub expression is expanded inside a shell script.

**Operational note (branch protection):** if `main` is protected, the scheduled job's direct push via `secrets.GITHUB_TOKEN` may still be rejected regardless of the detached-HEAD fix — this pre-dates the PR (the previous workflow also pushed directly to `main`). If maintainers prefer, I can switch the commit step to open a PR via `peter-evans/create-pull-request` (the original file already had this as a disabled alternative), which sidesteps branch protection and keeps the refresh reviewable. Happy to do that if that's the preferred model — otherwise this needs a PAT/GitHub App token with bypass configured on the repo side.

PR #2115, opened 2026-07-18
fix(map): restore last-viewed region on launch (fixes (0,0) cold-open)

bruschill:
## What & why

Follow-up to the "path forward" from @garthvh's review of #2018.

The Mesh Map's one-shot `frameInitialRegionIfNeeded()` centers on phone GPS → connected device position → node centroid, guarded by `didInitialFrame`. A **location-denied user with no connected device and no positioned nodes** hits none of the three, so the frame never fires and `MKMapView` falls back to its default world view centered near **(0,0)** — the "South Atlantic / Null Island" cold-open — on **every** launch, since nothing was ever persisted.

Per the agreed path forward, this is the **region-persistence-only** fix — no pending-camera-command plumbing, no follow-the-user floor, no new camera machinery:

- Persist the last-viewed `MKCoordinateRegion` via `@AppStorage` (four `Double`s, since `MKCoordinateRegion` isn't `Codable`).
- Restore it as the **first choice** in `frameInitialRegionIfNeeded()`, before the GPS/device/centroid path. Anyone who has ever moved the map reopens exactly where they left it, independent of GPS/device/node data. It consumes the one-shot, so the user still owns the camera immediately after.

## Not persisting the bug

Saving happens on pan/zoom settle (the existing `regionRefreshKey` `onChange`), guarded so the degenerate default is never written back:

- **Whole-world guard** — a region must span `< 90°` latitude **and** `< 180°` longitude; you can't fit the entire world under half-extent on both axes, so MapKit's settled default is always rejected.
- **Null-Island guard** — a *broad* span still centered within 1° of (0,0) is rejected too (the intermediate pre-interaction default), while a genuine small-span local view near (0,0) still saves.
- Restore re-validates finiteness, coordinate range, and positive span; a non-positive saved span is the "nothing saved yet" sentinel.

## Behavior

- Fresh install (nothing saved): unchanged — frames current location → connected radio → node centroid.
- After you've moved the map: reopens your last-viewed area and zoom on next launch.

Analogous to the already-documented "Filters are remembered between launches" (nodes.md).

## Docs

Documented the behavior in `docs/user/map.md` and regenerated the bundled docs (`scripts/build-docs.sh`); generated HTML/JSON not hand-edited.

## Testing

- SourceKit-LSP diagnostics clean for `MeshMapMK.swift`; SwiftLint clean for the new code (only the pre-existing `type_body_length` on the struct).
- Reviewed end-to-end through the `ClusterMapView` camera coordinator: the one-shot `cameraCommand` id-dedup + `isApplyingExternalRegion` feedback guard mean restore correctly consumes the one-shot.
- Manually walked the persist/restore guards for NaN / ±inf / out-of-range / whole-world inputs — all rejected on both write and read.

Note: the persist/restore validation currently lives as `private` methods on the view; a small extraction into a testable helper (to pin the accept/reject boundaries in `MeshtasticTests`) is a reasonable follow-up if desired.

## Summary by CodeRabbit

* **New Features**
  * The Map now reopens to the last viewed area and zoom level between launches by restoring the previously visible camera region.
  * On first use, the map frames in priority order: current location → connected radio → center of the nodes (with safe handling for uninitialized camera data).
* **Documentation**
  * Updated Map & Waypoints help to explain the restore behavior and the initial framing sequence on a fresh install.


garthvh (approved):
Approved. Region-persistence fix for the (0,0) South Atlantic cold-open (issue #2016) — persists the last-viewed MKCoordinateRegion and restores it first in frameInitialRegionIfNeeded(), independent of GPS/device/node availability. Persist/restore guards correctly reject NaN/±inf, out-of-range coords, whole-world spans, and MapKit's intermediate near-(0,0) default, so the degenerate view is never written back. Verified the fix is still needed (main's framing still bails to MapKit's (0,0) default for a location-denied user with no device/nodes) and complementary to #2132.

Merged current main (conflict-free with #2132's MeshMapMK changes) and regenerated the docs bundle via build-docs.sh (no diff — already current). SwiftLint clean (pre-existing type_body_length only). CodeRabbit: no actionable comments.

Note: not build-verified locally (watchOS-gated scheme, no build CI).

PR #2118, opened 2026-07-19
a11y: Honor Reduce Motion for continuous animations

bruschill:
## What changed?

Three continuously-running animations now honor the system **Reduce Motion** accessibility setting via `@Environment(\.accessibilityReduceMotion)`; when it's enabled they stop or render statically.

- **RadarSweepView** (Discovery) — the `TimelineView`-driven Canvas sweep freezes via `paused: reduceMotion`.
- **CircularProgressView** (firmware update) — the indeterminate spinner (`repeatForever` rotation) stops and leaves the segment static; re-evaluated on `onChange(of: reduceMotion)`.
- **PulsingCircle** (map node halo) — renders a static halo instead of the breathing `scaleEffect` (a map full of online nodes otherwise pulses many halos at once).
- 3 files changed, +20 / −3.

**Scope note:** This PR was trimmed to the clearly motion-based cases (sweep / spin / scale-pulse). The earlier revision also gated several `.symbolEffect(.variableColor)` sites (Connect, Wi-Fi provisioning, position popover, sensor pin), but `.variableColor` is a color/opacity shimmer rather than positional motion, so those were dropped as out-of-scope for Reduce Motion. Opacity cross-fades (e.g. `LEDIndicator`) are likewise left alone — Apple recommends cross-fade as the *substitute* for motion.

## Why did it change?

The app had 0 uses of `accessibilityReduceMotion` against a continuously-sweeping radar, an infinitely-spinning firmware loader, and breathing map-pin halos. Users who enable Reduce Motion (often for vestibular / motion sensitivity) were still shown all of them.

## How is this tested?

Built the `Meshtastic` scheme for iOS Simulator (iPhone 17 Pro, Xcode 26) → `** BUILD SUCCEEDED **`. SwiftLint clean. Verified on-simulator with Reduce Motion enabled: recorded before/after clips of all three and confirmed via frame-diff that the "after" display stops updating entirely (0 pixel change) while "before" keeps animating. A focused Swift/SwiftUI review earlier confirmed the Reduce Motion toggle reaches live map pins through the `UIHostingConfiguration` trait bridge (EquatableView memoization does not swallow environment-driven invalidation).

https://github.com/user-attachments/assets/d50fb222-35f5-4765-90cf-03fd46e78b20

## Checklist

- [x] I have verified whether these changes require updates to the in-app documentation under `docs/user/` or `docs/developer/`. No doc update is needed (motion-only accessibility change, no UI/flow change) — `skip-docs-check` applied.

## Summary by CodeRabbit

* **Accessibility**
  * Respected the system’s Reduce Motion setting across map pins, radar sweeps, and firmware progress indicators.
  * Disabled or paused animations when Reduce Motion is enabled, including when the setting changes.

### DIFF
--- a/Meshtastic/Views/Nodes/Helpers/Map/MapContent/AnimatedNodePin.swift
+++ b/Meshtastic/Views/Nodes/Helpers/Map/MapContent/AnimatedNodePin.swift
@@ -73,6 +73,7 @@ struct AnimatedNodePin: View, Equatable {
 /// recycles/reconfigures the annotation view (which it does constantly on pan/zoom/declutter). The
 /// per-node `calculatedDelay` phase-shifts each node so they don't all pulse in unison.
 struct PulsingCircle: View {
+	@Environment(\.accessibilityReduceMotion) private var reduceMotion
 	let nodeColor: UIColor
 	let calculatedDelay: Double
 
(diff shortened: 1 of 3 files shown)


garthvh (approved):
Approved after detailed review (accessibility correctness verified; CI green). Part of the a11y series.

PR #2122, opened 2026-07-19
a11y: VoiceOver support for the Watch app

bruschill:
## What changed?

Adds VoiceOver support to the **Meshtastic Watch app**, which previously had **zero** accessibility modifiers.

- **NearbyNodesListView** — each node row announces as one element (name, foxhunt-target flag, distance, bearing°, SNR in dB, relative last-heard) plus an actionable hint; decorative images hidden.
- **DeviceConnectionView** — reachable/unreachable status blocks combined into single spoken elements ("Phone connected" + "N nodes synced"); status icons hidden; Refresh button labeled + hinted.
- **FoxhuntCompassView** — the compass (decorative rotating shapes) is collapsed via `.accessibilityElement(children: .ignore)` into one element with a static label and a **live** `.accessibilityValue` ("Heading N°, Target bearing N°, Distance …") that recomputes as heading/bearing/distance change, plus `.accessibilityAddTraits(.updatesFrequently)` so VoiceOver re-polls while focused.
- **WatchCircleText** — optional `nodeName` so the monogram can announce the full node name.
- 4 files changed, +110 / −24. No catalog changes (see below).

## Why did it change?

The Watch app was completely unusable with VoiceOver — node lists, connection state, and the Foxhunt compass exposed nothing meaningful. This makes the whole app operable without sight.

## How is this tested?

Built the `Meshtastic Watch App` scheme for `generic/platform=watchOS Simulator` (Xcode 26) → `** BUILD SUCCEEDED **`. SwiftLint clean. Passed a focused Swift/SwiftUI review confirming `MeshNode.snr` is dB (so "decibels" wording is correct), `MeasurementFormatter`/relative-date formatting localize via system locale, and there are no force-unwraps or Swift 6/concurrency concerns. A live VoiceOver pass on watchOS (especially the live compass readout) is recommended before merge.

Note: the Watch target does not compile the root `Localizable.xcstrings` (its Resources phase is empty; the catalog belongs to the iOS target only), so strings use `String(localized:)` which resolves to the literal at runtime — **no catalog changes** in this PR.

N/A — no visual change; adds spoken VoiceOver labels/values on watchOS.

## Checklist

- [x] I have verified whether these changes require updates to the in-app documentation under `docs/user/` or `docs/developer/`. No doc update is needed (VoiceOver labels only) — please apply the **`skip-docs-check`** label. (A short VoiceOver note could optionally be added to `docs/user/watch.md` in a follow-up.)

## Summary by CodeRabbit

- **Accessibility**
  - Improved VoiceOver support across device connection, foxhunt compass, and nearby node screens.
  - Added clearer spoken descriptions for phone reachability, refresh actions, node details, compass readings, distance, bearing, signal, and last-heard time.
  - Decorative images and visual elements are now excluded from accessibility navigation.
  - Accessibility information updates dynamically as device and navigation data changes.


garthvh:
Reviewed as part of the a11y series. Holding this one (not approving yet) — one VoiceOver **operability** concern that should be verified on-device before merge.

### `NearbyNodesListView.swift` (~L103) — `.accessibilityElement(children: .ignore)` on the `Button` itself

```swift
Button { selectedNode = node } label: { nodeRow(node) }
    .accessibilityElement(children: .ignore)   // ← applied to the interactive element
```

`.ignore` discards descendant accessibility, which can include the Button's own activation action and `.isButton` trait. If it does here, VoiceOver users hear the composed row label but may be unable to double-tap to open the compass, and may not hear "button" — which defeats the purpose for the exact audience this PR serves.

Contrast the working iOS pattern in `NodeList.swift` (`NodeListItem`): there `.accessibilityElement(children: .ignore)` sits on the **label content inside** the `NavigationLink`, so the outer link still provides activation. Putting `.ignore` on the interactive element itself is the risky case.

**Suggested fix (any one):**
1. Move the three a11y modifiers onto `nodeRow(node)` *inside* the `label:` closure (mirrors the iOS pattern). — preferred
2. Use `.accessibilityElement(children: .combine)` instead of `.ignore` — preserves the child action + button trait; the explicit `.accessibilityLabel` still overrides the spoken label, so output is unchanged.
3. Keep `.ignore` but add `.accessibilityAddTraits(.isButton)` + `.accessibilityAction { selectedNode = node }`.

This **couldn't be verified here** — the Watch target has no build/unit-test CI (only SwiftLint runs), and the watchOS 26.5 simulator runtime needed to build wasn't available, so this is a code-read finding. A quick VoiceOver pass on the Watch (double-tap a node row → compass opens, row announces as a button) would confirm.

Minor / non-blocking notes:
- New `String(localized:)` strings won't localize on watchOS — `Localizable.xcstrings` is only in the iOS target's Resources phase, not the Watch target's. VoiceOver still speaks the English default, and this matches existing Watch code, so it's not a regression — just noting for future localization.
- `WatchCircleText.swift` new `.accessibilityElement(children: .ignore)` + label is currently inert at both call sites (parent already collapses with `.ignore`). Harmless.

Everything else (DeviceConnectionView, FoxhuntCompassView traits/value, unreachable/reachable labels) looks correct. Happy to approve once the node-row activation is confirmed or adjusted.

garthvh (approved):
Approved. Pushed a fix for the VoiceOver operability concern I flagged: the combined a11y element + label/hint are now on the row's label content instead of the Button, so the Button keeps its activation action and `.isButton` trait (mirrors the shipping `NodeListItem` iOS pattern). SwiftLint-clean; current main merged in (conflict-free).

Note: the Watch target isn't build-verified (no Watch build/test CI; local build gated by watchOS runtime). The fix is a 3-line modifier move replicating a proven in-app pattern. A VoiceOver pass on-device (double-tap a node row → compass opens, row announces as a button) is still worth doing post-merge. The two minor notes from review (Watch strings not localized on watchOS; inert WatchCircleText a11y element) are non-blocking non-regressions.

PR #2125, opened 2026-07-19
Fix read messages leaving notifications visible

RCGV1:
## Summary
- remove delivered Meshtastic message notifications when their messages are read in-app
- retain the existing cancellation of pending requests
- cover the delivered-notification identifier cleanup with a focused regression test

## Root cause
The read path only removed pending local notification requests. Once a notification had already been delivered, iOS continued showing it on the Lock Screen and in Notification Center.

## Validation
- `xcodebuild test -project Meshtastic.xcodeproj -scheme Meshtastic -destination platform=iOS\ Simulator,id=5669CB03-8FA2-4A97-BE04-A8566204116D -only-testing:MeshtasticTests/LocalMessageNotificationCleanupTests -derivedDataPath /private/tmp/meshtastic-notification-dismiss-derived-data CODE_SIGNING_ALLOWED=NO`
  - 1 test passed

## Notes
The local SwiftLint pre-commit hook could not run because its installed SourceKit bridge crashes with the active Xcode (`sourcekitdInProc` load failure). The focused Xcode simulator test completed successfully.

## Summary by CodeRabbit

* **Bug Fixes**
  * Canceling notifications by message ID now also removes notifications that have already been delivered, preventing stale notifications from remaining visible.

* **Tests**
  * Added coverage verifying that notifications for multiple message IDs are canceled correctly.

### DIFF
--- a/Meshtastic/Helpers/LocalNotificationManager.swift
+++ b/Meshtastic/Helpers/LocalNotificationManager.swift
@@ -9,9 +9,16 @@ import OSLog
 class LocalNotificationManager {
 
     var notifications = [Notification]()
+	private let removeDeliveredNotifications: ([String]) -> Void
 	let thumbsUpAction = UNNotificationAction(identifier: "messageNotification.thumbsUpAction", title: "👍 \(Tapbacks.thumbsUp.description)", options: [])
 	let thumbsDownAction = UNNotificationAction(identifier: "messageNotification.thumbsDownAction", title: "👎  \(Tapbacks.thumbsDown.description)", options: [])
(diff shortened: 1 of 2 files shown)


garthvh (approved):
Approving. Reviewed the diff: correct fix for delivered notifications lingering after in-app read. The new work is properly gated behind the existing unread early-out (never runs on the repeated poll-tick path once messages are read), makes a single non-blocking batched removeDeliveredNotifications call, and preserves the batching optimization from #1901 — no hot-path regression. All CI green.

PR #2134, opened 2026-07-20
Support generic offline PMTiles and MBTiles maps

RCGV1:
## Summary

This replaces the event-specific offline-map experience with a generic Offline Maps feature.

- Download a map area using the Protomaps Outdoors source, which includes trails, roads, terrain, and points of interest.
- Import and share downloaded `.pmtiles` and `.mbtiles` archives through the native Files, Open In, and share-sheet workflows.
- Preserve each imported archive's original file type in the managed offline-map store.
- Show the current location and existing downloads while choosing a new map area.
- Render saved maps automatically when iOS reports no available network route, without changing the user's Offline Tiles preference.
- Apply the same duplicate, map-count, and storage safeguards to every downloaded map.

## Scope

Only PMTiles and MBTiles are supported in this PR. Other offline-map formats can be added later without changing the user-facing model.

## Validation

- Added PMTiles and MBTiles import coverage, including managed-copy extension preservation.
- Added offline-fallback policy coverage.
- Built successfully for the iOS Simulator with:

```sh
xcodebuild -workspace Meshtastic.xcworkspace -scheme Meshtastic \
  -sdk iphonesimulator -configuration Debug \
  -destination 'generic/platform=iOS Simulator' build CODE_SIGNING_ALLOWED=NO
```


RCGV1:
@coderabbitai review

garthvh (changes requested):
Thanks for putting this together — the pedestrian-street rendering fix and the offline-region plumbing are useful on their own. Two behavioral concerns before this can merge:

**1. The automatic download isn't scoped to the event — it fires for every user**

`BurningManOfflinePack.policy(now:location:)` returns `.install` for any date before `outsideAreaCleanupDate`, with no location check at all:

```swift
static func policy(now: Date, location: CLLocation?) -> BurningManPackAction {
    if now >= noLocationCleanupDate { return .remove }
    if now < outsideAreaCleanupDate { return .install }
    guard let location, isUsable(location, at: now) else { return .retain }
    return contains(location.coordinate) ? .retain : .remove
}
```

Combined with the unconditional call from `MeshtasticAppleApp`'s `scenePhase == .active` handler, this downloads a high-detail Black Rock City region to every install of the app on its next foreground — regardless of whether that user has ever opened Offline Maps or is anywhere near Nevada. Please gate the `.install` branch on the same location-containment check the post-event branch already uses, so the automatic download only happens for users actually at (or near) the event.

**2. Please drop the automatic deletion**

The `.remove` branch — both the outside-area cleanup after event end and the hard `noLocationCleanupDate` purge — deletes the downloaded region without the user asking for it. Once someone has this data, removing it should be their call, same as any other offline map region (swipe-to-delete already covers manual removal in the Map Overlays / Offline Maps UI). Could you drop the automatic-removal path entirely and just leave the pack installed until the user deletes it themselves?

garthvh:
This should only be automatic for the event firmware, you can check for it in the enum

RCGV1:
@garthvh I have talked with Feeb, most burning man people don't install the firmware till the burn.

ayysasha:
The issue is not that the firmware wont be installed - is that a camp admin brings 20+ nodes to give out and tells them to download the app before they get to playa. Then, when they get to playa, they are given a node, connect to app, and then have no map. Having it pre-loaded will be the best UX IMO.

garthvh:
> The issue is not that the firmware wont be installed - is that a camp admin brings 20+ nodes to give out and tells them to download the app before they get to playa. Then, when they get to playa, they are given a node, connect to app, and then have no map. Having it pre-loaded will be the best UX IMO.

This downloads it for every user in the world, which is not going to work.

RCGV1:
It is going to work, it doesn't add that much data to the app and they can remove it if it's a problem

RCGV1:
Although an alternative I'm considering right now may just be to have it be easily shareable via airdrop so camp admins can airdrop the map to people and it just shows up in the app

RCGV1:
## Burning Man offline-map update

Black Rock City offline map rendered in the Apple simulator

### What changed

- Commit f183985a adds an explicit, Burning-Man-firmware-gated download instead of automatic foreground, date, or location-driven downloading.
- Downloaded raw PMTiles files use the native iOS share sheet and can be imported through Files, Open In, and AirDrop.
- Commit 87176699 incorporates the CodeRabbit follow-up: it moves PMTiles validation, hashing, and copying off the main actor; serializes import/download operations; checks the final copied archive before registration; preserves the system-pack quota exception; avoids offering a missing archive for sharing; and redacts externally supplied file names and errors from logs.

### Verification

- 8 focused PMTiles and Burning Man offline-map tests passed on an iPhone 17 Pro simulator.
- Strict SwiftLint reported 0 violations across the other 13 changed Swift files. MeshMapMK has a pre-existing type-body-length violation on current main; this branch reduces that file by 5 lines.
- The image above is a simulator capture of the downloaded Black Rock City map. The native share sheet was exercised on that simulator. iOS Simulator does not offer AirDrop, so an AirDrop destination must be confirmed on a physical iPhone when one is connected.

garthvh:
Reviewed — the file count decomposes cleanly (new import pipeline + fallback policy + tests + the UTType/Files plumbing; the archive readers and OfflineMapManager were reused, not reinvented) and all nine CodeRabbit findings were addressed. One security hardening pushed in 0db34d2d: imported MBTiles are attacker-controllable SQLite, and selecting from a crafted schema can execute SQL defined inside the file — the archive now sets `PRAGMA trusted_schema=OFF` and `PRAGMA cell_size_check=ON` right after the read-only open (SQLite's recommended untrusted-database mitigations; no-ops for honest archives). Import + fallback suites pass (11/11). LGTM after that.

garthvh:
Added a zoom-coverage advisory in 4bca1e54 (follow-on to the review above). Imported archives can carry any zoom range, so one may be missing the levels the map renders at — no close-up basemap when zoomed to a node (max zoom too low) or no basemap when panned out (min zoom too high). `OfflineMapZoomCoverage` classifies the archive's declared range against the app's render band and shows an orange advisory caption on the offline-map row and detail view.

Deliberately a **warning, not a rejection** — a narrow range is a legitimate regional export, not a corrupt file, so import still succeeds. It reads the min/max already captured during validation (no tile scan), so it doesn't detect gaps *inside* a declared range. Thresholds (detail floor z10, overview ceiling z6) are derived from `OfflineMapDetailLevel`'s own z0–z13/z15 targets and are the knobs to tune. Zoom-coverage + import suites pass (13/13).

PR #2123, opened 2026-07-19
a11y: Dynamic Type support for Compact Widget metric tiles

bruschill:
## What changed?

Adds **Dynamic Type** support to the node-detail Compact Widget metric tiles, whose large numeric readouts used hardcoded `.font(.system(size: N))` (30–72pt) that ignored the user's text-size setting and would clip inside fixed-height tiles.

- Large numeric readouts → `@ScaledMetric(relativeTo: .largeTitle) private var valueSize…: CGFloat = N` + `.font(.system(size: valueSize))`, so the font scales with Dynamic Type. Base values equal the prior hardcoded sizes → pixel-parity at the default size.
- Added `.lineLimit(1)` + `.minimumScaleFactor(0.5)` so enlarged text scales-to-fit the uniform grid tile instead of clipping (grid-row alignment preserved).
- Small unit labels `.font(.system(size: 14))` → semantic `.font(.footnote)`.
- Decorative glyphs (`aqi.medium` SF Symbol, `☢` verbatim symbol) left untouched.
- 10 files changed, +71 / −18 (Compact Widgets + `PowerMetrics.swift`).

## Why did it change?

Users who increase their system text size (or use Accessibility larger text) saw these sensor readouts stay fixed-size and, at large settings, clip. Making them scale respects Dynamic Type while keeping the tile grid aligned.

## How is this tested?

Built the `Meshtastic` scheme for iOS Simulator (iPhone 17 Pro, Xcode 26) → `** BUILD SUCCEEDED **`. SwiftLint clean. Passed a focused Swift/SwiftUI review (ship-it) confirming the `@ScaledMetric` usage/anchor is correct and does not fight `minimumScaleFactor`, default-size parity holds, the glyph-vs-text split is right, and there are no caller/layout regressions in `PowerMetrics.swift`/`SoilCompactWidgets.swift`. Verify visually via Settings → Accessibility → Display & Text Size → Larger Text.

## Screenshots/Videos (when applicable)
dyntype_comparison

## Checklist

- [x] My code adheres to the project's coding and style guidelines.
- [x] I have conducted a self-review of my code.
- [x] I have commented my code, particularly in complex areas.
- [x] I have verified whether these changes require updates to the in-app documentation under `docs/user/` or `docs/developer/`. No doc update is needed (presentational Dynamic Type change under `Views/Helpers/`, no behavior/flow change) — please apply the **`skip-docs-check`** label.
- [x] I have tested the change to ensure that it works as intended.

## Summary by CodeRabbit

[...]

PR #2142, opened 2026-07-21
a11y: localize remaining a11y strings, fix isLight() to use WCAG relative luminance (T032-T038)

bruschill:
## What changed?

Two independent fix clusters from the accessibility audit tracker (`specs/016-accessibility-audit/tasks.md`, T032-T038):

**Localization of accessibility strings (T032-T036)**
- `Channels.swift` — `ChannelStatusIcon` labels ("Position sharing", "MQTT uplink/downlink enabled")
- `TextMessageField/FormattingToolbarButtons.swift` — Bold/Italic/Strikethrough/Code/Link toolbar button labels
- `Model/Firmware/FirmwareUpdateNotifier.swift` — the firmware-update notice's `accessibilityHint` (surfaced in `Connect.swift`)
- `Lockdown/LockdownSheet.swift` — passphrase byte-count label and the disabled-submit hint
- `Messages/MessageSearchBar.swift` — "No matches" / "Match X of Y" label

All were hardcoded English literals or unlocalized `Text()` interpolations. Wrapped each in `String(localized:, comment:)` and added the corresponding keys to `Localizable.xcstrings`, matching the minimal-entry format already used for sibling a11y strings in this project. Also backfilled three keys (`Send message`, `Cancel reply`, `Emoji picker`) that an earlier commit on this branch had wrapped in `String(localized:)` but never added to the catalog.

**Contrast correctness (T037-T038)**
- `Meshtastic/Extensions/Color.swift` — `Color.isLight()` / `UIColor.isLight()` used BT.601 luma (`0.299R + 0.587G + 0.114B`) with a flat `>0.5` cutoff to decide black-vs-white text color. Replaced with the actual WCAG 2.x relative luminance formula: linearize each sRGB channel (gamma correction), weight by 0.2126/0.7152/0.0722, and use a `>0.179` cutoff — the luminance value where black and white text give equal contrast (~4.5:1 either way) against the background. Added a `relativeLuminance()` accessor on both extensions.
- `AnimatedNodePin.swift` — the detection-sensor pin's `sensor.fill` icon was hardcoded to white regardless of the node's color. Now uses `swiftUIColor.isLight() ? .black : .white`, matching the pattern `CircleText.swift` already used one line below.
- `Meshtastic Watch App/Views/FoxhuntCompassView.swift` — the watchOS target has its own separate `isWatchLight` (can't share `Color.swift` because that file imports UIKit, unavailable on watchOS). It carried the same stale BT.601/0.6-cutoff bug. Ported the identical WCAG formula/cutoff in, so the Foxhunt compass's close-proximity green state no longer renders white-on-green text.

Because `isLight()`/`isWatchLight` are shared helpers, several other call sites (`CircleText.swift`, `Routes.swift`, `NodeMapContent.swift`, `CompassView.swift`, `WatchCircleText.swift`) get the corrected contrast decision automatically without being touched directly.

## Why did it change?

The audit found two separate correctness gaps:

1. Several VoiceOver-only strings were never routed through the localization system, so non-English users got English announcements from VoiceOver even where the visible UI was already translated.
2. The app's "pick black or white text for this background color" logic used a luma approximation with an arbitrary 0.5 cutoff instead of the WCAG contrast-ratio math that actually predicts legibility. This produces genuinely wrong answers for saturated colors — e.g. pure red's correct WCAG luminance (~0.2126) is above the 0.179 crossover where black text wins, but the old formula scored it at 0.299 luma, under its 0.5 cutoff, and picked white text instead, giving worse contrast than the alternative.

## How is this tested?

- Verified all touched Swift files parse cleanly (`swiftc -parse`).
- Built the `Meshtastic` scheme (Debug, iOS Simulator) via `xcodebuild build` — succeeded.
- Rendered the same color swatches through both the old BT.601 algorithm and the new WCAG algorithm side-by-side, via the real `CircleText` and `AnimatedNodePin` views, using a throwaway DEBUG-only harness fully reverted before the final commit (`git status` confirms a clean tree). Screenshot below shows the fix visually: old algorithm keeps white text on red/orange/yellow/green (wrong), new algorithm correctly switches those to black while keeping white on blue/purple.
- Updated `MeshtasticTests/ExtensionTests.swift`'s `UIColor.red.isLight()` expectation from `false` to `true` to match the corrected math, with an inline comment explaining why (5.25:1 contrast with black vs 4:1 with white).
- Independently re-derived the WCAG formula and the 0.179 crossover value by hand, and had it independently re-verified by the project's `meshtastic-swift-reviewer` agent — recomputed the same reference luminance values (red=0.2126, white=1.0, black=0.0, crossover≈0.179129) from the committed code.
- **Not done**: did not run `xcodebuild test` for the full unit test suite (only manually/programmatically verified the specific updated assertion's expected value); did not build or run the watchOS target itself (only syntax-checked `FoxhuntCompassView.swift` via `swiftc -parse`).
- Validated `Localizable.xcstrings` diffs parse as valid JSON with no duplicate keys after each edit.

T037/T038 contrast fix comparison
*Side-by-side comparison (old BT.601 algorithm vs new WCAG algorithm) rendered through the real `CircleText` and `AnimatedNodePin` (detection-sensor) views across 7 color swatches. Old column: white text on every swatch, including yellow and green. New columns: black text correctly appears on red/orange/yellow/green/gray, white text correctly stays on blue/purple.*

## Checklist

  - No doc update needed — these are presentational/VoiceOver-only correctness fixes (localized string wrapping, corrected text-color contrast math). No new screen, setting, icon meaning, or user-facing flow changed. Adding `skip-docs-check`, same as PR #2123.
  - Built and ran the iOS app in Simulator and visually confirmed the contrast fix via screenshot (see above). Did not run the full automated test suite or build the watchOS target — noted honestly under "How is this tested?".


bruschill:
Closing — opened against the wrong repo by mistake. Re-opening against bruschill/Meshtastic-Apple (my fork) instead, not upstream.

PR #2144, opened 2026-07-21
a11y: fix dropped message badges in combined VoiceOver label, label DM contact rows (T003-T004)

bruschill:
## What changed?

- **T003**: `UserMessageRow.swift` and `ChannelMessageRow.swift`'s combined VoiceOver `accessibilityLabel` for a message bubble only checked encryption and signature flags, so it silently dropped the store-and-forward, detection-sensor, and translated-text badges that `MessageText.swift` already draws and labels individually. Added `MessageEntity.activeStatusBadges(destination:isCurrentUser:)` (in `MessageEntityExtension.swift`) as a single source of truth for which badges apply to a message and what each one's label is. `MessageText.swift`'s corner badges and overlays, and both message rows' combined labels, now all read from this one function instead of three separate hand-written checks. Also moved the destination-based sensor-badge gate out of a private extension in `MessageText.swift` into `MessageDestination.swift` (as `showsDetectionSensorBadge`) since it's shared logic now.
- **T004**: `UserList.swift`'s `DirectMessageUserRow` had no accessibility support at all — the unread dot was announced even when invisible, the lock/key icons and favorite star had no labels, and the row wasn't grouped into a single VoiceOver stop. Ported the unread-dot pattern already used by `ChannelList.swift`'s channel row, added labels to the three lock/key states and the favorite star, and grouped the row with `.accessibilityElement(children: .combine)`.
- Small follow-up: `ChannelMessageRow.swift` had its own local duplicate of the detection-sensor check; pointed it at the new shared `MessageEntity.isDetectionSensorMessage(destination:)` instead.

Also checks off T003/T004 in `specs/016-accessibility-audit/tasks.md`.

## Why did it change?

VoiceOver users could not tell whether a message was verified, store-and-forward, a detection-sensor reading, or showing translated text, because the row-level summary silently dropped those states. The two call sites (DM row, channel row) also hand-listed their own copy of "which badges exist," which is exactly the kind of duplication that causes this drift to reopen the next time a badge is added. Centralizing it in one function on `MessageEntity` means adding a future badge only requires one change, and both rows (and the visual badges themselves) can't get out of sync again. The DM contact list had zero accessibility, making the entire Direct Messages tab unusable with VoiceOver.

## How is this tested?

- Read-reviewed each call site against `MessageText.swift`'s badge conditions to confirm the shared function's boolean logic is byte-for-byte identical to what was inline before (no behavior change to which badges show, only where the check lives).
- Built the `Meshtastic` scheme for `generic/platform=iOS Simulator` (Debug) after each change — build succeeded with no new warnings.
- Installed and launched the app on an iPhone 17 (iOS 26.5) simulator using the existing marketing-seed debug harness, navigated to a seeded channel thread, a seeded DM thread, and the DM contact list, and visually confirmed the affected rows render unchanged (this is a VoiceOver-only change with no visible pixel difference — screenshots below are for reviewer context on which screens/rows the fix touches, not a before/after diff).
- Ran an independent code review pass via the `meshtastic-swift-reviewer` agent against the full branch diff (SourceKit-LSP + SwiftLint backed), specifically checking cross-file consistency between `UserMessageRow.swift`/`ChannelMessageRow.swift`/`MessageText.swift`, SwiftData/`@Model` and Swift 6 concurrency correctness of the new `MessageEntity` extension methods, and localization string correctness. Verdict: approve, no blocking issues; two non-blocking nits noted and not actioned (out-of-scope `ChannelList.swift` row-grouping parity, and two new label strings awaiting normal build-time `.xcstrings` extraction, consistent with this repo's existing convention).
- Did not verify with an actual VoiceOver pass on-device/Simulator — recommend a manual VoiceOver spot-check on the three affected rows before merge.

T003 channel message badges
*T003 — Channel message thread (ChannelMessageRow), showing the bubbles the fixed combined accessibilityLabel applies to.*

T003 DM message badges
*T003 — Direct-message thread (UserMessageRow), same fix applied to the DM call site.*

T004 DM contact list
*T004 — Contacts (DM) list (DirectMessageUserRow), showing the lock/star icons and rows that now have accessibility labels and row grouping.*

Note: the seeded demo data used to generate these screenshots doesn't include a verified/store-and-forward/detection-sensor/translated message, so none of the screenshots show those specific badges rendered — only the screens/rows the fix applies to.

## Checklist

- [ ] I have verified whether these changes require updates to the in-app documentation under `docs/user/` or `docs/developer/`, and updated accordingly. No update needed — this is a presentational, VoiceOver-only accessibility change with no new UI, icon meaning, or user-facing flow (icons/colors/layout are all unchanged; only the accessibility tree and labels changed), and no SwiftData schema change. Recommend adding the `skip-docs-check` label, per the same precedent as PR #2123.
- [x] I have tested the change to ensure that it works as intended. (Build verified on Simulator + independent reviewer-agent pass; see "How is this tested?" for the one gap — no live VoiceOver pass was performed.)

### DIFF
--- a/Meshtastic/Enums/MessageDestination.swift
+++ b/Meshtastic/Enums/MessageDestination.swift
@@ -16,4 +16,13 @@ enum MessageDestination {
 		case let .channel(channel): return channel.index
 		}
 	}
+
+	/// Whether the detection-sensor badge overlays messages sent to this destination. Detection
+	/// Sensor telemetry only ever overlays channel broadcasts, never direct messages.
+	var showsDetectionSensorBadge: Bool {
+		switch self {
+		case .user: return false
+		case .channel: return true
+		}
+	}
 }
--- a/Meshtastic/Extensions/SwiftData/MessageEntityExtension.swift
+++ b/Meshtastic/Extensions/SwiftData/MessageEntityExtension.swift
@@ -9,6 +9,7 @@
 import CoreLocation
 import Foundation
 import MapKit
+import MeshtasticProtobufs
 import SwiftUI
 
 extension MessageEntity {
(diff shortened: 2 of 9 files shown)


bruschill:
Closing — opened against the wrong repo by mistake. Re-opening against bruschill/Meshtastic-Apple (my fork) instead, not upstream.

PR #2156, opened 2026-07-22
fix(build): main doesn't compile — ua868 stragglers from #2151 and an over-budget #Predicate from #2150

garthvh:
Current main fails to build for device and simulator, from two independent merges landing together:

## 1. `ua868` stragglers (#2151)

#2151 removed `case ua868` from the app's `RegionCodes` enum, but four `RegionCodes`-typed references survived, so every switch over the enum stopped compiling:
- the duty-cycle and channel-support switches in `LoraConfigEnums.swift`
- the `toProto` mapping arm
- the frequency table in `LoRaChannelCalculator.swift`
- `.ua868` entries in two test arrays

All removed. The protobuf enum keeps `UA_868` (wire format unchanged); the app simply no longer offers or maps it — matching #2151's intent, since UA_868 is now identical to EU_868.

## 2. `deleteUserMessages` predicate over the type-check budget (#2150)

The five-term compound `#Predicate` (two optional-chained `num` comparisons in an OR, plus three scalar screens) fails with "the compiler is unable to type-check this expression in reasonable time" — the same `#Predicate` macro budget this codebase already documents in `UserMessageList`'s ACK-count fetches. The selective OR stays in the predicate; the scalar screens (`toUser != nil`, `isEmoji`, `admin`, `portNum != 10`) now run in-memory over the single-conversation result. Semantics unchanged.

## Verification

- `generic/platform=iOS` device build: succeeds
- `build-for-testing` (app + test targets) on simulator: succeeds
- Both failures reproduced deterministically on this machine before the fixes (Xcode 26.x); the type-check budget one may be toolchain/machine-sensitive, which would explain green CI on #2150.

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed the unsupported UA868 region option from regional configuration and locale handling.
  * Preserved existing frequency and channel calculations for all remaining regions.
  * Improved deletion of user messages while retaining existing filtering behavior.

* **Tests**
  * Updated region presets and locale classifications to reflect the removal of UA868.

### DIFF
--- a/Meshtastic/Enums/LoraConfigEnums.swift
+++ b/Meshtastic/Enums/LoraConfigEnums.swift
@@ -275,8 +275,6 @@ enum RegionCodes: Int, CaseIterable, Identifiable {
 			return 100
 		case .ua433:
 			return 10
-		case .ua868:
-			return 10
 		case .lora24:
 			return 100
 		case .my433:
@@ -354,8 +352,6 @@ enum RegionCodes: Int, CaseIterable, Identifiable {
 		case .th:
 			return true
 		case .ua433:
-			return true
-		case .ua868:
 			return true
 		case .lora24:
 			return false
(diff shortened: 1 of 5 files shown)


sashko:
@garthvh My apologies, I don't have a dev environment, so I couldn't build the project - just cleaning up all the mentions/uses of the UA 868.

PR #2160, opened 2026-07-22
fix(settings): stop reading stale SwiftData node snapshots

RCGV1:
## Summary

- Stop the Settings refresh task immediately when its sleep is cancelled.
- Exclude deleted or detached `NodeInfoEntity` instances before the Settings picker reads persisted properties.
- Add regression coverage for a deleted node held by a stale snapshot.

## Root cause

Settings keeps a throttled `@State` node snapshot to avoid broad live-query invalidations. The snapshot can outlive a node-database reset or connected-node change. A redacted TestFlight 2.7.17 (2153) crash showed both `Settings.refreshNodes()` after a cancelled task and a `NodeInfoEntity.num` getter on destroyed backing data while switching nodes. The previous `try? await Task.sleep` consumed cancellation and performed one more fetch; later Settings paths also read properties directly from the stale snapshot.

This is separate from the recently merged [position-cache stale-handle repair](https://github.com/meshtastic/Meshtastic-Apple/commit/cb1a6515d9954acad279870d847a35ff68c12898): this fix covers the Settings-owned `NodeInfoEntity` snapshot. An open-PR search for the Settings/SwiftData crash terms found no duplicate.

## Validation

```
DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcodebuild test -project Meshtastic.xcodeproj -scheme Meshtastic -destination 'platform=iOS Simulator,name=iPhone 17 Pro' -derivedDataPath /private/tmp/meshtastic-settings-test -only-testing:MeshtasticTests/SettingsNodeSortTests\n```\n\nPasses: 8 tests, including the deleted-node regression. That regression failed before the fix because the deleted favorite remained in the picker result.\n\nNo crash reports or tester data are included in this PR.

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

- **Bug Fixes**
  - Settings now consistently prioritizes favorite nodes across the radio configuration and related navigation/selection.
  - Deleted or detached nodes are excluded from the configuration picker.
  - If the selected node disappears or the database is reset, Settings clears the selection instead of opening invalid configuration.
  - The periodic refresh loop now stops more reliably when cancelled or interrupted.
- **Documentation**
  - Added “Snapshot liveness” guidance for stale SwiftData snapshots.
  - Expanded Settings documentation with a “Node selection” explanation.
  - Updated transport documentation for expected BLE status behavior during Bluetooth state changes.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


garthvh (approved):
Reviewed and independently verified: adminPickerOrder now screens modelContext == nil / isDeleted before reading persisted properties, and that filtering is correctly threaded through every other stale nodes.first(where:) lookup in Settings.swift via sortedNodes, not just the visible picker list. The try? await Task.sleep cancellation-swallowing fix in the refresh loop is a real, valid bug — same anti-pattern I independently caught and avoided in my own PR's upload-progress task earlier today. Ran the full SettingsNodeSortTests suite myself (not just trusting the PR's claim): 9/9 pass, including both new deleted/detached regression tests. Docs correctly updated in both docs/developer/swiftdata.md and docs/user/settings.md.

Issue #2161, opened 2026-07-22
BLE transport status never settles on .error when Bluetooth is powered off


bruschill:
### Context

Found while reviewing the fix for #2139 (BLE "Enable Bluetooth" alert reappearing on launch for TCP/WiFi-only users). That fix suppresses the system power alert (`CBCentralManagerOptionShowPowerAlertKey: false`) on the assumption that in-app status already covers the off-state gracefully. On closer look, it doesn't.

### The bug

In `BLETransport.handleCentralState(_:central:)`, the `.poweredOff` case:

```swift
case .poweredOff:
    status = .error("Bluetooth is powered off")
    if let connection = activeConnection {
        Task {
            ...
        }
    }
    status = .ready   // <-- immediately overwrites the line above, same branch, synchronously

    // Close the gate to make people wait
    Task { await setupCompleteGate.reset() }
```

`status` is set to `.error("Bluetooth is powered off")` and then unconditionally overwritten to `.ready` a few lines later, in the same synchronous branch — before the cleanup `Task` above even runs. So the transport's own status never actually reports the off-state to any UI/observer; it briefly touches `.error` and immediately reports `.ready` instead.

### Why it matters now

Before #2139's fix, the OS-level "Bluetooth is turned off" alert was an accidental backstop that told BLE-primary users their Bluetooth was off, even though the in-app status didn't reflect it. Now that the system alert is suppressed, a BLE-primary user who has Bluetooth off may get **no** signal at all — neither the system alert nor an accurate in-app status — that BLE is unavailable.

### Suggested fix

Have `.poweredOff` settle on `.error("Bluetooth is powered off")` (or a dedicated "off" status) rather than immediately overwriting it with `.ready`, and confirm downstream UI (Connect screen / device list) surfaces that status to the user. Worth checking whether `.ready` here was intentional for some other reason (e.g. so discovery UI doesn't get stuck showing an error state) — if so, the fix should reconcile "ready to scan once powered on" with "currently powered off" as distinct, both-surfaced states rather than one silently clobbering the other.

### Repro

1. Ensure the app has BLE authorization already granted.
2. Turn Bluetooth off on the device.
3. Launch the app (or bring to foreground) with no active BLE connection.
4. Observe: no system alert (after #2139 fix), and the BLE transport/device list gives no indication Bluetooth is off.

PR #2162, opened 2026-07-22
Fix: BLE 'Enable Bluetooth' alert dismiss/reappear loop on launch (#2139)

### LINKED ISSUE #2139: 🚀 [Feature Request]: Disbale "Bluetooh enable" popup request when (re)opening the app
### OS

iPadOS

### Description

Hey,

I switched connection from BT to WiFi, as I'm not always in BT range. Connection via IP works fine.
But, everytime I open the app, the popup "Enable BT" (re)appears, even though I don't want to use BT. If you close it, the popup reappers in under 1sec, you have to spam click close to and try to click on your TCP connection entry.
Would be nice to have at least the option to disbale this reqest during the app start

Same behaviour with disabled BT on iPadOS 26.5.2 and iOS 26.5.2, app version 2.7.16(2109).

Thanks a lot!

### Participation

### Additional comments

_No response_

### Code of Conduct

bruschill:
## What changed?

`Meshtastic/Accessory/Transports/Bluetooth Low Energy/BLETransport.swift`: factored the duplicated `CBCentralManager` options dictionary (previously built inline in both `init()` and `createCentralManager()`) into a single `static func centralManagerOptions(restoreIdentifier:)`, and added `CBCentralManagerOptionShowPowerAlertKey: false` to it.

Also adds `MeshtasticTests/BLETransportCentralManagerOptionsTests.swift`.

## Why did it change?

Fixes #2139: a TCP/WiFi-only user with Bluetooth off on the device reported the system "Enable Bluetooth" alert appearing on every app launch and reappearing within ~1s of being dismissed, over and over, blocking interaction with the Connect tab.

Root cause, confirmed by reading the code rather than guessing: `AccessoryManager.startDiscovery()` fans out to *every* registered transport (BLE, TCP, Serial) unconditionally at launch (`MeshtasticApp.swift`) and again on every `scenePhase -> .active` transition (`appDidBecomeActive()`, when there's no active connection yet and `discoveryTask == nil`) — regardless of which transport the user actually connects with. `BLETransport.discoverDevices()` creates/uses a `CBCentralManager` and calls `scanForPeripherals` on it. `CBCentralManagerOptionShowPowerAlertKey` was never set anywhere in the codebase, so it defaulted to `true` — showing the system "Bluetooth is turned off" alert any time the manager is `.poweredOff`.

Presenting that alert blips `scenePhase` (inactive/background then active again), and `appDidBecomeActive()` restarts BLE discovery in that state, which calls `scanForPeripherals` on the still-off manager again — retriggering the alert. That's the dismiss/reappear loop from the report, and it happens even though the user has no intention of using BLE this session.

Suppressing the system alert doesn't change BLE functionality: `handleCentralState`'s `.poweredOff` case still runs (status update, active-connection cleanup) — only the OS-level alert is suppressed. The explicit onboarding "enable Bluetooth" flow (`BluetoothAuthorizationHelper.requestAuthorization()`) constructs its own `CBCentralManager` with default options (no options dict at all), so that user-initiated prompt is untouched.

## How is this tested?

`MeshtasticTests/BLETransportCentralManagerOptionsTests.swift` — 4 cases on the pure options builder:
- power alert is suppressed (`ShowPowerAlertKey == false`)
- restore identifier is preserved (guards against breaking BLE state restoration while fixing the alert)
- pass-through of an arbitrary restore identifier is genuine, not hardcoded
- the dict contains exactly the two expected keys (no silent future additions/omissions)

```
xcodebuild test -scheme Meshtastic -destination 'platform=iOS Simulator,name=iPhone 17 Pro' \
  -only-testing:MeshtasticTests/BLETransportCentralManagerOptionsTests
```
4/4 passed.

Reviewed via the project's `meshtastic-swift-reviewer` agent: approved. Verified via sourcekit-lsp diagnostics (clean) and swiftlint (clean) on both changed files, confirmed the `static func` on the actor has no isolation concerns (pure function, no actor-state access), and confirmed both `CBCentralManager` construction sites (including the deferred-creation path when authorization was `.notDetermined` at init) route through the new helper.

Also filed #2161 as a follow-up: `handleCentralState`'s `.poweredOff` case sets `status = .error(...)` then immediately overwrites it with `.ready` in the same branch, so the transport's in-app status never actually settles on the off-state. That's pre-existing and out of scope for this PR, but worth tracking since this fix removes the system alert that was previously an accidental backstop signal for BLE-primary users. (Fixed separately in #2163.)

Not applicable — no UI change. The observable effect is the absence of the repeated system alert, not something a screenshot captures well; the behavior was reproduced by reading the discovery/scenePhase code path (see root-cause explanation above) rather than a device repro.

## Checklist

- [x] I have verified whether these changes require updates to the in-app documentation under `docs/user/` or `docs/developer/`, and updated accordingly (see [copilot-instructions.md](../.github/copilot-instructions.md#in-app-documentation) for the view → doc page mapping). If no doc update is needed, add the **`skip-docs-check`** label. — `docs/developer/transport.md` doesn't describe the power-alert behavior; no update needed. `skip-docs-check` label applied.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced unnecessary Bluetooth power alerts during connection and state restoration.
  * Preserved Bluetooth state restoration while using more consistent connection settings.

* **Tests**
  * Added coverage to verify Bluetooth configuration values and ensure no unexpected options are included.

### DIFF
--- a/Meshtastic/Accessory/Transports/Bluetooth Low Energy/BLETransport.swift
+++ b/Meshtastic/Accessory/Transports/Bluetooth Low Energy/BLETransport.swift
@@ -51,7 +51,7 @@ actor BLETransport: Transport {
 		if CBCentralManager.authorization != .notDetermined {
 			centralManager = CBCentralManager(delegate: delegate,
 											  queue: centralQueue,
-											  options: [CBCentralManagerOptionRestoreIdentifierKey: kCentralRestoreID]
+											  options: Self.centralManagerOptions(restoreIdentifier: kCentralRestoreID)
 			)
 		}
 		self.delegate.setTransport(self)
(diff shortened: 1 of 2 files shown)


garthvh (approved):
Independently verified all three claims against the actual code: startDiscovery() genuinely fans out to every transport unconditionally regardless of user intent, the explicit onboarding Bluetooth-authorization flow uses a completely separate CBCentralManager untouched by this change, and (the one real gap) there's currently no in-app signal anywhere for a BLE user with Bluetooth off once this system alert is suppressed — filed as #2175 to track separately, since it's a pre-existing gap this PR exposes rather than one it introduces (the same user hit an interaction-blocking loop before this fix, not a clean working alert).

PR #2174, opened 2026-07-23
fix(map): reorganize Map Options sheet, merge Manage Map Data into overlay list

garthvh:
## What changed?

Cleaned up the Map Options sheet (the "i" button on the Map tab):

- Moved **Offline Tiles** into the **Offline Maps** section, alongside the downloaded-regions link
- Moved **Cluster Nodes** up under **Precise Locations Only**
- Folded the separate **Manage Map Data** screen's functionality into the **Map Overlays** file list directly — per-file metadata (format, feature count, size, upload date) now shows inline, and deleting a file is a swipe action instead of a persistent trash button
- Replaced the old **Manage Map Data** link with an inline **Upload Map Data** button

Also fixes a layout bug found while testing this on a phone at a larger Dynamic Type size: the file-type pill (e.g. "GEOJSON") could wrap onto two lines. It's now pinned to a single line, and the feature count drops to its own line instead when there isn't room for both.

`MapDataFiles.swift` (the standalone Manage Map Data screen, still reachable from Settings > App Data) is untouched.

## How is this tested?

Built and installed to a physical device; walked through the reorganized sheet — section moves, merged overlay file list with swipe-to-delete, and the pill/feature-count layout at a large Dynamic Type size.

## Summary by CodeRabbit

* **New Features**
  * Upload map overlay files directly from Map Settings.
  * View upload progress and processing status, with error messages when needed.
  * Enable or disable individual uploaded overlays.
  * Delete overlay files with a swipe action.
  * View overlay format, feature count, file size, and upload date.
  * Added a node clustering option to the map settings.

* **Changes**
  * Removed the Offline Tiles toggle from the offline maps section.


garthvh:
Addressed CodeRabbit's review:

- **Progress bar (Major)** — removed the fake fixed-duration progress simulation, which was stacking an artificial delay before real work started and still showing "Processing..." after hitting 100% while the actual upload was ongoing. Replaced with an indeterminate spinner tied directly to the real operation.
- **Delete rollback (Minor)** — the overlay's enabled/disabled config entry is now only removed after the file actually deletes, instead of optimistically upfront.
- **`.localized` misuse (nitpick)** — both error-message call sites were interpolating the error text before localizing, so the lookup key changed per-error and never matched a translation. Switched to a static format template.
- **Docs (nitpick)** — updated `docs/user/map.md`, which is the actual home for Map Options documentation (the `Views/Nodes/Helpers/Map/` path prefix seems to have thrown off the nodes.md suggestion). Regenerated the in-app docs bundle.
- **Task cancellation (nitpick)** — left as-is. These Tasks are intentionally fire-and-forget so an upload/delete finishes even if the sheet is dismissed mid-operation; nothing calls `.cancel()` on them, so adding `Task.checkCancellation()` alone wouldn't do anything without also wiring up real cancellation, which would change the intended behavior.

PR #2184, opened 2026-07-24
fix: eliminate offline-map freeze at street-level zoom

bruschill:
## What changed?

- **`MeshMapMK.offlineVectorOverlayGroups`** — the offline-map basemap's overlay construction (earth fill, water/park fills, road casing/fill, rail/boundary) now runs as a chunked, yielding `AsyncStream` — one role group per iteration, `await Task.yield()` between chunks — instead of building every `MKPolygon`/`MKMultiPolygon`/`MKMultiPolyline` in one uninterrupted synchronous main-actor pass.
- Superseded builds are cancelled for real via `AsyncStream.onTermination` (+ `Task.isCancelled` checks before each yield), instead of just being ignored by the consumer.
- Results are staged locally and swapped into `offlineVectorOverlays` atomically once, so a rebuild never flashes the basemap to blank mid-build.
- **`MapConnectivityMonitor`** (new, `Meshtastic/Helpers/Map/`) — an `NWPathMonitor`-backed `isOffline` flag, wired into `clusterConfiguration` to skip MapKit's live-network extras (traffic, POI) while there's no network path at all.
- **`MeshtasticTests/OfflineOverlayBuildPerfTests.swift`** (new) — perf evidence for the chunking change against synthetic decode data, plus a dark-mode regression test for the light-only road-casing group count.
- **`docs/developer/architecture.md`** — new "Offline Maps" section describing the pipeline.

## Why did it change?

Field report: zooming into a downloaded offline map region to ~1.5mi scale would freeze the map UI, requiring a force-quit. It only reproduced fully offline — "no issues once I had even the slightest of service again."

Root cause: `MeshMapMK.rebuildOfflineVectorOverlays` built the entire offline basemap in one synchronous main-actor pass, triggered the first time the downloaded region enters the viewport — which lands right at the same z13-14 "full street grid" zoom tier the ~1.5mi report scale corresponds to (see the code's own tile-count/zoom-selection comments). That CPU burst colliding with MapKit's own live-basemap network-fetch retries under true zero connectivity — the app had no connectivity awareness anywhere before this change — is what turns a survivable cost into an unresponsive freeze that needs a force-quit: any connectivity lets MapKit's requests fail/succeed fast instead of retrying, and the two builds no longer contend for the same run loop.

**Reviews on this PR:**
- Ran the `meshtastic-reviewer` agent before opening the PR. It caught a real Task-lifecycle bug (a superseded overlay build wasn't actually cancelled, just ignored — it kept running concurrently with the build that superseded it, partially defeating the point of chunking) plus a visible flash-to-blank regression and a missing `swiftlint:disable disable_print` pragma. All fixed in the first pushed commit.
- CodeRabbit's automated review caught two more real issues in a follow-up commit: (1) the early-return branch (offline tiles disabled/unavailable/empty coverage) cleared `offlineVectorOverlays` but didn't invalidate the build generation counter, so a build already in flight could later overwrite that deliberate clear with stale results; (2) the production perf-log line used `Duration.formatted()`, which rounds to whole seconds and would print `"0 s"` for nearly every real (millisecond-scale) build. Both fixed, plus a trivial cast-clarity nit in the new test.

## How is this tested?

- `xcodebuild build` / `build-for-testing` — clean, no new warnings, including no new Swift 6 `Sendable`/concurrency warnings under `SWIFT_STRICT_CONCURRENCY = targeted`.
- `swiftlint lint --quiet` on all touched/new Swift files — clean (the one remaining `MeshMapMK` `type_body_length` warning is pre-existing, confirmed via `git stash` diff-of-diff before this change).
- 11/11 tests pass across the new `OfflineOverlayBuildPerfTests` suite and the pre-existing `OfflineVectorPerfTests` / `MapDataMetadataTests` suites.

### Performance evidence

`OfflineOverlayBuildPerfTests` exercises `MeshMapMK.offlineVectorOverlayGroups` directly against **synthetic** decode output (no `.pmtiles` fixture needed — this pipeline stage only consumes already-decoded `OfflineMapPolygon`/`OfflineMapPolyline` value types), sized like a dense "High detail" archive's full street grid: 24,000 road segments across every role + 3,000 polygon fills. It measures:

- **total build time** — this *is* the old code's single uninterrupted main-thread block, since the old code did all this work synchronously in one call.
- **worst single yielded chunk** — this *is* the new code's worst-case single uninterrupted main-thread block, since every chunk boundary is now a run-loop turn where touches (and the force-quit gesture) get serviced.

```
OFFLINE-OVERLAY-PERF: 12 groups / 12 overlays over 3000 fills + 24000 road segments
  total build time (== worst-case main-thread block BEFORE this fix): 25.72 ms
  worst single yielded chunk (== worst-case main-thread block AFTER this fix): 5.00 ms
  max-single-block reduction: 5.1x
  per-chunk breakdown (ms): 3.82, 4.63, 0.40, 0.38, 5.00, 1.59, 1.57, 1.60, 1.57, 1.57, 1.64, 1.57
```

**~5x reduction in the single longest uninterrupted main-thread block**, consistent across repeated runs (4.6x–5.3x observed). The effect grows, not shrinks, on denser real archives, since group count is fixed at 12 (9 in dark mode) regardless of how much data lands in each group.

Not applicable — no visible UI/behavior change. This is an internal scheduling change (overlay construction moved off one synchronous pass) plus a new background connectivity flag and a perf test; nothing new renders on screen.

## Checklist

- [x] I have verified whether these changes require updates to the in-app documentation under `docs/user/` or `docs/developer/`, and updated accordingly (see [copilot-instructions.md](https://github.com/meshtastic/Meshtastic-Apple/blob/main/.github/copilot-instructions.md#in-app-documentation) for the view → doc page mapping). If no doc update is needed, add the **`skip-docs-check`** label. — added a new "Offline Maps" section to `docs/developer/architecture.md`.

## Summary by CodeRabbit

* **New Features**
  * Map views now detect offline connectivity and automatically disable network-dependent traffic and point-of-interest overlays.
  * Downloaded offline maps now build and display incrementally for smoother map interaction.

* **Bug Fixes**
  * Reduced freezing and responsiveness issues when viewing detailed areas in downloaded maps.

* **Documentation**
  * Added documentation describing offline map processing and connectivity behavior.


garthvh:
Closing this — the performance case doesn't hold up against the reported symptom, and the added complexity isn't proportionate to what's actually demonstrated.

The perf test itself reports:
- Total build time (old code, unchunked): ~24-27ms
- Worst single chunk (new code): ~4.5-5.85ms
- Reduction: 4.6-5.3x

27ms divided by 5 is still under 6ms. A main thread blocked for 27ms is roughly a frame and a half at 60fps — not perceptible as lag, let alone a freeze severe enough to require a force-quit. The reported bug describes total unresponsiveness that persisted until the app was killed, which implies multiple seconds of a blocked main thread, not tens of milliseconds. A 4.6-5.3x speedup on a baseline that was never large enough to produce the reported symptom doesn't close that gap.

Two things make the case weaker, not stronger:
- The 24,000 road segment / 3,000 fill synthetic dataset is an estimate, not measured from a real archive. The repo already has a benchmark against real data (bellevue.pmtiles in MapDataModelTests.swift), but that fixture isn't checked in, so this was never validated against an actual downloaded region.
- The fix only chunks object construction. ClusterMapView.syncOverlays still hands the complete overlay set to mapView.addOverlay(...) in one synchronous, unchunked loop after the chunked build finishes. If the real cost is in MapKit rendering/tessellating the geometry rather than in allocating the Swift objects, this change doesn't touch it, and nothing here measures it.

We also haven't independently confirmed the other half of the diagnosis (MapConnectivityMonitor, disabling traffic/POI while offline) — the diagnosis doc itself proposed verifying with Network Link Conditioner plus an Instruments trace which of the two hypothesized causes actually dominates, and that verification wasn't done before this was opened.

Given the one piece of hard evidence here shows roughly a 20ms improvement rather than something that plausibly explains a multi-second freeze, we're going to hold off and keep the existing code. Appreciate the depth of the investigation — happy to revisit if there's a way to get real numbers (an Instruments trace during an actual repro, or real archive data) that show where the time is actually going.

PR #2192, opened 2026-07-24
Fix crashes on malformed or truncated PMTiles archives

RCGV1:
## Summary
- Reject truncated or malformed PMTiles headers and directories before indexed reads.
- Make PMTiles offset arithmetic and file slicing overflow-safe.
- Add regression tests for truncated headers, invalid directories, out-of-range directories, and overflowing tile offsets.

## Why

This is a crash-robustness fix, not a security fix. The archive fetch is hardcoded HTTPS to `build.protomaps.com` with default ATS (no exceptions), so attacker-chosen bytes would require compromising Protomaps or its CDN first — that is not the concern here. And Swift integer/bounds traps are memory-safe, so the worst outcome was always a crash, never code execution.

The concern is that malformed bytes reach this parser through entirely benign paths, and the old code trapped (crashed the app) deterministically instead of returning "no tile":

- `fetchRange` validates only the HTTP status, not body length — a short 200/206 body (a range clamped at EOF, an empty error body from a misbehaving edge) reached `parseHeader` with fewer than 7 bytes, which trapped on the first `subdata` read.
- The leaf-directory path falls back to `gunzip(raw) ?? raw` — any failed decompression (e.g. a body truncated on a flaky connection) feeds raw gzip bytes into the varint parser, where values above `Int.max` trapped on integer conversion.
- An extraction interrupted mid-write leaves a corrupt archive on disk that the map reopens on every launch — a deterministic crash loop until the file is manually deleted.
- `Documents/OfflineMaps` is user-writable via the Files app (`UIFileSharingEnabled` is on), and community-shared `.pmtiles` files become ordinary inputs once offline packs circulate — corrupt or hand-assembled archives will happen.

## Verification

- The new regression tests crash **unpatched `main`** outright — verified by grafting only the test file onto main and running it: repeated test-runner crashes, including `Fatal error: Not enough bits to represent the passed value`. Each guarded path was a real, reachable trap: a 5-byte directory (`0 - 1` UInt64 underflow), an 8-byte header (out-of-range `subdata`), a `UInt64.max` offset (`Int` conversion), and overflowing offset addition.
- On this branch the same suites pass 12/12 with no crashes, and every pre-existing round-trip test still passes — valid-archive behavior is unchanged.
- Focused suites also pass on an iPhone 17 Pro simulator per the original validation run; SwiftLint clean on touched files.

## Follow-up (not in this PR)

`PMTilesExtractor` still contains the same unchecked arithmetic on remote bytes (`header.tileDataOffset + entry.offset` when planning tile reads, `offset + length` in `directory()`); the same guard treatment should be applied there.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of malformed or truncated map tile archives.
  * Prevented invalid tile lookups when offsets overflow or point outside the available data.
  * Added stricter validation for directory entries, headers, and byte ranges.

* **Tests**
  * Added coverage for invalid headers, directory offsets, tile offsets, and malformed directory data.

### DIFF
--- a/Meshtastic/Helpers/Map/PMTilesArchive.swift
+++ b/Meshtastic/Helpers/Map/PMTilesArchive.swift
@@ -101,6 +101,7 @@ final class PMTilesArchive {
 
 	static func parseHeader(_ data: Data) -> PMTilesHeader? {
 		// Magic "PMTiles" + version 3.
+		guard data.count >= 127 else { return nil }
 		let magic = data.subdata(in: 0..<7)
 		guard magic == Data("PMTiles".utf8), data[7] == 3 else { return nil }
 
@@ -154,10 +155,13 @@ final class PMTilesArchive {
 
 			if entry.runLength == 0 {
 				// Leaf directory pointer — descend.
-				dirOffset = header.leafDirOffset + entry.offset
+				let (leafOffset, overflow) = header.leafDirOffset.addingReportingOverflow(entry.offset)
+				guard !overflow else { return nil }
+				dirOffset = leafOffset
 				dirLength = UInt64(entry.length)
 			} else {
-				guard let raw = slice(offset: header.tileDataOffset + entry.offset, length: entry.length) else { return nil }
+				let (tileOffset, overflow) = header.tileDataOffset.addingReportingOverflow(entry.offset)
+				guard !overflow, let raw = slice(offset: tileOffset, length: entry.length) else { return nil }
 				return decompress(raw, using: header.tileCompression)
 			}
 		}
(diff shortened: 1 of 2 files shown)


garthvh:
Maintainer edit: retitled from "Harden PMTiles archive parsing" and rewrote the description to frame this accurately as a crash-robustness fix rather than security hardening — the fetch is hardcoded HTTPS to build.protomaps.com under default ATS, so adversarial bytes would require compromising Protomaps/its CDN first, and Swift traps are memory-safe (crash, not code execution). None of that diminishes the fix: the guarded paths are reachable through entirely benign corruption (short range responses, the gunzip-fallback feeding raw bytes to the varint parser, interrupted writes, Files-app-writable archives), and we independently verified every one of them by running this PR's tests against unpatched main — the test process crashes outright there, and passes 12/12 here with valid-archive behavior unchanged. Good fix; also noted a follow-up for the same arithmetic in PMTilesExtractor.

PR #2193, opened 2026-07-24
Avoid orphan PAX counter records

RCGV1:
Summary
- Verify the sending node exists before allocating a PAX counter entity.
- Preserve PAX history for known nodes.
- Add persistence regressions for both unknown and known senders.

Why
- Unknown PAX packets previously allocated an unlinked SwiftData row; the surrounding receive path could persist these orphans, allowing noisy traffic to grow unusable local data.

Validation
- `xcodebuild test -project Meshtastic.xcodeproj -scheme Meshtastic -destination "platform=iOS Simulator,name=iPhone 17 Pro" -derivedDataPath /private/tmp/meshtastic-defcon-hardening-tests -only-testing:MeshtasticTests/PaxCounterPacketIngestTests`
- Both unknown-sender rejection and known-node persistence passed on iPhone 17 Pro simulator.
- Pre-commit SwiftLint completed with no serious violations.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PAX counter packet handling so counter records are only persisted when the related node information is available.
  * Ensured ingested PAX counter data is correctly associated with the node and saves the expected Wi‑Fi and Bluetooth values.
* **Tests**
  * Added coverage for PAX counter ingestion, including scenarios for unknown nodes (no records persisted) and existing nodes (one correctly linked record saved).

### DIFF
--- a/Meshtastic/Helpers/MeshPackets.swift
+++ b/Meshtastic/Helpers/MeshPackets.swift
@@ -799,7 +799,7 @@ actor MeshPackets {
 			let fetchedNode = try modelContext.fetch(fetchDescriptor)
 
 			guard let node = fetchedNode.first else {
-				Logger.data.info("Node Info Not Found")
+				Logger.data.debug("🗄️ [PaxCounterEntity] Ignoring packet from unknown node \(packet.from.toHex(), privacy: .public)")
 				return
 			}
 			guard let paxMessage = try? Paxcount(serializedBytes: packet.decoded.payload) else { return }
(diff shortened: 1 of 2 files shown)


garthvh:
Closing. The underlying defect is real — the dispatch path has no upstream sender upsert, so a PAX packet from an unheard node reaches this handler with no node row, and on main that inserts a permanently orphaned PaxCounterEntity for the next debounced save to sweep in. But the fix direction is wrong: in Meshtastic, hearing any packet from a node is how the app learns that node exists. Dropping the reading contradicts the pipeline's own convention — telemetryPacket and deviceMetadataPacket both resolve the sender via findOrCreateNode (whose doc comment exists precisely for the "data packet arrives before the sender's NodeInfo" case), and PAX is just another telemetry flavor. The correct fix is `fetchedNode.first ?? findOrCreateNode(num:context:)`, which creates the stub node and user and links the reading, not a guard-and-return that throws mesh-learned data away.

PR #2198, opened 2026-07-25
fix(ble): stop a failed background restore from killing BLE discovery and later connects

bruschill:
## What changed?

`BLETransport.handleWillRestoreState` sets up four pieces of transport state — the `restoreInProgress` discovery gate, `activeConnection`, `restoredConnectContinuation`, and a `discoveredPeripherals` entry — but every failure path reset exactly one of them, and the wait for `didConnect` had no timeout.

- The restore task now calls `endRestore(clearingConnection:)` from a `defer`, so no current or future early return can leave the discovery gate latched. `beginRestore(for:)`/`endRestore(clearingConnection:)` are the only writers of that gate. The `.connecting` branch no longer spawns a nested `Task`, so the `defer` covers the whole restore instead of firing before the work starts.
- `awaitRestoredConnect(timeout:connect:cancelConnect:)` bounds the wait for `didConnect` (`BLETransport.restoredConnectTimeout`, 30s) and withdraws the outstanding request with `cancelPeripheralConnection` when it expires. It is cancellation-aware, resolves a continuation it would otherwise supersede, and carries a generation number so a late watchdog can't take down a newer restore's wait.
- `endRestore` resolves any pending restored continuation and clears `activeConnection`, but only while that is still the exact connection the restore installed.
- `connectionDidDisconnect`, already the single funnel for "this peripheral is gone", resolves a pending restored connect and routes its gate clear through `endRestore`.
- The restore short-circuits in `handleDidConnect` / `handleDidFailToConnect` are scoped to the peripheral being restored (`restoreInProgress` plus an identifier match), the same way the normal connect path checks `connectingPeripheral`.
- `docs/developer/transport.md` gains a "BLE State Restoration" section describing the restore lifecycle; the bundled copies were regenerated with `bash scripts/build-docs.sh --output Meshtastic/Resources/docs`.

## Why did it change?

The trigger is ordinary user behaviour: iOS kills the app in the background while it is connected, then relaunches it through `willRestoreState` with the peripheral in `.connecting`, and the radio has been powered off or moved out of range in the meantime. Three separate failures fell out of that.

**BLE discovery stopped completely, with no error.** CoreBluetooth connect requests never expire, so with the radio absent none of `didConnect`, `didFailToConnect` or `didDisconnectPeripheral` ever arrives. The continuation stayed suspended and `restoreInProgress` was never cleared. That flag gates all three scan entry points — `didDiscover` returns early, the `discoverDevices()` setup task skips `scanForPeripherals`, and the `.poweredOn` re-scan is skipped — so Available Radios stayed empty for the life of the process, autoconnect never fired, and Bluetooth reported healthy. Toggling Bluetooth did not recover it, because the `.poweredOn` recovery path is gated on the same flag. Only force-quitting the app cleared it.

**Every later connect was rejected.** A failed restore left `activeConnection` pointing at a connection that never reached `BLEConnection.connect()`. Nothing tears that down, since `connectionDidDisconnect` only runs from `BLEConnection.disconnect()`, so the busy guards in `connect(to:)` rejected the next attempt with "Connect request while an active connection exists", which the user sees as "Too Many Retries".

**A stale continuation hijacked the next real connect.** `handleDidConnect` checked `restoredConnectContinuation` before `connectContinuation` and returned early, so a later legitimate connect's `didConnect` resumed the dead restore instead. No `BLEConnection` was built, the real continuation was never resumed, and that connect hung to its full timeout (5s, or 90s on a first-ever bond). The revived restore task then called `AccessoryManager.shared.connect(to:)` with the old device and the dead connection; that method's only reentrancy guard (`activeConnection != nil`) passed because it was still nil, starting a second connect state machine over the top of the first.

## How is this tested?

`MeshtasticTests/BLETransportRestoreStateTests.swift` (new, 6 tests) covers the teardown contract: an unanswered restored connect times out and withdraws its CoreBluetooth request; a disconnect and an `endRestore` each resolve a pending restored connect, asserted by which error resolved it rather than by elapsed time; cancelling the enclosing task resolves the wait; a second restore resolves the one it supersedes; and ending a restore reopens discovery.

Each test was verified to fail against the unfixed behaviour before being kept — removing the watchdog hangs the timeout test, and removing the continuation resolution from `connectionDidDisconnect` / `endRestore` plus the gate clear fails the other three on their specific assertions.

`handleWillRestoreState` itself cannot be driven from a test, because `CBPeripheral` has no public initializer and the restore dictionary requires a real one. The tests therefore drive the pieces it delegates to, in the same order the restore task uses. For the same reason the identity checks added to `handleDidConnect` / `handleDidFailToConnect` are not covered by a test.

Full suite: `xcodebuild test -project Meshtastic.xcodeproj -scheme Meshtastic -destination 'platform=iOS Simulator,id=<sim>'`. 23 tests fail, and the identical 23 fail on a pristine `upstream/main` build (2 `CoreDataMigrationServiceTests`, 1 `IntervalConfigurationDetailedTests.allCases_count`, 20 SwiftUI snapshot tests). This branch adds none. SwiftLint on the changed files reports only the pre-existing `type_body_length` warning on the `BLETransport` actor body, which was already over the limit before this change.

Not verified: the end-to-end restore on real hardware. Reproducing it needs a physical radio, a background kill, and the radio powered off before relaunch.

`Meshtastic/Resources/docs/index.json` is deliberately not included. Rebuilding the docs bundle reorders that file's keyword arrays nondeterministically; the same churn appears on a pristine `upstream/main` rebuild, so it is pre-existing noise rather than part of this change.

No UI change. This is transport-layer state handling; the user-visible effect is that Available Radios keeps working after a failed background restore instead of going permanently empty.

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Bluetooth connection restoration after app relaunches.
  - Prevented stale or unrelated connection events from disrupting active restores.
  - Added bounded waiting, cancellation handling, and cleanup when restored connections fail, disconnect, or time out.

- **Documentation**
  - Added guidance explaining BLE state restoration behavior and failure handling.


garthvh:
Not going to merge any more of these huge ai ble pulls, can't afford regressions like that

PR #2199, opened 2026-07-25
Treat a cancelled connect step as a failure instead of a success

bruschill:
## What changed?

A cancelled step in the connect state machine (`SequentialSteps` in `AccessoryManager+Connect.swift`) is no longer treated as a step that succeeded.

- **`case is CancellationError:` now throws instead of `break stepRetryLoop`.** A `break` does not throw, so the enclosing `do` completed normally, the outer failure handler never ran, and the machine moved on to the next step. The existing `externalError` substitution is kept, so a recorded error still replaces the bare cancellation.
- **Step 5a (wait for the end of the node-DB dump) gets an explicit timeout.** It had none. It is set to 120s: the cap is a backstop for a narrow teardown race, not a service level for the dump, and Step 5a is `.retryAll`, so a cap tight enough to abort a slow-but-healthy dump on a large mesh would restart the whole pipeline and end in `tooManyRetries` with autoconnect trying again.
- **`connect(to:)` refuses to start while `connectionStepper` is still running.** The only reentrancy guard was `activeConnection != nil`, which a teardown has already nilled.
- **`defer { isRunning = false }` in `run()`**, replacing the three explicit assignments. They covered the normal return, `.fail` and retry-exhaustion paths only, so a throw out of the retry sleep at the top of the step loop (that sleep sits outside the per-step `do`) left `isRunning` stuck at `true`.

Also fixes a stale log label on Step 5a's skip path (it said "Step 4").

## Why did it change?

Only `.retryStep` steps could reach that cancellation branch. Every `.retryAll` step has `stepRetries == 1`, so `stepRetryAttempt` is always the last attempt and the error is rethrown before the switch. Step 5 (`sendWantDatabase`, `.retryStep(attempts: 3)`) is the pipeline's only `.retryStep` step, and it is also the one that parks on a cancellable continuation.

The trigger is a teardown that resumes that continuation without going through `cancelCurrentlyExecutingStep`. The `.disconnected` event handler in `didReceive(_:)` calls `closeConnection()` directly, which resumes `firstDatabaseNodeInfoContinuation` with a bare `CancellationError` and then does `wantDatabaseGate.cancelAll()` + `reset()`. `externalError` is nil, which is exactly the case that hit the `break`.

The machine then advanced into Step 5a, which waits on the gate `closeConnection()` had just reset to closed. Only a `configComplete` for `NONCE_ONLY_DB` reopens that gate, and the link was gone, so with no timeout the step never finished. `connect()` never returned and `connectionStepper` was never cleared.

User-visible consequence: the connect hangs. State is `.discovering`, so `isConnected` and `isConnecting` are both false, discovery is live and autoconnect fires. The next `connect()` passed the `activeConnection != nil` guard and overwrote `connectionStepper` while the first machine was still suspended. If that connect's node dump opened the gate, the orphaned machine woke up and ran Steps 6 through 8 against the new connection: a second `.subscribed` transition, a second `initializeMqtt()`, a second `setupPeriodicHeartbeat()`, and a duplicate connect analytics action.

The `isRunning` leak is smaller but related. `didReceive(_:)` reads `connectionStepper?.isRunning` to decide whether to move the UI to `.discovering`, and `connect(to:)`'s new guard reads it too, so a leaked `true` would refuse later connects.

## How is this tested?

Two new suites in `MeshtasticTests/`, both verified to fail against the unfixed code and pass with the fix:

- `SequentialStepsCancellationTests.swift` drives `SequentialSteps` directly. The key test cancels a `.retryStep` step mid-flight with no external error recorded and asserts the machine does not enter the following step and that `run()` reports failure. Companions cover the external-error substitution, that a non-cancellation failure still retries the step in place, and that `isRunning` clears when `run()` throws out of the retry sleep.
- `AccessoryManagerDatabaseStepCancellationTests.swift` drives the real `AccessoryManager.connect(to:)` pipeline with a mock `Transport`/`Connection` pair, following the pattern in `AccessoryManagerConnectStopsDiscoveryTests.swift`. It parks Step 5 on its continuation, injects a `.disconnected` event on the connection's own event stream (the production trigger, not a hand-mirrored predicate), and asserts `connect()` returns promptly, ends in `.discovering` with `tooManyRetries`, and clears `connectionStepper`. A second test asserts the reentrancy guard refuses a connect while a machine is still running and leaves the in-flight one in place.

Reverting only `throw error` back to `break stepRetryLoop` fails the end-to-end test on four assertions and takes 10s instead of 0.2s.

Full run: `xcodebuild test -project Meshtastic.xcodeproj -scheme Meshtastic -destination 'platform=iOS Simulator,id=...'`. The 21 pre-existing `SwiftUIViewSnapshotTests` failures are unchanged and none were added; `StringExtensionDetailedTests.allCases_count` (`IntervalConfiguration` 11 vs 10) also fails on the base commit and is unrelated. SwiftLint on the changed files reports only the pre-existing `function_body_length` violation on `AccessoryManager+Connect.swift`, which was already over the limit on `main` (202 lines, limit 200) and is 206 here.

Not verified: real-hardware behaviour. Everything here was exercised against mocks and the simulator.

Known residual risk, out of scope for this PR and worth a follow-up: several steps (0, 2, 4, 6, 7, 8) still have no timeout, and `BLEConnection.send` parks on a continuation with no timeout and no cancellation handler. If one of those wedges, `run()` never returns, and the new guard turns that into a refusal of every later connect rather than the zombie-machine behaviour it replaced. Deliberately left alone here, along with the per-step `retryDelay` multiplication, Step 6's `.fail` behaviour, and the `lastConnectionError` reset at the top of `connect(to:)`, which are handled on separate branches.

None. No SwiftUI view is touched; the change is the connect state machine, its tests, and developer docs.

## Checklist

- [x] I have verified whether these changes require updates to the in-app documentation under `docs/user/` or `docs/developer/`, and updated accordingly. `docs/developer/transport.md` gains a "Step Failure and Cancellation Semantics" section under "Connection Sequencing", and the bundled copies were regenerated with `bash scripts/build-docs.sh --output Meshtastic/Resources/docs`. `Meshtastic/Resources/docs/index.json` churns non-deterministically on a pristine build, so only the `transport` entry is included.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved connection reliability when a previous connection attempt is still running.
  - Added a timeout for database configuration waits to prevent connections from hanging indefinitely.
  - Improved cancellation and retry handling so connection failures are reported correctly.
  - Ensured connection state is properly reset after success, failure, or cancellation.

- **Documentation**
  - Added guidance on connection failures, cancellations, retries, timeouts, and concurrent connection attempts.


garthvh:
Reverting the bad pull

PR #2202, opened 2026-07-25
fix(pax): create the sending node instead of orphaning the reading

garthvh:
## What changed?

`paxCounterPacket` now resolves the sender with `fetchedNode.first ?? findOrCreateNode(...)` — the same idiom `deviceMetadataPacket` and `telemetryPacket` already use — and always links the reading. Two regression tests: a PAX packet from an unheard node creates the node (with user stub) and persists one linked reading with no orphan rows; a known sender persists a linked reading as before.

## Why did it change?

Hearing a packet from an unheard node is how the app learns the node exists. The old code instead inserted the `PaxCounterEntity` first and only linked it if the sender was already in the database — an unknown sender left a permanently unlinked row that the next debounced save (from any packet type) swept into the store. Orphans are unreachable by any query (readings are fetched through `paxNode`) and untouchable by node purge, so noisy traffic grew dead data forever.

Replaces #2193, which guarded against the orphan by dropping the packet — the opposite of the pipeline's create-on-hear convention (see `findOrCreateNode`'s doc comment and `environmentTelemetryFromUnknownNode_createsNodeAndLinksReading` in the existing suite).

## How is this tested?

- New `PaxCounterPacketIngestTests` (2 tests) pass, alongside the neighboring `TelemetryPacketIngestTests` and `DeviceMetadataIngestTests` suites — 7/7.
- SwiftLint: no new warnings on touched files.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved PAX counter telemetry processing for previously unknown nodes.
  - Automatically creates and links node information when needed.
  - Ensures readings are saved correctly without creating duplicate or orphaned records.
  - Preserves accurate Wi-Fi and Bluetooth counter values for known nodes.

- **Tests**
  - Added coverage for PAX counter ingestion with both new and existing nodes.
  - Verified correct persistence, linking, and debounced saving behavior.

### DIFF
--- a/Meshtastic/Helpers/MeshPackets.swift
+++ b/Meshtastic/Helpers/MeshPackets.swift
@@ -799,6 +799,9 @@ actor MeshPackets {
 			let fetchedNode = try modelContext.fetch(fetchDescriptor)
 
 			if let paxMessage = try? Paxcount(serializedBytes: packet.decoded.payload) {
+				// Create the sending node when unheard, same as deviceMetadataPacket and
+				// telemetryPacket — an unlinked reading would persist as an orphan row.
+				let node = fetchedNode.first ?? findOrCreateNode(num: packetFrom, context: modelContext)
 
 				let newPax = PaxCounterEntity()
 				modelContext.insert(newPax)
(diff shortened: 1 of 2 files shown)


garthvh:
@RCGV1 the agents wild out if you tell them to harden things, there is a pattern for this already on every other packet type.

PR #2203, opened 2026-07-25
security: DEF CON hardening — malformed-packet crashes, MQTT MITM, App Intent confirmation

garthvh:
## Remote crash / crash-loop — the important ones (unauthenticated, any node in RF range)

- **Position packet single-packet crash (Critical).** `upsertPositionPacket` converted the `UInt32` fields (`seqNumber`, `satsInView`, `groundSpeed`, `groundTrack`, `precisionBits`) with the trapping `Int32(_:)`. One crafted `POSITION_APP` packet with a field > `Int32.max` — valid lat/lon to pass the coordinate check — deterministically `SIGTRAP`ped the app, and crash-looped on reconnect if it landed in the radio's node DB. Now `Int32(truncatingIfNeeded:)` (matching the rest of the file), with `groundTrack` range-checked on the `UInt32` before conversion.
- **Same trap class swept everywhere reachable** — the admin `getConfig`/`getModuleConfigResponse` upserts, `channel`, `hopsAway`, `batteryLevel`, `mapReportSettings.*`, etc. (~90 conversions). Enum `.rawValue` and already-safe conversions untouched.
- **SwiftData unique-collision crash (High).** `NodeInfoEntity.num`/`UserEntity.num` are `@Attribute(.unique)`; SwiftData dedups uniques only against the *saved* store, so two crafted packets from a new node inside the debounced-save window double-inserted and tripped a fatal assertion. `upsertNodeInfoPacket`, `createUser`, and `textMessageAppPacket` now go through the pending-insert-aware `findOrCreateNode`/`findOrCreateUser` (hardened `createUser` to be find-or-create, protecting all its callers).

## MITM / credential exposure (High)

- **MQTT TLS was trust-all** — `allowUntrustCACertificate = true` and the `didReceive` trust callback returned `completionHandler(true)` on *both* branches. Now rejects invalid/self-signed certs (`completionHandler(isValid)`, `allowUntrust = false`), closing the on-path MITM that captured broker credentials and injected spoofed downlinks. Also warns when credentials would be sent over a non-TLS connection.

## Silent mutation via App Intents (Medium)

- `SaveChannelSettingsIntent` / `AddContactIntent` now `requestConfirmation` before mutating radio state — mirroring the in-app QR/URL confirmation sheets and the destructive intents — so an untrusted Shortcut can't silently replace your channel list/PSK or inject a contact in the background.

## File-parsing / transport / info-disclosure

- Route-CSV import clamps lat/lon and rejects non-finite before scaling (no `Int32(Double)` trap).
- `MBTilesArchive` clamps attacker metadata zoom before `UInt8(...)`; `PMTilesExtractor` gets the overflow-safe offset arithmetic already applied to the reader in #2192.
- TCP reader skips zero-length frames instead of an undocumented `receive(0,0)`.
- Connect **Step 5a** gets a 120s timeout so a radio that withholds the DB-complete nonce can't wedge the connect flow forever.
- TAK logs redact full-precision coordinates (`privacy: .private`).

## Tests

- New `PositionPacketIngestHardeningTests`: an oversized-field position packet now truncates and persists instead of crashing.
- `PaxCounter` / `Telemetry` / `DeviceMetadata` / `ChannelURL` / `ShareContactURL` / `Entity` suites (26 tests) still pass — the ingestion-path changes didn't regress.

## Summary by CodeRabbit

* **New Features**
  * Added confirmation prompts before adding contacts and before saving channel/LoRa settings.
* **Bug Fixes**
  * Improved connection reliability with a bounded database-sync wait and safer TCP frame reading.
  * Reduced persistence issues by preventing duplicate entries and hardening extreme/malformed packet and config value handling.
  * Hardened map tile/route ingestion (zoom bounds, overflow-safe extraction, sanitized CSV coordinates).
  * Enforced MQTT TLS trust validation.
* **Privacy**
  * Redacted latitude/longitude details in diagnostic logs.
* **Tests**
  * Added coverage for POSITION packet ingestion hardening with oversized values.

### DIFF
--- a/Meshtastic/Accessory/Accessory Manager/AccessoryManager+Connect.swift
+++ b/Meshtastic/Accessory/Accessory Manager/AccessoryManager+Connect.swift
@@ -186,7 +186,11 @@ extension AccessoryManager {
 			}
 			
 			// Step 5a: Wait for end of WantConfig (database)
-			Step { @MainActor _ in
+			// Bounded like its sibling steps: without a timeout, a malicious/misbehaving radio
+			// that completes config but never sends the database-complete nonce would wedge the
+			// connect flow in .retrievingDatabase forever (no watchdog until Step 8). 120s is
+			// generous for a large legitimate node-DB dump.
+			Step(timeout: .seconds(120)) { @MainActor _ in
 				guard wantDatabase else {
 					Logger.transport.info("👟 [Connect] Step 4: wantDatabase = false, skipping waitForWantDatabase")
 					return
--- a/Meshtastic/Accessory/Transports/TCP/TCPConnection.swift
+++ b/Meshtastic/Accessory/Transports/TCP/TCPConnection.swift
@@ -104,6 +104,10 @@ actor TCPConnection: Connection {
 					// Logger.transport.debug("[TCP] startReader: Found magic byte, waiting for length")
 
 					if let length = try? await readInteger() {
+						// Skip zero-length frames (e.g. a crafted `94 c3 00 00`): calling
+						// receiveData(min:0,max:0) issues an undocumented empty socket read that
+						// can stall the reader loop. A real frame always has a positive length.
+						guard length > 0 else { continue }
 						let payload = try await receiveData(min: Int(length), max: Int(length))
 						switch FromRadioDecoder.classify(payload) {
 						case .decoded(let fromRadio):
--- a/Meshtastic/AppIntents/AddContactIntent.swift
+++ b/Meshtastic/AppIntents/AddContactIntent.swift
@@ -29,6 +29,12 @@ struct AddContactIntent: AppIntent {
 				let decodedString = contactData.base64urlToBase64()
 				if Data(base64Encoded: decodedString) != nil {
 					do {
+						// Confirm before adding a contact, mirroring the in-app
+						// AddContactConfirmationView — an untrusted Shortcut should not be able to
+						// silently inject a contact (and its public key) in the background.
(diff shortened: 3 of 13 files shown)


garthvh:
Addressed the CodeRabbit review in 61b2544d:

- **UpdateSwiftData / MeshPackets NodeInfo ingestion** — routed the remaining bare `UserEntity()` / `NodeInfoEntity()` inserts through the find-or-create helpers (`findOrCreateNode` / `findOrCreateUser`) in the complete-user and existing-node branches of `upsertNodeInfoPacket` and in `nodeInfoPacket`, so repeated packets can no longer enqueue duplicate rows for the same unique `num`.
- **precisionBits** — now clamped with `Int32(min(positionMessage.precisionBits, 32))` instead of `truncatingIfNeeded`, so an out-of-range value stays a valid precision and the history-prune logic no longer misreads it as reduced-precision.
- **NodeDB-dump heading** — range-checks `groundTrack <= 360` before persisting, mirroring `upsertPositionPacket`.
- **TAK bridge** — added `privacy: .private` to the `pli.latitudeI` / `pli.longitudeI` debug interpolations.
- **AppIntents** — `requestConfirmation` cancellations now propagate instead of being wrapped as generic failures (`AddContactIntent`, `SaveChannelSettingsIntent`).
- **Test** — scoped the oversized-UInt32 position assertion to the packet's node instead of the bare sentinel.

Build succeeds; the touched ingest suites (position hardening, PAX, telemetry, device metadata) pass.

PR #2205, opened 2026-07-25
feat(caps): global node/waypoint eviction + raise position history cap

garthvh:
Implements the resource caps deferred from the security-hardening work, with the policy now decided: bound the two attacker-growable, previously-uncapped local stores so noisy mesh traffic (or a deliberate flood of NODEINFO/WAYPOINT packets) can't grow on-disk data without limit.

## Caps
- **Nodes: 10,000**, evicting **least-recently-heard first** (`lastHeard` ascending; nil/never-heard sorts first). **Favorites are never evicted** — a user-starred node is explicitly kept even if it's the stalest.
- **Waypoints: 5,000**, evicting **oldest-`lastUpdated` first**.
- **Positions per node: 5,000 → 25,000** (existing per-node cap, raised).

## How it runs
Enforcement lives in `enforceEntityCaps()`, called from `savePendingChanges` — the single choke point every packet handler's save funnels through, so it catches nodes/waypoints created by any path without instrumenting each call site. It's throttled to once every 64 saves: a `COUNT` is cheap, and the sorted fetch+delete only executes when actually over cap (at steady state over cap, that's ~one eviction per new insert). Saves are already debounced under load, so the cost is negligible.

Eviction is split into `evictNodesIfOverCap(_:)` / `evictWaypointsIfOverCap(_:)` so the sort order and favorite-protection are unit-testable with tiny datasets rather than needing 10k+ inserts.

## Tests
`EntityCapEvictionTests` (3, all green):
- nodes evict least-recently-heard while **favorites survive even when oldest**,
- waypoints evict oldest-last-updated,
- under-cap is a no-op.

## Note
Favorite-protection for nodes is a small addition beyond the bare spec — evicting a user's starred node would be a clear UX bug, so favorites are excluded from the candidate set. Flagging it in case you'd rather cap strictly by recency regardless of favorite status.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added stricter automatic storage limits for saved nodes and waypoints, enforced periodically during save operations.
  * When caps are exceeded, the oldest non-favorite nodes and the least-recently-updated waypoints are evicted while favorites and the newest data are preserved.
  * Increased retained position history per node.
  * Prevents newly created nodes from being incorrectly evicted when creation and limit enforcement occur in the same transaction.
* **Tests**
  * Added in-memory eviction test coverage for over-cap, under-cap, and same-transaction scenarios.


garthvh:
Addressed in 75ef1eec. `evictNodesIfOverCap`/`evictWaypointsIfOverCap` now set `includePendingChanges = false` on the deletion-candidate fetch, so an entity created earlier in the same `savePendingChanges` transaction (a pending insert with a nil `lastHeard`/`lastUpdated`) is no longer a deletion candidate. The `fetchCount` stays inclusive of pending inserts for cap accounting, so enough persisted rows are freed and the post-commit total still lands at the cap. Added `newNodeInTransaction_survivesEviction` to cover the create-then-evict path.

PR #2207, opened 2026-07-25
feat(map): pause trace-route flyover under heavy mesh traffic

garthvh:
## What

Pauses the trace-route 3D **flyover** on the map when live mesh traffic is high enough that the guided camera flythrough would render uselessly slow/janky.

The flyover (`TraceRouteFlyover`) drives the `MKMapView` camera on **every display frame**. On a busy mesh that animation contends with the SwiftUI re-renders that packet ingestion triggers (node pins, etc.), so the flythrough stutters. When traffic is high we now hold it back instead of launching (or continuing) a doomed animation.

## How

**New `MeshTrafficMonitor`** (`Meshtastic/Accessory/Accessory Manager/MeshTrafficMonitor.swift`)
- Estimates inbound mesh packets/second. Fed one cheap `record()` per inbound mesh packet at the `.packet` dispatch in `AccessoryManager.processFromRadio`; resampled on a 1 Hz timer while a connection is active (`start()` on connect, `stop()` on disconnect).
- Exposes a smoothed `packetsPerSecond` (EWMA) as the underlying rate, and a debounced `isHighTraffic` gate.
- **Hysteresis + sustain:** trips **on** once the raw per-second rate holds at/above the high watermark for the sustain window, and trips **off** once it falls below a lower resume watermark. The sustain timer keys off the **raw** per-window rate (not the EWMA), so a single burst — whose smoothed tail lingers above the line for a second or two — can't trip the gate on its own.

**AccessoryManager** mirrors `isHighTraffic` onto `@Published isHighMeshTraffic` so views already observing the manager react without observing the nested monitor.

**MeshMapMK** (the route banner):
- Disables the flyover play button while traffic is high, and also guards `start()` so a flyover can never launch under heavy traffic.
- The stop button stays live so a running flyover can always be dismissed manually.
- Shows a localized reason under the banner: **"Flyover paused — mesh traffic is high"** (also surfaced as the play button's accessibility hint).
- An in-flight flyover is stopped once the gate trips (the monitor's sustain/hysteresis prevents flapping).

## Design choices (flagged for review)
- **Threshold:** high watermark 25 pkt/s, resume watermark 15 pkt/s, sustain 2 s. Single named constants with rationale comments in `MeshTrafficMonitor`. These are judgment calls — the mid-20s pkt/s line is where the flythrough starts visibly stuttering on device; tuning welcome.
- **Stopping an in-flight flyover:** chose to stop it (not just block new launches) when traffic stays high, since letting it grind is the worse experience. The 2 s sustain + asymmetric watermarks keep it from flapping.
- Prefer **disabling** over hiding the control, per the map's existing affordance style.

## Tests
`MeshtasticTests/MeshTrafficMonitorTests.swift` — 9 Swift Testing cases covering the rate/threshold/hysteresis logic deterministically (drive `record()` + `sample(now:)` with an injected clock, no real timer): quiet stays low, sustained heavy trips, sustain delay, single-burst ignored, recovery below resume watermark, mid-band hysteresis (both directions), and `stop()` reset. All pass.

## Verification
- `xcodebuild build` (Debug, iOS Simulator): **BUILD SUCCEEDED**.
- Unit tests on simulator: **9/9 passed**.

## Summary by CodeRabbit

* **New Features**
  * Trace-route flyovers now monitor inbound mesh traffic and automatically pause when traffic remains high.
  * The map banner shows **“Flyover paused — mesh traffic is high”** while paused.
  * Flyover controls are disabled appropriately during high-traffic pauses and resume when conditions improve.

* **Bug Fixes**
  * Prevented in-progress flyovers from continuing during sustained high mesh traffic, improving animation smoothness.

* **Tests**
  * Added deterministic unit tests covering the traffic threshold and pause/resume behavior.


garthvh:
Valid finding, fixed in e409f0d3. Moved `meshTrafficMonitor.start()` out of the top of `connect()` and into Step 1 (right after the connection is established), so it re-runs on every attempt — a retry's `closeConnection()` stops the monitor, and the successful retry now restarts it instead of leaving the flyover gate dead for the session. `start()` is idempotent; `closeConnection()` still stops it on real disconnect. Added a stop-then-start retry-lifecycle test to MeshTrafficMonitorTests.

PR #2208, opened 2026-07-25
fix: keep device image/link network off the BLE connect path (#2196)

### LINKED ISSUE #2196: 🐞 [Bug]: App freezes and fails Bluetooth sync on captive portals / zero-rated cellular networks
### Before submitting

### Firmware Version

2.7.16.2109

### What did you do?

1. Connect an iPhone to a Meshtastic node (e.g., SenseCAP T1000) via Bluetooth.
2. Downloaded offline maps for the region within the app.
3. Turn on Cellular Data on a SIM card with depleted data allowance where the carrier (e.g., Movistar) redirects all HTTP/HTTPS traffic to a captive portal / top-up page.
4. Open the Meshtastic app outdoors away from Wi-Fi.

### Expected Behavior

The app should handle network request timeouts/captive portal responses gracefully without blocking the main UI thread, background tasks, or the Bluetooth synchronization stack with the node.

### Current Behavior

The app attempts background network requests (NTP/API/Connectivity checks) that receive HTML captive portal responses instead of expected data formats. The app hangs, preventing Bluetooth sync and message transmission with the node. 

Workaround: Disabling Cellular Data specifically for the Meshtastic app in iOS Settings fixes the issue immediately and restores Bluetooth operation.

### Participation

### Additional comments

This is particularly critical in emergency or off-grid situations. When a user runs out of cellular data, the app enters an "internet limbo" where the iPhone thinks it has active data, rendering the portable node useless unless cellular access is explicitly toggled off for the app.

### Code of Conduct

bruschill:
## What changed?

`MeshtasticAPI.refreshBundledDevicesData()` is awaited inline by BLE connect Step 3 (`AccessoryManager+Connect.swift`). Despite the name, it did network work: one `URL.eTag()` HEAD request per device image — 82 of them — and then `importDeviceLinks()`.

- **The bundled seed is now local only.** Decode plus SwiftData upsert, nothing else. The network tail moved into `refreshDeviceImagesAndLinks()`.
- **One image/link pass instead of two.** `refreshDevicesAPIData()` no longer runs its own image and link phases; it delegates to the shared pass, which unions the bundled catalog with the live API device list so hardware present in only one of them still gets images. Launch, Reset Database, and connect Step 3b all go through `refreshDevicesPreferringAPI()` — API first, bundle-only pass on failure.
- **Step 3b restores images and links after a database clear**, detached so it stays clear of the step's 30s budget.

Files: `Meshtastic/API/MeshtasticAPI.swift`, `Meshtastic/Views/Settings/AppSettings.swift`, `Meshtastic/Accessory/Accessory Manager/AccessoryManager+Connect.swift`, `MeshtasticTests/MeshtasticAPIBundledSeedTests.swift`.

### Notes for review

- **Why Step 3 changed at all.** The first version of this PR left `AccessoryManager+Connect.swift` untouched, which was wrong. `clearDatabase` batch-deletes `DeviceHardwareImageEntity` and `DeviceLinkEntity` (`Persistence.swift`), and it runs mid-session on a foreign-database device switch (`AccessoryManager+FromRadio.swift`). Step 3's own comment promised hardware info *and* "I want one" links survive any database clear — and the links half lived in the network tail this PR moved out, so nothing restored them until the next launch. Step 3b now does, detached.
- **Dedup is on image file name, not platform + name.** The request URL derives from the file name alone, and `DeviceHardwareImageEntity` is keyed by `fileName` with a single `device` relationship, so a name shared by several platforms (3 in the current catalog; 82 entries collapse to 78 names) can only ever belong to one device row. Previously that was one fetch per platform racing to claim the row; now it's one fetch attached to the first device in catalog order. Strictly more deterministic, but it is a behavior change.
- **`isLoadingDeviceList` is now a refcount.** Two passes can overlap, and a plain `Bool` let whichever finished first clear the flag while the other was still running. It's also raised across the image pass, so the hardware views keep showing a spinner rather than "UNSET" artwork mid-download, and it's cleared from a `defer` so throwing paths can't pin it.
- **`private init` became `init(container:startupRefresh: Bool = true)`.** `@testable` doesn't reach `private`, and the flag suppresses the launch cascade so tests can exercise one function without detached work racing them. `shared` is still the only app-side entry point.
- **Cancellation is honest but limited.** `refreshDevicesPreferringAPI()` distinguishes `CancellationError` from a network failure so teardown doesn't kick off a second 78-request pass. But an unstructured `Task` doesn't inherit cancellation from its enclosing task, so nothing actually cancels this work today.

## Why did it change?

Fixes #2196.

Connect Step 3 has a 30s budget and, declaring no `onFailure`, inherits `.retryAll` — a timeout restarts the *entire* connect from step 0, up to `maxRetries = 2`, then fails and auto-reconnects into the same stall. `URL.eTag()` sets no timeout, so it inherits `URLSession.shared`'s 60s default.

On a captive portal or a zero-rated cellular link those requests neither succeed nor fail fast, so Step 3 can't finish inside its budget and the radio never syncs. That's exactly why the reporter's workaround works: disabling cellular for the app makes every request fail instantly, so the step completes.

### Not addressed here

The issue also describes HTML captive-portal responses being accepted as data. That reproduces at a 100% hit rate — a portal answering 200 + `text/html` + an ETag gets all 82 bodies written into `DeviceHardwareImageEntity.svgData`, because `processImage` gates only on the presence of an ETag header. It's independent of this fix and doesn't affect the connect path, so I'll file it separately along with the cleanup needed for already-poisoned rows.

## How is this tested?

**Repro harness.** A local server simulating the two real portal behaviours, driven by verbatim copies of `URL.eTag()` / `URL.data(timeout:)` and the real `DeviceHardware.json` (99 devices, 82 image entries).

Blackholed requests (zero-rated carrier silently drops):

```
t= 30s  imagesCompleted=  0/82   <-- BLE connect Step 3 timeout fires here
=== COMPLETED in 61.0s ===
```

Zero of 82 resolve before the 30s budget; nothing completes until `URLSession.shared`'s 60s default expires. That's a lower bound — measured on macOS, which allows 6 connections per host to iOS's 4, and without TLS handshake stalls against a real hijacked endpoint.

**Five tests in `MeshtasticAPIBundledSeedTests`:**

| Test | Guards |
|---|---|
| `bundledSeedIssuesNoNetworkRequests` | the seed stays local |
| `bundledSeedPopulatesDeviceCatalog` | the seed still fills the catalog |
| `offlineImageRefreshRequestsEachImageOnce` | one request per unique image, offline path |
| `apiRefreshCoversApiOnlyHardwareWithoutDuplicating` | API-only hardware covered, no double fetch, links imported once |
| `startupCascadeRunsOneImagePass` | a real launch does one pass |

Both load-bearing tests were checked against the pre-fix code rather than assumed to work. Re-adding the network tail to the seed makes `bundledSeedIssuesNoNetworkRequests` fail and list the `flasher.meshtastic.org` image URLs and `msh.to/api/urls`. Restoring the old double-pass init makes `startupCascadeRunsOneImagePass` fail with `(images.count → 156) == (expected.count → 78)`.

That last test drives the actual detached launch cascade rather than a single function, because "every startup repeats the work" was the specific concern. It stubs the API endpoint with the bundled catalog so the union is exactly the bundled set: one pass is N requests, the old double pass is 2N.

The `URLProtocol` recorder hooks `startLoading()`, not `canInit(with:)` — the loading system may call `canInit` several times while deciding who handles a request, which inflated one count 6x before I moved it.

**Suite comparison.** Full `MeshtasticTests` before and after produces an identical failure set (18 snapshot tests, `allCases_count`, `validNumCreatesUserWithDefaults`, 2 `CoreDataMigrationServiceTests` — all pre-existing on clean `main`). 2582 → 2587 tests. SwiftLint clean on all four files.

Not verified on a physical iPhone against a real carrier portal; the evidence above is the local harness plus the code path reading. `isLoadingDeviceList`'s throwing paths are verified by inspection only — they need a missing or corrupt app-bundle resource, which a hosted test can't induce.

Not applicable — no view, layout, or user-facing copy changed. The perceptible differences are timing only: device images and "I want one" links populate a moment after the catalog rather than blocking on it, and they now repopulate after a mid-session database clear instead of waiting for the next launch.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Improved device-catalog refresh during startup and “Clear App Data” by preferring API-based hardware updates, with automatic fallback to bundled image and link restoration.
* **Bug Fixes**
  * Fixed overlapping device-list loading so loading indicators don’t stop prematurely.
  * Strengthened throttling for device images and links to prevent redundant network requests, and ensured the throttle resets immediately after clearing data.
* **Tests**
  * Added/expanded end-to-end tests that record network activity, verify bundled vs API refresh separation, and confirm throttle and deduping behavior.


garthvh:
Pushed a small commit on top (maintainer edit, `07d2a30c`) that throttles the Step 3b image/link pass to once per 48h.

Rationale: Step 3b fires on every connect, and `processImage` issues a remote ETag HEAD per image (~78) before it even consults the cache, so a routine reconnect re-hits the network for data that hasn't changed. The commit gates the shared pass on a new `UserDefaults.lastDeviceImageAndLinkUpdate` timestamp (mirroring the existing `lastFirmwareAPIUpdate` throttle). `clearDatabase` (both implementations) resets that timestamp to `.distantPast`, so a factory/NodeDB reset or foreign-database device switch still forces an immediate restore — the whole reason Step 3b exists. That's safe because `processImage` falls back to the app bundle, so restoration never depends on the network; the window only bounds how often we re-check for updated artwork.

Added two throttle tests plus a suite `init()` that resets the timestamp so the existing network tests stay deterministic against the process-global store. Full suite (7 tests) green locally. Happy to adjust or drop it if you'd rather handle the throttle differently.

garthvh:
Pushed one more (maintainer edit, `ced29407`) that makes the detached Step 3b pass cancellable — the residual tail of #2196 the pass moved off the critical path but didn't eliminate.

On a captive portal the pass's ~78 image ETag HEADs each hang ~60s (`URL.eTag` sets no request timeout) *past* a disconnect, wasting network and pinning the hardware-list spinner. And the `catch is CancellationError` in `refreshDevicesPreferringAPI` was effectively dead: nothing cancelled the task, and async `URLSession` surfaces cancellation as `URLError.cancelled` (or, via `data(timeout:)`'s timer child, `CancellationError`) anyway.

- Hold the Step 3b `Task` on `AccessoryManager` and cancel it in `closeConnection` (a rapid reconnect also cancels the prior pass).
- The worker checks `Task.isCancelled` before claiming a throttle token, in the fetch loop, and before arming the throttle — so a torn-down pass does no more work and leaves the throttle un-armed, letting the next connect run a real restore (plays nicely with your `DeviceImageLinkThrottle` generation logic).
- `refreshDevicesPreferringAPI` now guards its fallback on `Task.isCancelled` (covers both cancellation error types) so teardown can't kick off a second 78-request pass.

Added a deterministic cancellation test (self-cancel before the pass runs — no scheduling race). Full suite green (10 tests). Happy to drop or adjust if you'd rather handle teardown differently.

PR #2215, opened 2026-07-27
fix(ble): retry a radio that is out of buffers instead of ending the session

bruschill:
## What changed?

A single BLE write the radio could not allocate a buffer for was treated as a fatal link failure, ending the session with no reconnect. Whatever operation was in flight died with it.

`CBATTError.insufficientResources` means the peripheral could not allocate for **this one write**. The link is healthy and the next write usually succeeds. But `didWriteValueFor` escalated every write error to `handlePeripheralError`, which classifies all `CBATTError`s as fatal with `shouldReconnect = false`, so the session ended.

- `send()` now retries `insufficientResources` up to 4 times with a short escalating backoff (120/240/360 ms), extracted into a `performWrite` helper.
- `didWriteValueFor` no longer escalates that specific error, so the retry gets a chance before any teardown. Every other error keeps its existing behaviour.
- It remains a case in `handlePeripheralError` as a **reconnecting** error, for the path where retries are exhausted.
- Each `ToRadio` write's size is now logged against the negotiated write limit, at a level that survives in `OSLogStore`.

## Why did it change?

Observed on a Heltec V4 (ESP32-S3, NimBLE) driving a device-profile import from an iPhone over BLE.

ATT MTU 255 was negotiated, so CoreBluetooth reported a 512-byte write ceiling. Writes of 8–33 bytes all succeeded; the first 104-byte one (`set_owner`) came back `CBATTError.insufficientResources`. Not a size limit and not a broken link — the radio was momentarily out of mbufs.

**This is not specific to config import.** Any admin message large enough to hit it is affected, including renaming a node and most config writes, so it may account for some reports of the app dropping its connection.

The size-vs-limit logging exists because an oversized payload and an out-of-buffers radio surface as the same opaque "resources are insufficient" at the peripheral, and nothing recorded which one it was. Diagnosing the original failure required adding this first. Note the levels are deliberately not `.debug`/`.info`: only notice-and-above are persisted to `OSLogStore`, which is what the in-app log viewer reads, so a `.debug` line here is invisible in the field.

## How is this tested?

- Full app build succeeds for the iOS 18 simulator on top of `main`.
- **On hardware:** reproduced the original failure on a Heltec V4 running 2.7.27, where a 104-byte `set_owner` was refused and the app ended the session. With this change the same operation retries and completes. Verified against the radio's own serial debug log, not just the app UI.
- Also exercised on a Heltec V4 running 2.8.0-develop and a Seeed T1000-E (nRF52840/SoftDevice). The retry path never triggers on nRF52 — SoftDevice does not produce this error — so it is inert there.

### Not tested

- **No controlled reproduction.** The allocation failure is transient and environmental; it was reproduced opportunistically, not on demand. The fix is validated by the operation succeeding afterwards rather than by forcing the error and observing the retry.
- The exhausted-retries path (all 4 attempts refused) has not been observed.
- ESP32-S3 and nRF52840 only.

## Related

- The underlying allocation failure is firmware-side. On 2.8.0/ESP32 the same radios continuously fail BLE notifications with `BLE_HS_ENOMEM`, reported as [meshtastic/firmware#11245](https://github.com/meshtastic/firmware/issues/11245) — likely the same buffer exhaustion seen from the notify direction. This change does not fix that; it stops the app treating a symptom of it as fatal.
- **Split out of #2023** (device configuration export/import), where this was found. There is no code dependency in either direction — this touches only `BLEConnection.swift`, and nothing in that PR references it, so the two can land in any order. Landing this first is preferable only because #2023's config-edit transaction raises the cost of a mid-operation transport failure: with a transaction open, a link that dies before the commit leaves the node holding it. This reduces how often that happens.

## Checklist

- [x] I have verified whether these changes require updates to the in-app documentation under `docs/user/` or `docs/developer/`. No user-facing behaviour changes, so no doc updates.

---

### Notes for reviewers

- **The retry is deliberately narrow.** Only `CBATTError.insufficientResources` is retried; every other `CBATTError` and `CBError` keeps its current handling. The concern worth checking is whether 4 attempts over ~720 ms is the right budget, and whether resending identical bytes can double-apply an operation. On nRF52 it cannot — `NRF52Bluetooth.cpp` deduplicates consecutive identical `ToRadio` writes — and on ESP32 the admin writes in question are idempotent, but that reasoning is worth a second opinion.
- **`handlePeripheralError` still treats every other `CBATTError` as fatal and non-reconnecting.** That may be too blunt for other codes too, but changing it without a specific observed failure seemed worse than leaving it, so this change stays scoped to the one error with evidence behind it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Bluetooth reliability by handling recoverable “radio out of buffers” (`insufficientResources`) failures for `.withResponse` `TORADIO` writes with a bounded retry loop and backoff.
  * Adjusted error handling so these cases no longer trigger premature teardown; the connection enters reconnect behavior instead of being treated as dead.
  * Added diagnostics for negotiated maximum write sizes and payload/limit awareness to better explain write failures.

* **Documentation**
  * Updated BLE transport docs with a new “BLE Writes When the Radio Is Out of Buffers” section, including retry/escalation, cancellation-safe behavior, and logging/diagnostics details.


bruschill:
Addressed all four review comments.

**Reconnect after the final resource-exhaustion failure** (`a419195b`) — valid, and it made the `.insufficientResources` case in `handlePeripheralError` dead code: `didWriteValueFor` bypasses that handler for this code, so nothing could reach the branch. `send` now calls `handlePeripheralError(error:)` on the exhausted path before propagating the original error, which is what the PR description already claimed the code did.

**Do not swallow task cancellation during retry backoff** (`a419195b`) — valid. `Task.sleep` now throws instead of `try?`, and each attempt calls `Task.checkCancellation()` first, so a cancelled send cannot put another write on the wire.

**Make the limit represent retries, not total attempts** (`a419195b`) — valid as a naming bug, fixed by renaming rather than by adding an attempt. The 4 attempts / 120-240-360ms backoff is the timing that was reproduced and validated on hardware, so the constant is now `writeAttemptLimit`, documented as the initial write plus three retries. Behaviour is unchanged.

**Scope the BLE-buffer-retry text to response writes** (`34dc5ca5`) — valid, and it applies to more than wording. `performWrite` resumes its continuation immediately for `.withoutResponse`, so no ATT error can come back and the retry is unreachable on that path. The section now says so, and notes that a `.withoutResponse` radio running out of buffers is invisible to the app and that backpressure is unhandled. No code change: the retry is correctly inert there, it was only the docs that overstated the scope.

The docs themselves are new here (`28bb055f`). `transport.md` already documents the inbound version of this rule — an invalid UTF-8 string field must not tear down a healthy stream — so the outbound equivalent belongs next to it. Bundle regenerated with `scripts/build-docs.sh`.

Validation: full app build succeeds for the iOS 18 simulator, SwiftLint clean on the changed file. No new hardware run since the retry timing is unchanged.

---

Unrelated CI note for maintainers: `Check docs coverage` was failing on this PR for a reason that has nothing to do with the change. `docs-staleness.yml` is advisory (its last step always exits 0), but on a fork PR the `GITHUB_TOKEN` is read-only, so `issues.createComment` throws `403 Resource not accessible by integration` and fails the job. It will fail on every fork PR that touches `Views/`, `Model/`, `Tips/`, or `Accessory/` without a `docs/` change. Wrapping the two `github-script` steps so a 403 degrades to a log line would fix it. I've kept that out of this PR as unrelated infra — happy to open a separate one.

bruschill:
One more on the cancellation fix (`186109b9`). The check at the top of the retry loop stops a cancelled send from issuing another write, but cancellation arriving *during* the final failed write lands past it, so an exhausted send could still call `handlePeripheralError` and start a reconnect. Cancellation there usually means the app is tearing the link down deliberately, so that reconnect would fight it. Checks once more before escalating.

PR #2227, opened 2026-07-31
fix: stop snapshot tests from modifying source files

### LINKED ISSUE #2213: 🐞 [Bug]: Snapshot tests write untracked PNGs into the source tree on every run
### Before submitting

- [x] I have updated to the latest version of the software to verify the issue still exists (reproduced on `main` @ de26cc46)

### Firmware Version

N/A — developer tooling bug, not firmware-dependent

### What did you do?

Ran the snapshot test suite, then the docs screenshot pipeline:

```
xcodebuild test -scheme Meshtastic ...   # runs MeshtasticTests
bash scripts/copy-snapshots.sh --output Meshtastic/Resources/docs/assets/screenshots
git status --porcelain
```

### Expected Behavior

A clean tree after a test run. Reference snapshots are committed artifacts, so running tests against an unmodified checkout should produce no untracked or modified files.

### Current Behavior

Two PNGs are written into the source tree on every run and never converge, so the tree has to be reset repeatedly:

```
?? MeshtasticTests/__Snapshots__/SwiftUIViewSnapshotTests/aboutMeshtastic.png
?? docs/assets/screenshots/formattingToolbar.png
```

**Root cause.** `assertViewSnapshot` in `MeshtasticTests/SwiftUIViewSnapshotTests.swift` (~L85-155) resolves the repo root from `#filePath` and records-on-missing, silently and unconditionally:

```swift
} else {
    // First run - record reference silently (test passes; verify visually)
    try pngData.write(to: snapshotFile)
}
```

Any test whose reference PNG is absent writes one into the working tree and still passes. Two tests hit that path.

**1. `docs/assets/screenshots/formattingToolbar.png` — an unresolvable loop.**

`SwiftUIViewSnapshotTests.swift:1732` writes it with `forDocs: true`. No file under `docs/**.md` references it and it is not listed in `docs/assets/screenshots/MANUAL_SCREENSHOTS`, so `scripts/cleanup-screenshots.sh` classifies it as orphaned and deletes it. That script is invoked unconditionally at the end of `scripts/copy-snapshots.sh`.

The cycle: test records it → `copy-snapshots.sh` runs → `cleanup-screenshots.sh` deletes it → next test run records it again. It can never be committed and always reappears after a reset.

**2. `MeshtasticTests/__Snapshots__/SwiftUIViewSnapshotTests/aboutMeshtastic.png`.**

Written by the test at `SwiftUIViewSnapshotTests.swift:1556`, but the reference was never committed. It is silently regenerated on every clean checkout.

**Related: three stale committed reference PNGs.** Two tests were switched to `forDocs: true` but their old `__Snapshots__` copies were left in the repo, and one has no test writing it at all. All three are dead files that will never be compared against:

| File | Status |
|---|---|
| `MeshtasticTests/__Snapshots__/SwiftUIViewSnapshotTests/batteryPluggedIn.png` | superseded by `docs/assets/screenshots/batteryPluggedIn.png` (`:493` uses `forDocs: true`) |
| `MeshtasticTests/__Snapshots__/SwiftUIViewSnapshotTests/securityVersionNag.png` | superseded by `docs/assets/screenshots/securityVersionNag.png` (`:431` uses `forDocs: true`) |
| `MeshtasticTests/__Snapshots__/SwiftUIViewSnapshotTests/signalBLE_all.png` | no test writes this name |

### Participation

- [x] I am willing to submit a pull request for this issue.

### Additional comments

Suggested fix, in two parts.

**Stop the churn:**

- `formattingToolbar` — decide its purpose. Either reference it from a docs page, or drop `forDocs: true` so it lands in `__Snapshots__/` and commit it. It is currently a docs asset that no doc uses.
- `aboutMeshtastic` — verify visually and commit the reference.
- Delete the three stale `__Snapshots__` PNGs above.

**Prevent recurrence.** The underlying defect is that record-on-missing is silent and unconditional. Gating it behind an opt-in env var means a normal test run never writes into the tree, and a missing reference fails loudly instead of quietly manufacturing an untracked file:

```swift
} else if ProcessInfo.processInfo.environment["SNAPSHOT_RECORD"] != nil {
    try pngData.write(to: snapshotFile)
} else {
    Issue.record("Missing reference snapshot \(name).png — re-run with SNAPSHOT_RECORD=1")
}
```

Two supporting changes:

- Decouple `cleanup-screenshots.sh` from `copy-snapshots.sh`. A script that deletes files another script's tests regenerate should not run implicitly as a side effect.
- Add a CI step after the test run that fails when `git status --porcelain` is non-empty. That converts "reset the tree again" into a one-time build failure.

### Code of Conduct

bruschill:
## What changed?

- Require `SNAPSHOT_RECORD=1` before snapshot tests create or replace reference PNGs.
- Add the missing `aboutMeshtastic` reference and move `formattingToolbar` to test-only snapshots.
- Remove three stale test-only references whose tests either write to the docs directory or no longer exist.
- Insert node-row snapshot fixtures into the shared SwiftData container so the views render instead of returning `EmptyView`.
- Run all 38 snapshot suites on pull requests and fail if they change reference files.
- Keep screenshot cleanup separate from `copy-snapshots.sh`.

## Why did it change?

A normal snapshot test run silently created missing PNGs and rewrote references when dimensions changed. `copy-snapshots.sh` also deleted an unreferenced docs snapshot, creating a record-delete-record loop.

The node-row snapshot fixtures had also been detached from SwiftData since the liveness guard in `98b5444e`, so 19 tests rendered one-point-high empty views and rewrote their references.

Fixes #2213.

## How is this tested?

- Reproduced the missing-reference bug on `upstream/main`: the test passed and created an untracked PNG.
- Confirmed the fixed helper fails on a missing reference without creating a file.
- Confirmed `TEST_RUNNER_SNAPSHOT_RECORD=1` records the two missing references.
- Ran the About and MessagePreview suites: 6 tests passed with no reference changes.
- Reproduced the node-row failure in a clean `upstream/main` worktree: 19 one-point-height failures and 19 rewritten references.
- Confirmed the fixed fixtures render all 19 rows and do not rewrite references.
- Recorded 16 canonical reference updates with the pull request workflow's Xcode 26.3/iOS 18.5 renderer: 15 repaired node-row snapshots and `ackErrors.png`.
- Replaced the stale Xcode 16 workflow pin after CI showed that Xcode 16.3 lacked its simulator runtime and Xcode 16.4 could not compile the current Translation APIs. The final Xcode 26.3 verify-only run passed without reference changes.
- Confirmed `copy-snapshots.sh` copied 99 referenced files without deleting an orphan sentinel.
- Ran `git diff --check`, shell syntax checks, and workflow suite discovery. The workflow finds all 38 suites.

New reference images:

- `MeshtasticTests/__Snapshots__/SwiftUIViewSnapshotTests/aboutMeshtastic.png`
- `MeshtasticTests/__Snapshots__/SwiftUIViewSnapshotTests/formattingToolbar.png`

## Checklist

- [x] I have verified whether these changes require updates to the in-app documentation under `docs/user/` or `docs/developer/`, and updated accordingly (see [copilot-instructions.md](../.github/copilot-instructions.md#in-app-documentation) for the view → doc page mapping). No user-facing documentation change is needed; the `skip-docs-check` label is applied.

## Summary by CodeRabbit

* **Bug Fixes**
  * Documentation builds are now validated for pull requests and main-branch updates.
  * Snapshot tests now fail clearly when reference images are missing or dimensions differ.
  * Test-generated snapshots no longer unintentionally modify documentation assets.

* **Documentation**
  * Clarified screenshot management and release behavior.
  * Screenshot copying no longer removes unrelated image files automatically.

* **Tests**
  * Snapshot validation now runs against a consistent simulator configuration and verifies documentation assets remain unchanged.


garthvh:
Reworked this onto current main (maintainer edit — thanks for the strong foundation here, the fixture diagnosis especially). What changed:

- **Mechanism**: #2234's `SnapshotReferenceStore` landed first and covers this PR's recording guard, so `assertViewSnapshot` now routes through it (read-only by default, `MESHTASTIC_RECORD_SNAPSHOTS=1` to record, per-platform reference dirs). The parallel `SNAPSHOT_RECORD` helper from this branch is dropped.
- **Kept intact** — the parts only this PR has: the SwiftData fixture insertions that revive the 19 dead node-row snapshot tests, their re-recorded references, the missing `aboutMeshtastic`/`formattingToolbar` references, the three stale test-directory duplicate removals (re-verified against the docs pipeline: their `docs/assets/screenshots` sources remain, and the new referenced-image guard in `build-docs.sh` from #2256 now enforces that class permanently), the PR snapshot-suite CI gate, and the `copy-snapshots.sh` cleanup decoupling.
- **Restored** the three `EventFirmware` imagesets (DEFCON/FAB/Hamvention) that an earlier upstream merge on this branch removed — they're the Connect-box event icons' bundled fallbacks, not snapshot artifacts.
- **CI gate suite count** updated to 39 (main gained a suite since the job was written), and the bundled docs screenshots regenerated from the updated sources.

Validation: all 39 snapshot suites — 127 tests — pass in verify mode, and the working tree stays byte-clean afterward, which is exactly the churn this PR set out to stop. Closes the loop on #2213 once merged.

PR #2244, opened 2026-08-09
fix(tak): gate V2 on confirmed firmware

RCGV1:
## Summary

- send legacy TAK packets until the active radio confirms firmware 2.8.0 or newer
- reject missing or malformed version metadata instead of sending an incompatible V2 packet
- cover unknown, malformed, legacy, and supported firmware versions

## Testing

- `MeshtasticTests/TAKCapabilityTests` was selected in the iOS simulator test runner
- test compilation is currently blocked before execution by an unchanged upstream compiler error in `Meshtastic/Views/Settings/Config/SecurityConfig.swift:54`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TAK protocol compatibility detection.
  * Devices with missing, malformed, or unknown firmware versions now correctly fall back to TAK v1.
  * TAK v2 is recognized only for well-formed firmware version 2.8.0 or newer, including supported hexadecimal build suffixes.

* **Tests**
  * Added coverage for unknown, malformed, legacy, and supported firmware versions.

* **Documentation**
  * Updated TAK protocol guidance to explain firmware requirements and version handling.

### DIFF
--- a/Meshtastic/Accessory/Accessory Manager/AccessoryManager.swift
+++ b/Meshtastic/Accessory/Accessory Manager/AccessoryManager.swift
@@ -1107,10 +1107,24 @@ extension AccessoryManager {
 	/// ATAK_PLUGIN port (72) with the original `TAKPacket` schema, which only
 	/// supports PLI and GeoChat (no shapes, markers, routes, etc.).
 	///
-	/// Returns `true` when the firmware version is unknown (radio not yet
-	/// handshook) since v2 is now the predominant firmware in the field.
 	var supportsTAKv2: Bool {
-		checkIsVersionSupported(forVersion: "2.8.0")
+		Self.isTAKv2Supported(firmwareVersion: connectedVersion)
+	}
+
+	static func isTAKv2Supported(firmwareVersion: String?) -> Bool {
+		guard let firmwareVersion else {
+			return false
+		}
+
+		let components = firmwareVersion.split(separator: ".", omittingEmptySubsequences: false)
+		guard components.count >= 3,
+			  let major = Int(components[0]),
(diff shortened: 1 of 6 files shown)


RCGV1:
@coderabbitai review

RCGV1:
@coderabbitai review

RCGV1:
@coderabbitai review

RCGV1:
@coderabbitai review

garthvh (approved):

Meshtastic/Accessory/Accessory Manager/AccessoryManager.swift
```diff
@@ -1107,10 +1107,35 @@ extension AccessoryManager {
...
+		}
+		if components.count == 4 {
+			let hexDigits = CharacterSet(charactersIn: "0123456789abcdefABCDEF")
+			guard !components[3].isEmpty,
+				  components[3].unicodeScalars.allSatisfy(hexDigits.contains) else {
+				return false
+			}
+		}
+
+		return major > 2 || (major == 2 && (minor > 8 || (minor == 8 && patch >= 0)))
```
garthvh, line 1138: Verified on a local merge with current main: the upstream compile error that blocked your test run is gone, and `TAKCapabilityTests` passes 4/4. One simplification — the parse already guarantees a non-negative patch, so the `patch >= 0` term is always true and the comparison reduces to:

```suggestion
		return major > 2 || (major == 2 && minor >= 8)
```

Same semantics (behavior is pinned by the tests); it just stops implying a patch-level threshold that does not exist. Tried to push this directly but your fork rejected the branch push, hence the suggestion.

PR #2261, opened 2026-08-13
tvOS map: standalone selection halo, hot-path perf, settings polish

bruschill:
## What changed?

Three focused changes to the tvOS mesh map:

1. Moved node-list sorting and coincident-spread calculations off the hot path. Cached lists now refresh when membership changes and on a two-second tick.
2. Added a standalone, non-clustering selection halo with the node's colored short-name marker at its center. Selecting a node from the side list moves the camera to it.
3. Replaced the segmented map-type control with an inline picker that works better with the tvOS focus engine.

## Why did it change?

SwiftData models compare by identity, so `onChange(of: allNodes)` does not fire for in-place mutations. Nodes that gained positions after the initial dump could remain absent from the map.

MapKit hides pin views when it absorbs them into clusters. A halo attached to the pin therefore disappeared in dense meshes, while a cluster badge appeared at the cluster centroid instead of the selected node's location. A standalone annotation keeps the selection visible at the correct coordinate.

## How is this tested?

- Built the `Meshtastic TV` scheme for the tvOS Simulator after syncing with `main`.
- Tested on a physical Apple TV 4K against a 500-node conference-stress simulated mesh at 140 packets per second. The halo and node marker remain visible, the camera centers on selection, and the list stays live during sustained ingestion.

None.

## Checklist

- [x] I verified that this change does not require in-app documentation updates; this PR uses the `skip-docs-check` label.

## Summary by CodeRabbit

* **Improvements**
  * Node lists and map data now refresh more reliably when device information changes.
  * Selected nodes are highlighted with an animated ring for easier visibility.
  * Node selections remain active even when location data is unavailable.
  * Map markers update more reliably as nodes move or disappear.
  * Map selection zoom is tighter and responds more quickly.
* **Style**
  * Updated the map type selector to use a compact inline picker.


dzienisz:
WDYM tvOS? Meshtastic App works on TVs?

PR #2262, opened 2026-08-13
feat(import): cancellation-aware transport sends + shielded commit

bruschill:
## What changed?

- Made BLE writes cancellable without double-resuming their continuations. Each write-with-response continuation is keyed by UUID; cancellation and the CoreBluetooth callback race to remove the same entry.
- Made TCP sends cancellable without closing the shared `NWConnection`, so a cancelled import can still send its final commit.
- Shielded `commit_edit_settings` from inherited task cancellation. A cancelled import stops applying sections, then commits the completed subset instead of leaving the radio in deferred-save mode.
- Reduced the import sheet's force-dismiss fallback from 90 seconds to 15 seconds.
- Included the deterministic XcodeGen group-ordering fix from #2261 so the committed project passes the drift check while that PR is pending. I will update this branch after #2261 lands so the overlap drops from the diff.
- Updated the user import guide, transport internals, and XcodeGen project-generation documentation, then rebuilt the bundled in-app HTML and Markdown.

## Why did it change?

The import Cancel button previously took effect only between sends. If a BLE or TCP send was suspended, the UI could remain stuck until the transport callback or the 90-second fallback fired.

Cancelling the transport alone is not enough. Meshtastic configuration imports use a firmware edit transaction, and an interrupted transaction still needs `commit_edit_settings`. The commit therefore runs in a fresh task that does not inherit the cancelled state.

## How is this tested?

- `xcodebuild` build-for-testing passed for the iOS Simulator.
- Six focused tests in `DeviceProfileImportCancellationTests` and `WriteContinuationStore` passed. They cover mid-import cancellation, skipped remaining sections, the shielded commit, FIFO continuation handling, targeted cancellation, and disconnect draining.
- Verified on hardware over BLE with a full-profile import. Cancelling during the apply phase stopped additional section writes in under a second, sent one commit, saved once, rebooted as expected, and produced no post-reboot configuration difference from the committed subset.
- SwiftLint, docs coverage, and the XcodeGen drift check pass on the fork PR.
- Rebuilt all 31 in-app documentation pages; canonical and bundled Markdown are byte-identical, and the bundle remains below its 10 MB limit.
- `TCPConnectionSendTests` passed.

Not applicable.

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved cancellation handling during device profile imports, reducing unnecessary packet sends and delays.
  - Prevented Bluetooth and TCP operations from hanging or completing more than once during cancellation or disconnection.
  - Ensured pending connection operations finish with appropriate errors when disconnected.
  - Improved reliability when sharing TV functionality with the visionOS app.

- **Documentation**
  - Expanded guidance for imports, cancellation behavior, transport operations, and shared TV project setup.

- **Tests**
  - Added coverage for import cancellation, commit behavior, and connection operation handling.


garthvh:
Merged main (the visionOS target removal made this branch's copy of the Vision group fix obsolete — that whole block is gone now) and regenerated the project, so the conflict is resolved. Also updated the transport docs the remaining CodeRabbit comments were anchored to: the text still described the pre-review single-slot TCP send and the tombstone-free BLE store. It now documents the keyed/tombstone designs the review-fix commit implemented, with the bundled copies rebuilt via build-docs.sh.

PR #2264, opened 2026-08-14
Fix crashes and wrong reconnects when switching nodes

garthvh:
## Summary

Switching between nodes could crash the app (Datadog `324bff02`, since 2.7.13), and when the connect to the new node failed the app went back to the old one instead of retrying.

## What changed

The switch flow was tearing down and recreating the SwiftData container. SwiftUI keeps observers attached to the old container after that, and the next database save crashes — often without leaving a crash report. The container is now created once at launch and never replaced. A switch clears the data in place and imports the backup into the same container, behind a "Switching Radios" placeholder so no views are on screen while the data changes underneath them.

Other problems found and fixed along the way:

- Restores from a backup written by an older schema failed because migration can't run on a read-only file. Backups now open through a temp copy.
- Store files were being deleted while still open, which breaks the open SQLite connection. Old files are left in place and cleaned up at next launch.
- The preferred device only updated after a successful connect, so a failed switch fell back to the old node and dumped its data on top of the restored database. It now updates when you pick the device.
- Nodes still on event firmware after the event ends now get the event sheet with an Update Firmware button on connect (once per connection). The end-date check existed but nothing used it.

## Testing

32 automated switch cycles between two replay radios on the simulator using my real database and backups, no crashes (previously crashed within 1-2 cycles). The `-switch-stress N` debug harness is included for regression testing. Verified on my phone against real nodes.

### DIFF
--- a/.github/workflows/docs-deploy.yml
+++ b/.github/workflows/docs-deploy.yml
@@ -50,12 +50,19 @@ jobs:
             exit 1
           fi
 
-          xcodebuild test \
+          # tee the full log so a failure can surface the actual error lines — a bare
+          # `tail -20` swallowed the file/line of compile failures (e.g. the
+          # "unable to type-check this expression in reasonable time" incidents).
+          if ! xcodebuild test \
             -workspace Meshtastic.xcworkspace \
             -scheme Meshtastic \
             -destination 'platform=iOS Simulator,name=iPhone 16,OS=18.5' \
             "${snapshot_tests[@]}" \
-            2>&1 | tail -20
+            2>&1 | tee /tmp/xcodebuild-snapshots.log | tail -20; then
+            echo "── error lines from full log ──" >&2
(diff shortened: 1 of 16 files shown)


garthvh:
CodeRabbit review dispositions (f5c3bb5b):
- **SwitchStress non-positive cycles trap** — fixed, clamped to ≥1.
- **Escalation path double container recreation** — fixed via `rebindToCurrentContainer()` (repoint minus the recreate); the escalation now recreates exactly once.
- **preferredPeripheralNum not moved with the Id** — fixed, moves at switch initiation (0 = unknown for a never-seen radio).
- **SecurityConfig post-save persistence across a device switch** — not taken: with the container now process-lifetime, the captured context can no longer belong to a retired store outside the rare destroy escalation; the mid-switch save race predates this PR and is tracked separately.

PR #2265, opened 2026-08-14
fix: split SecurityConfig type checking

bruschill:
## What changed?

Split `SecurityConfig` into smaller private `some View` properties and moved the inline change-handler bodies into private methods. The existing modifier order, state, bindings, save behavior, and view hierarchy are unchanged.

The existing helper methods now live in a same-file extension so the primary view stays below the SwiftLint `type_body_length` limit.

## Why did it change?

Xcode 26.3 could not type-check the original `SecurityConfig.body` expression within its time limit. The diagnostic run in #2263 identified the whole body at `SecurityConfig.swift:54:22`.

Each private view property gives the compiler an opaque boundary. The named change handlers also keep closure inference out of the SwiftUI modifier expression.

## How is this tested?

- `Deploy Docs / build-snapshots` passed under Xcode 26.3: [run 31766703893](https://github.com/meshtastic/Meshtastic-Apple/actions/runs/31766703893/job/94663996549).
- Xcode 27 simulator build passed with `-warn-long-expression-type-checking=50`; `SecurityConfig.swift` produced no expression timing warning above 50 ms.
- SwiftLint passed with zero violations.
- SourceKit reports no diagnostics in the changed file.
- A fresh review compared every moved handler and modifier against `main`; no behavioral differences were found.

## Screenshots/Videos (when applicable)

Not applicable. The rendered view is unchanged.

## Checklist

- [x] My code adheres to the project's coding and style guidelines.
- [x] I have conducted a self-review of my code.
- [x] I have commented my code, particularly in complex areas.
- [x] I have verified whether these changes require updates to the in-app documentation under `docs/user/` or `docs/developer/`, and updated accordingly (see [copilot-instructions.md](../.github/copilot-instructions.md#in-app-documentation) for the view → doc page mapping). No documentation changes are needed; the `skip-docs-check` label is applied.
- [x] I have tested the change to ensure that it works as intended.

PR #2266, opened 2026-08-14
Preserve firmware-provided LoRa presets

ayysasha:
## What changed?

- Preserve a legal modem preset supplied by firmware when a region is selected.
- Restrict the US Long Turbo recommendation to factory-fresh nodes whose stored and current preset are both stock Long Fast.
- Separate explicit region-change recommendations from compatibility-map refreshes, which now only repair illegal selections.
- Add regression coverage for firmware-provided and unsaved presets, map refreshes, and illegal-preset repair.
- Update and regenerate user and developer documentation.

## Why did it change?

The app treated every node with an unset region as stock and locally replaced its preset with Long Turbo when US was selected. Custom firmware can intentionally ship with an unset region and another legal preset, such as Short Turbo. In that case, the picker showed Long Turbo even though the radio and other clients still reported Short Turbo; pressing Save could then overwrite the firmware-provided setting.

This keeps the intentional stock Long Fast → Long Turbo recommendation while following the region-preset contract: preserve a legal current preset and use the advertised regional default only when the current preset is illegal.

## How is this tested?

- Added Swift Testing cases covering:
  - stock US Long Fast → Long Turbo, with and without compatibility metadata;
  - preservation of firmware-provided Short Turbo;
  - preservation of an unsaved legal selection;
  - compatibility-map refresh without replaying the recommendation;
  - repair of an illegal preset after map refresh;
  - existing old-firmware, custom-mode, non-US, and legality behavior.
- Regenerated the bundled documentation twice and confirmed deterministic output.
- Ran `git diff --check` successfully.
- Ran an independent 12-case transition model successfully.
- Native Xcode/SwiftLint execution is unavailable on the Windows/WSL development host; GitHub/macOS CI is the authoritative compile, lint, and Swift Testing gate.

Not applicable; this corrects preset-selection state without changing UI layout or strings.

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved LoRa modem preset selection when changing regions.
  - Preserves existing compatible presets, including firmware-provided and unsaved selections.
  - Newly flashed US-region devices using Long Fast are now recommended to use Long Turbo.
  - Prevents compatibility-map refreshes from unexpectedly changing valid presets.
  - Automatically repairs incompatible presets using the region’s default option.

- **Documentation**
  - Updated user and developer guidance to reflect the revised US-region preset behavior.

- **Tests**
  - Expanded coverage for preset preservation, recommendations, refreshes, and invalid selections.


ayysasha:
CI note: the Deploy Docs snapshot build reaches macOS compilation but fails in unchanged `Meshtastic/Views/Settings/Config/SecurityConfig.swift` with “The compiler is unable to type-check this expression in reasonable time.” This is the same current upstream failure tracked by #2265 and visible on PRs #2257, #2262, #2261, and others; the failing command lists `LoRaConfig.swift` only as part of the batch, while the primary failed source is `SecurityConfig.swift`. Branch-specific checks pass: SwiftLint, docs coverage, XcodeGen drift, CodeRabbit, CLA, and merge freeze.

garthvh:
This needs to be managed in the firmware before the 2.8 alpha release

PR #2280, opened 2026-08-15
Mesh stats strip on the Apple TV map

garthvh:
## Summary

A compact strip over the TV map showing the connected mesh's health at a glance: nodes online/total, channel utilization, air-util TX, and packet counters. Settings gets a toggle (default on) and a Top/Bottom position picker.

## What changed

- **MeshClient** now decodes `TELEMETRY_APP` packets, which it previously dropped. Only packets from the connected node are used — other nodes' LocalStats describe their radios, not this session. LocalStats fills every cell; DeviceMetrics and the connected node's NodeInfo entry fill the two load fields as a fallback, so numbers appear at connect time instead of waiting out the first telemetry interval. Stats are held in memory on the client — session data, and adding fields to `MeshNode` would risk the store's wipe-and-recreate migration fallback.
- **MeshStatsStrip** (new) renders the cells over the map with the same material as the header scrim. Channel utilization colors at the firmware's congestion thresholds (warning ≥ 25%, error ≥ 50%), air TX warns at the 10% duty-cycle ceiling, and a relative "updated" timestamp ages the data. It is display-only by hard rule — nothing focusable may sit right of the side list (see the focus-trap comments in MeshTVMapView).
- **MapScreen** attaches it as an overlay on the map column only, at the top or bottom title-safe margin. Overlay rather than safeAreaInset so toggling it never resizes the MKMapView and re-triggers its region framing.
- **SettingsView** gains the toggle and position picker in the Map section, using the same duplicated-`@AppStorage`-key pattern as `tv.mapType` (`tv.statsBar.enabled`, `tv.statsBar.edge`).
- **RootView** gains a DEBUG-only `-tv-connect host:port` launch argument so the TV app can be driven headlessly in the simulator (MarketingCapture pattern) — that's how this was verified.

## Testing

Against the DEFCON replay on a tvOS 26.5 simulator: strip verified in all three states (top, bottom, hidden) with node counts live from the store (56/64 online). The replay never emits LocalStats, so packet counters show an em dash there and the full-cell rendering is covered by the view's #Preview mocks; real LocalStats end-to-end needs a physical radio. Focus regression checked — the strip introduces no focusable content, so remote navigation is unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added an optional mesh statistics strip to the map, displaying node counts, channel utilization, transmissions, packet totals, duplicates, and telemetry age.
  * Added settings to enable or disable the strip and position it at the top or bottom of the map.
  * Added visual warnings for high channel utilization and clear indicators when metrics are unavailable.
  * Added a debug launch option to connect to a TV client using a host and port.

### DIFF
--- a/Meshtastic TV/App/MapScreen.swift
+++ b/Meshtastic TV/App/MapScreen.swift
@@ -30,6 +30,10 @@ struct MapScreen: View {
 	/// the representable's updates; it publishes nothing until a region is downloaded.
 	@StateObject private var offlineVectors = OfflineVectorTileProvider()
 
+	/// Mesh stats strip visibility + position, shared with SettingsView via the same keys.
+	@AppStorage("tv.statsBar.enabled") private var statsBarEnabled = true
+	@AppStorage("tv.statsBar.edge") private var statsBarEdge: StatsStripEdge = .top
+
 	/// All nodes for the side list: located first, then alphabetically.
 	private var sortedNodes: [MeshNode] {
 		allNodes.sorted {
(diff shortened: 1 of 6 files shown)


garthvh:
Both review comments taken (see the latest commit): load fields are now optional so a partial DeviceMetrics packet can't fabricate a 0.0% reading, and the `-tv-connect` hook validates the port range.

PR #2289, opened 2026-08-15
Preserve legal preset when selecting US region

### LINKED ISSUE #2273: presetToSelect forces Long Turbo on US for a factory-fresh node, ignoring a deliberate preset
Tom says "Clod, remember to be nice when you make the request"

### What happens

Community/vendor firmware builds can pin a modem preset (`USERPREFS_LORACONFIG_MODEM_PRESET`) while leaving the region unset, so the user picks their own region on first connect. A fresh flash of such a build reaches the app as **region unset + Short Turbo** (or whatever the mesh runs on).

In `LoraConfigEnums.swift`, `ModemPresets.presetToSelect(…)` returns Long Turbo for that node the moment US is selected:

```swift
if factoryFresh, region == .us,
   regionInfo == nil || regionInfo?.presets.contains(.longTurbo) == true {
    return .longTurbo
}
```

`currentPreset` isn't consulted on this path, so a legal, deliberate Short Turbo is replaced with Long Turbo — even though the fall-through below it (`if info.presets.contains(current) { return nil }`) would have kept it, Short Turbo being legal in US.

The user gets no indication. The node ends up on different airtime parameters and a different frequency slot from the mesh it was flashed for, and simply sees nobody.

Scope is narrower than the Android equivalent, which I've filed separately: the US guard means non-US regions already fall through to "keep if legal", and `guard supports2_8` means pre-2.8 radios are untouched. So this bites specifically on **US + 2.8 firmware**.

### Why it's reasonable today

For a stock install the rule is right: an unset region normally carries a meaningless Long Fast placeholder, and the app's US convention is Long Turbo. The app has no way to tell a placeholder from a deliberate preset, because the firmware doesn't currently say.

### Suggestion

Two parts, and the first stands alone:

1. **Let the existing legality check run first.** If `currentPreset` is already legal in the newly-selected region, keep it, and apply the US → Long Turbo convention only when it isn't (or when there's no current preset). That preserves the rule for genuinely fresh nodes without discarding a deliberate choice.

2. Firmware side, filed at meshtastic/firmware: have a build that pins a preset advertise it as the unset region's single entry in `LoRaRegionPresetMap`, as an explicit statement of intent. `loRaRegionPresets[.unset]` would then be non-nil precisely when the preset is deliberate, making the distinction exact rather than inferred. Stock builds keep emitting nothing for the unset region, so existing behaviour and the "absent ⇒ no constraint" rule in `docs/developer/lora-region-presets.md` are unchanged.

Happy to send a PR for (1) if the approach looks right to you.

bruschill:
## Summary

A fresh firmware 2.8 node can carry a deliberate modem preset from a community or vendor build. Selecting US replaced it with Long Turbo, which could move the node off its mesh without warning. At the same time, the US default for 2.8 stays Long Turbo — Long Fast's bandwidth is not US-compliant, and we want new US nodes moved off it. Fixes #2273.

## What changed

- Selecting US on a factory-fresh node still migrates stock Long Fast (or no preset) to Long Turbo.
- Any other preset the firmware advertises as legal in the selected region is kept — a deliberate Short Turbo survives.
- An illegal preset still falls back to the region's advertised default.
- Long Fast remains selectable in the US, with a compliance warning under the preset picker.
- User docs updated to match.

## Testing

LoRaPresetSelectionTests and ModemPresetsTests: 24 tests covering the stock-Long-Fast migration, deliberate-preset preservation, missing region map, non-US regions, and pre-2.8 firmware. iOS build passes; docs regenerated with scripts/build-docs.sh.


garthvh:
Maintainer edits pushed (87c4ad04): the deliberate-preset preservation is right, but as written it also stopped the factory-fresh US Long Turbo default from ever firing — a fresh node always reports stock Long Fast, which the region map lists as legal, so the keep-legal rule short-circuited the migration. For 2.8 the US default is Long Turbo because Long Fast's bandwidth is not US-compliant, so the rule order now migrates stock Long Fast (or no preset) first, then keeps any other legal preset. Long Fast also stays selectable in the US with a compliance warning under the preset picker instead of being blocked. Tests updated (24 passing) and docs regenerated.

PR #2303, opened 2026-08-16
Add the offline terrain layer spec

garthvh:
## Summary

Spec for adding hillshade and contour lines to offline map regions using Mapterhorn elevation extracts — the same PMTiles bbox-extraction flow the offline basemap already uses. Rendering stays MapKit: hillshade as a locally computed raster tile overlay, contours as polyline overlays. 3D terrain and MapLibre are explicitly out of scope.

## What changed

Adds specs/018-offline-terrain-layer/spec.md: two delivery phases (terrain download + hillshade, then contours), map-type behavior (hybrid/satellite get contours only), edge cases (partial coverage, add-on downloads for existing regions, tile-edge continuity, attribution), and the measurable outcomes.

## Testing

Docs only.

### DIFF
--- a/specs/018-offline-terrain-layer/spec.md
+++ b/specs/018-offline-terrain-layer/spec.md
@@ -1,0 +1,119 @@
+# Feature Specification: Offline Terrain Layer
+
+**Feature Branch**: `docs/offline-terrain-spec` (spec only; implementation branches follow)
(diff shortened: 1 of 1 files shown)


garthvh:
Superseded by the full spec kit PR.

PR #2313, opened 2026-08-16
Use the tvOS Back button to disconnect

bruschill:
## What changed?

- Handle the tvOS Back/Menu command on the connected map screen.
- Pop Settings or node details when they are open, and ask for confirmation before disconnecting at the map root.
- Remove the on-screen Disconnect button. Confirming the alert disconnects and returns to the connection screen.

## Why did it change?

Pressing Back from the connected map root exited to the tvOS Home Screen. Disconnect was also a separate list action. The remote button should keep the user in the app and provide the disconnect path with confirmation.

## How is this tested?

- Built the `Meshtastic TV` scheme for the tvOS simulator.
- Ran an isolated tvOS UI test that used `XCUIRemote.shared.press(.menu)`, confirmed the alert, selected Disconnect, and verified the connection screen. 1 test passed with 0 failures.
- Captured the connected map, confirmation alert, and returned connection screen on an Apple TV 4K 1080p simulator against a replay node.
- Ran SwiftLint on `Meshtastic TV/App/MapScreen.swift` with 0 violations.
- Ran `git diff --check`.

Captured on an Apple TV 4K 1080p simulator against a replay node:

- Connected map with the Disconnect button removed.
- Back-button disconnect confirmation.
- Connection screen after confirming Disconnect.

## Checklist

There is no in-app documentation page mapped to the Meshtastic TV target. The `skip-docs-check` label is applied.

### DIFF
--- a/Meshtastic TV/App/MapScreen.swift
+++ b/Meshtastic TV/App/MapScreen.swift
@@ -13,6 +13,11 @@
 import SwiftUI
 import SwiftData
 
+private enum MapNavigationDestination: Hashable {
+	case settings
+	case node(UInt32)
+}
+
 struct MapScreen: View {
 	@Bindable var client: MeshClient
 	@State private var selectedNodeNum: UInt32?
@@ -22,7 +27,8 @@ struct MapScreen: View {
 	// focus for NavigationLink rows, so track it explicitly and mirror it into
 	// the map selection (debounced map-side).
 	@FocusState private var focusedNodeNum: UInt32?
-	@State private var navPath: [UInt32] = []
+	@State private var navPath: [MapNavigationDestination] = []
+	@State private var showingDisconnectConfirmation = false
 
 	/// The persisted node store — the map and list read from here, not the client.
 	@Query private var allNodes: [MeshNode]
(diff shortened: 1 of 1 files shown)


garthvh:
Thanks for this — the implementation is sound, and I want to be clear that the concern here isn't correctness. We checked the stranding risk carefully (Menu at the map root, at node detail, at depth 2, and with an alert already up) and the outer handler and the NavigationStack never both act on one press, so back-to-pop and back-to-disconnect don't collide. That part works.

I'm closing it for a platform-behavior reason instead.

On tvOS, Menu from an app's top-level screen is supposed to return to the Home Screen. This traps it at the map root permanently. The PR treats the current behavior as the bug, but exiting to Home is the correct system behavior, and it only backgrounds the app — it doesn't drop the radio connection. That's the kind of HIG deviation that draws an App Review rejection, and we've already been bitten by one of those on the Catalyst side.

Two smaller things if you want to revisit this:

- Removing the Disconnect row leaves disconnect with no visible affordance, and since it's also the only route back to ConnectView to pick a different node, it's worth keeping the row and treating Back as an additional shortcut rather than a replacement.
- `onMenuExit`/`escapeMap` in `MapScreen` feed `MeshTVMapView.menuPressed`, which isn't attached to any gesture recognizer and sits behind a non-interactive, non-focusable map — it's unreachable, so it can go.

A version that keeps Menu-to-Home at the root and adds an explicit disconnect affordance would be welcome. Also, ignore CodeRabbit's suggestion to wrap `client.disconnect()` in a Task — it's synchronous and non-throwing.

PR #2315, opened 2026-08-17
Fix tvOS accessibility

bruschill:
## What changed?

- Add explicit VoiceOver focus and labels for connection and discovery errors.
- Expose node online state and telemetry warning severity without relying on color.
- Start Settings on Map Type and return empty maps to the Re-center control instead of leaving focus trapped.
- Mark branding images as decorative and improve telemetry fallback values.
- Add a tvOS UI test target with an accessibility audit and Settings focus coverage.

## Why did it change?

The tvOS app had several accessibility gaps: errors could appear without being announced, node and telemetry states were not fully exposed to VoiceOver, warning severity relied on color, and focus could start on a destructive control or remain trapped on an empty map.

No documentation update is needed because this changes accessibility semantics and focus behavior without changing the user workflow.

## How is this tested?

- Built the `Meshtastic TV` scheme for a tvOS 26.5 simulator with `build-for-testing`.
- Ran the new tvOS UI test target: 2 tests passed with 0 failures.
- Ran Xcode's accessibility audit on the connection screen.
- Verified Settings initially focuses a Map Type segment instead of Clear Node Database.
- Ran SwiftLint on every changed Swift file.
- Regenerated with XcodeGen 2.46.0 twice and confirmed an identical project hash.
- Ran `git diff --check`.

Not included. The changes primarily affect VoiceOver output and focus behavior, which are covered by the tvOS UI tests.

## Checklist

## Summary by CodeRabbit

* **Accessibility Improvements**
  * Improved VoiceOver announcements and focus handling for connection, discovery, map, settings, and node information.
  * Added clearer status descriptions for packet activity, channel utilization, airtime, and update times.
  * Marked decorative branding images so they are skipped by assistive technologies.
* **Bug Fixes**
  * Restored focus more reliably when navigating maps and returning from errors.
* **Tests**
  * Added automated accessibility coverage for connection and settings screens.

### DIFF
--- a/Meshtastic TV/App/ConnectView.swift
+++ b/Meshtastic TV/App/ConnectView.swift
@@ -11,9 +11,15 @@
 import SwiftUI
 
 struct ConnectView: View {
+	private enum ErrorFocus: Hashable {
+		case connection
+		case discovery
+	}
+
 	@Bindable var client: MeshClient
 	@Environment(\.scenePhase) private var scenePhase
 	@StateObject private var discovery = NodeDiscovery()
+	@AccessibilityFocusState private var errorFocus: ErrorFocus?
 
 	@AppStorage("tv.lastHost") private var host: String = ""
 	@AppStorage("tv.lastPort") private var portText: String = "4403"
@@ -46,6 +52,9 @@ struct ConnectView: View {
 						Section {
 							Label(message, systemImage: "exclamationmark.triangle.fill")
 								.foregroundStyle(Color("MeshtasticError"))
+								.accessibilityLabel("Connection error: \(message)")
+								.accessibilityFocused($errorFocus, equals: .connection)
+								.onAppear { errorFocus = .connection }
 						}
 					}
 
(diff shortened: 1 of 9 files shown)


garthvh:
Merged — this shipped in #2319, which is your commit rebased onto current main with your authorship intact. Closing this one only because the branch point had drifted; the work itself is in.

This PR was branched before `MeshStatsStrip` was rewritten by #2307, #2314 and #2316, so that one file was the only conflict and needed reworking against the current layout. Three differences from what's here, so you know what changed:

- The `cell` helper keeps its `LocalizedStringKey` signature, and the packets metadata stays a single row rather than going back to separate Dupes and Updated cells.
- The severity glyph sits beside the column *label* instead of the value. Ch Util and Air TX are capped at 140pt now, and an icon in the value row pushed them back into the truncation #2316 had just fixed. VoiceOver output is unchanged, since the severity is read from the accessibility value rather than the glyph.
- Accessibility labels and values in the strip use `LocalizedStringKey`/`Text` rather than composed `String`s, so they resolve through the string catalog the TV target picked up in #2307. Worth keeping in mind for future accessibility strings — `accessibilityLabel("...")` with a literal localizes, but a built-up `String` silently won't.

Everything else went in untouched: the `@AccessibilityFocusState` handling for connection and discovery errors, decorative branding images, node row accessibility values, Settings starting on Map Type instead of Clear Node Database, the empty-map focus fallback, and the new tvOS UI test target. Both of your tests pass, and `performAccessibilityAudit()` on the connection screen is a good addition — thanks for that.

PR #2321, opened 2026-08-17
Remove unused tvOS map Menu handler

bruschill:
## What changed?

- Removed the unreachable tvOS map Menu callback path.
- Removed an unused duplicate Disconnect button view.
- Kept the visible Disconnect row and native tvOS Back/Menu handling unchanged.

## Why did it change?

This is a follow-up to closed PR #2313. Review there confirmed that the map callback could not run because the map is non-interactive and no press recognizer calls it. Removing the path avoids implying that the map handles Menu presses while preserving the system behavior at the map root.

## How is this tested?

- Built the `Meshtastic TV` scheme for a generic tvOS simulator.
- Ran SwiftLint on both changed files. It reported one existing `large_tuple` warning outside this diff and no serious violations.
- Ran `git diff --check`.
- Ran `/ponytail-review`: `Lean already. Ship.`

Not included. This removes unreachable code and does not change the visible UI.

## Checklist

No documentation update is needed because this is a deletion-only cleanup with no user-visible behavior change.

### DIFF
--- a/Meshtastic TV/App/MapScreen.swift
+++ b/Meshtastic TV/App/MapScreen.swift
@@ -48,7 +48,6 @@ struct MapScreen: View {
 				nodes: locatedNodes,
 				selectedNodeNum: $selectedNodeNum,
 				recenterToken: recenterToken,
-				onMenuExit: escapeMap,
 				offlineVectors: offlineVectors
 			)
 			.ignoresSafeArea()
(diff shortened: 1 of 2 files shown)


garthvh:
Merged as #2330 — your commit rebased onto current main with authorship intact. This one had gone conflicting after #2326 touched the same file, and the conflict was inside `escapeMap` itself.

Worth flagging what that conflict exposed: #2319 had put its empty-map focus fallback (`recenterFocused = true` when no nodes are located) *inside* `escapeMap`, and since that function is unreachable, the behaviour has never actually executed. Deleting it loses nothing that was running, but if you still want it, it needs wiring to a real trigger — happy to review that as its own change.

Thanks for the follow-up; I confirmed the path is dead before merging (no gesture recognizer calls `menuPressed`, and the map is both non-interactive and non-focusable).

PR #2328, opened 2026-08-18
Back off repeated firmware notifications

garthvh:
## Summary

A firmware notification that repeats on a timer nagged indefinitely. Repeats now back off.

Reported from the field on Burning Man event firmware: "Firmware Notification / Warning / Location sharing is disabled on this channel", arriving over and over.

## What was happening

`handleClientNotification` promoted **every** `ClientNotification` to a local notification with no throttling. That string is generated on the device — it isn't in this repo — and event firmware that deliberately turns position sharing off for a channel emits it on every position interval. So the app was faithfully relaying configuration as a repeating warning.

## What changed

- **Backoff per notice**: immediate, then 5 minutes, 30 minutes, 2 hours, then once every 12 hours. A genuine one-off still alerts with no delay; a standing condition settles to once or twice a day. A single flat window can't do both — short enough to report a real fault promptly is short enough to nag.
- A notice unseen for 24 hours is forgotten, so a recurrence after a real lull alerts immediately, and the history map can't grow unbounded.
- **Security notices bypass backoff entirely** — duplicate public key, low-entropy key, and the key-verification notices are rare and actionable, and burying them is the failure mode worth avoiding.
- **Notice identity** comes from the structured `payloadVariant` when present, so a reworded firmware string stays the same notice; otherwise level plus message text.
- **Per-message notification identifiers.** The previous fixed `"client.notification"` meant an unrelated notice could silently replace a security warning that was still pending.
- Logging stays unconditional, so Debug Logs records suppressed notices. The goal is fewer alerts, not less information.

State is in memory on `AccessoryManager`, so after a relaunch a standing notice alerts once more — which is the useful behaviour.

## Testing

7 new tests in `FirmwareNoticeBackoffTests`, all passing: the ladder's timing step by step, distinct notices not sharing a backoff, security notices never suppressed, the quiet-day reset, key derivation preferring the payload variant, and identifier bounds. One test simulates firmware repeating every three minutes for a day and asserts a handful of alerts rather than the 481 you would get today.

## Not addressed

Tapping still goes to Settings → Debug Logs, which isn't the useful destination for a position-sharing notice — the channel's settings would be. Routing by message text is fragile, so that wants a payload variant from firmware first. Worth raising upstream that an intentional configuration shouldn't be emitted as a repeating warning at all.

## Summary by CodeRabbit

* **Enhancements**
  * Improved firmware notifications with smarter repeat handling.
  * Repeated ordinary notices are shown at progressively longer intervals to reduce interruptions.
  * Notification history is maintained separately for each notice, preventing unrelated messages from being suppressed.
  * Stale notification history is automatically cleared after an extended quiet period.
  * Security-related notices continue to bypass suppression and are delivered immediately.


rancur:
Heads up on an overlap from #2358 / #2359, in case it's useful here.

While tracing why Meshtastic has no per-app notification entry in the Watch app, I went through every notification the app can emit. The firmware-update notice is one of four with no user-facing off switch at all — the others being the radio client alerts in `handleClientNotification`, Traceroute Complete, and TAK Route Received.

Backoff and an off switch seem complementary rather than competing: this PR stops the same notice repeating, and a toggle would let someone who doesn't want firmware notices at all opt out. Not suggesting scope creep here — just flagging it so the toggle work, if it happens, lands on top of your backoff rather than beside it.

No file conflict with #2359, which is Watch-target only. If a follow-up adds those toggles it would touch `AccessoryManager+FromRadio.swift`, so I'd rebase onto this after it merges.

garthvh:
Both review findings addressed: the notification message and the backoff key that embeds it are logged private (radio-provided text can carry location or key material), and identifier fragments append a digest of the complete key so normalization and truncation collisions can't make one pending notice replace another. Branch merged up to current main; full suite passes at 2,953 in 516.

PR #2332, opened 2026-08-19
Fix frequency override display precision (amended from #2329)

garthvh:
## Summary

@bruschill's #2329 with the documentation claim corrected and the tests tightened. Original commits and authorship preserved. Supersedes #2329; the code fix is unchanged and correct.

## Why this exists

The fix itself is right and worth merging: `TextField(value:formatter:)` writes the re-parsed display string back into the binding that gets saved, so the old 3-and-4-digit formatters were **corrupting stored frequencies**, not just drawing them short. Measured across 20,000 random floats per band, the old LoRa formatter lost data on roughly 39% of values at 868/915 MHz and 85% at 144 MHz; the licensed-operator path was worse at 94–98%. The new formatter round-trips losslessly everywhere.

What needed fixing was the claim around it.

## What changed relative to #2329

- **Docs reworded** in all three copies. "Shows up to six decimal places" overstates the field: `overrideFrequency` is a 32-bit `Float`, spacing about 61 Hz at 915 MHz and 244 Hz at 2.4 GHz, so five digits is the most that carries information on any Meshtastic band — `Float(869.441650) == Float(869.441651)`. Now says the stored frequency is shown without rounding, which is both true and the point.
- **Formatter documented**: why six digits is headroom rather than resolution, why grouping is off (a de_DE user saw "2.412" for 2412 MHz), and why it must stay a computed property. Turning it into a stored `let` to avoid the per-init allocation would convert each caller's `locale` assignment into shared global state, silently breaking the tests.
- **Tests** use frequencies the field can hold, and add two cases the original missed: that the **default locale** round-trips (the existing test pins pt_PT, so it would pass even if `Locale.current` were broken), and that zero renders as "0" — zero means "no override" per `LoRaChannelCalculator`.

## One consequence worth your call

Writing the tests surfaced something neither the PR nor my review had noted: at six digits, a value **typed as 433.175 redisplays as 433.174988**, because that is what the `Float` actually holds. It is pinned in a test rather than hidden.

This is the visible cost of the lossless round trip, and it is strictly better than the old behaviour, which silently rounded the stored value. But it does look like a bug to a user who typed a round number. Exactly-representable values (906.875, 2412, 0) show clean; others show noise.

If that bothers you, the fix is shortest-round-trip formatting rather than a fixed digit count — it gives `433.175`, `869.44165`, `906.875`, `2412` with no noise **and** no data loss. That means replacing `NumberFormatter` with manual locale handling, so it is a follow-up rather than something to fold in here.

## Testing

iOS builds. 3 tests in `FrequencyOverrideFormatterTests` pass, 3 executed (verified not a false green).

Worth knowing: this suite **never runs in CI**. The only workflow invoking `xcodebuild test` restricts `-only-testing` to the 39 snapshot suites, so it compiles but never executes on Actions. Repo-wide gap, not this PR's — but the regression test won't catch a regression by itself.

## Summary by CodeRabbit

* **Improvements**
  * Frequency Override values now display with up to six decimal places without rounding or grouping separators.
  * Frequency values can be entered and displayed consistently across LoRa and Licensed Operator settings.
  * Improved preservation of exact frequency values when formatting and parsing.

* **Documentation**
  * Added and clarified guidance for specifying custom frequencies in MHz.


garthvh:
Not needed — pushed the same commit directly onto #2329's branch instead, so the fix lands in @bruschill's PR with its review history intact rather than as a competing one.

Issue #2336, opened 2026-08-19
🚀 [Feature Request]: nRF52 bootloader upgrade + flash-level factory erase, at parity with Android


jamesarich:
### OS

iOS, iPadOS

### Description

Filed while auditing the two apps' firmware-update featuresets ahead of a docs update ([meshtastic/meshtastic#2609](https://github.com/meshtastic/meshtastic/pull/2609)) and the [OTAFIX bootloader fork migration](https://github.com/meshtastic/Meshtastic-Android/pull/6769). Two gaps found, verified against source on both sides (not guessed):

**1. No OTAFIX-aware bootloader upgrade.** Meshtastic-Android resolves and writes a per-board, sha256-pinned OTAFIX bootloader UF2 image directly to the device's mass-storage drive, keyed by the `Board-ID` read from `INFO_UF2.TXT` (`feature/firmware/.../MaintenanceUf2.kt`, `otafixUf2ForBoardId`). Surfaced in-app as **Advanced → Firmware Update → Upgrade bootloader**. Grepped this repo for `bootloader`/`OTAFIX` in `Firmware.swift` and the UF2 views — no hits. iOS users have no in-app way to get onto [Meshtastic's OTAFIX bootloader](https://github.com/meshtastic/Adafruit_nRF52_Bootloader_OTAFIX) (faster/more reliable BLE OTA) short of manually downloading and drag-and-dropping a UF2 file themselves.

**2. No flash-level factory erase (device-recovery path).** Android's "Erase and reinstall" (`MaintenanceUf2.kt`, `eraseUf2For`) writes a dedicated erase UF2 image to the mass-storage drive — this works even on a device that won't boot into the application, because it's a bootloader-level operation. This repo's `AppIntents/FactoryResetNodeIntent.swift` sends an **admin-protocol command** (`sendFactoryReset`) to an already-connected, running node. That's a different mechanism with a different failure mode: it requires the firmware to be alive and connected, so it cannot recover a bricked or corrupted device the way Android's flash-write erase can.

Also worth noting (not a gap, just different implementations of the same capability): both apps already do native in-app BLE OTA *firmware* DFU without a third-party app — Android via a hand-rolled protocol (`ota/dfu/SecureDfuProtocol.kt` etc.), Apple via Nordic's `NordicDFU` SPM package (`Views/Settings/Firmware/NRF DFU/DFUModel.swift`, reachable from **Settings → Firmware Updates**). No action needed there.

[...]

PR #2338, opened 2026-08-20
Add an OTAFIX bootloader upgrade flow for nRF52 boards

garthvh:
## Summary

First half of #2336. Android can upgrade a device onto Meshtastic's OTAFIX bootloader (faster, more reliable BLE OTA) in-app; iOS had no path at all short of finding a UF2 on GitHub by hand.

## What changed

The firmware screen gains a Bootloader section for nRF52 boards OTAFIX supports, opening a guided sheet in the style of the existing UF2 mass-storage flow:

1. Reboot the connected device into DFU (or double-tap reset) and connect its USB drive.
2. Pick the mounted drive in the file picker. The app reads `INFO_UF2.TXT` off the drive and identifies the board from its `Board-ID:` line.
3. The matching self-update image is downloaded from the pinned OTAFIX release, verified against its pinned SHA-256, and written onto the drive. The device installs it and reboots itself.

The safety model is Android's, unchanged: the Board-ID the device itself reports is the only input that selects an image — product names can't be correlated between the projects, and the XIAO BLE / BLE Sense split differs only by Board-ID, where writing the wrong image is unrecoverable without a debug probe. The connected node's platformio target only gates whether the action is offered. An unrecognized Board-ID, a missing `INFO_UF2.TXT`, or a checksum mismatch each refuse with nothing written.

The board map (14 boards) is mirrored verbatim from Meshtastic-Android's `MaintenanceUf2.kt`, and every pinned digest was re-verified against the actual release assets before committing.

New files land in synced folders, so no project file changes.

## Testing

- 11 new tests: the audited board set is pinned exactly, per-row URL/filename/digest integrity, the XIAO variant split resolves to distinct images, INFO_UF2.TXT parsing (CRLF, case, trimming, missing/empty lines), unknown-board refusal, and SHA-256 gating.
- Full unit suite: 2,912 tests in 509 suites passing. iOS Simulator and Mac Catalyst builds succeed.
- Not tested end-to-end against hardware: the drive flow needs an nRF52 board in DFU mode on USB-C. The flow mirrors the shipping UF2 firmware flow's mechanics (same picker, same drive), but a RAK4631 walk-through before release would be the real proof.

The erase/recovery half of #2336 is not in this PR.

## Summary by CodeRabbit

* **New Features**
  * Added a guided bootloader upgrade flow for supported nRF52840 devices.
  * Detects compatible hardware, reboots into upgrade mode, identifies the device, and installs the appropriate bootloader image.
  * Verifies downloaded images before installation and reports progress, completion, and errors.

* **Bug Fixes**
  * Prevents upgrades for unsupported devices, invalid drives, unrecognized hardware, and corrupted downloads.

* **Tests**
  * Added coverage for device detection, image selection, metadata parsing, supported targets, and checksum validation.

### DIFF
--- a/Meshtastic/Model/Firmware/MaintenanceUF2.swift
+++ b/Meshtastic/Model/Firmware/MaintenanceUF2.swift
@@ -1,0 +1,174 @@
+//
+//  MaintenanceUF2.swift
+//  Meshtastic
+//
+//  Copyright(c) Garth Vander Houwen 8/19/26.
+//
+//  A pinned, content-verified UF2 image used by a maintenance flow (bootloader
+//  upgrade). Unlike release firmware, these images have no versioned upstream
+//  the app can resolve against, so each one is pinned to an exact URL and
+//  SHA-256 digest, and the digest is verified before the image is ever offered
+//  for writing. Mirrors Meshtastic-Android's MaintenanceUf2.kt so both apps
+//  ship the same audited pairings.
+//
+
+import CryptoKit
+import Foundation
+
+struct MaintenanceUF2: Sendable, Equatable {
+	let url: URL
+	let fileName: String
+	/// Lowercase hex SHA-256 of the exact bytes at `url`.
+	let sha256: String
+
+	/// True when `data` hashes to the pinned digest. Any mismatch means the
+	/// upstream bytes changed (or the download was tampered with) and the image
+	/// must not be written.
+	func matches(_ data: Data) -> Bool {
+		let digest = SHA256.hash(data: data).map { String(format: "%02x", $0) }.joined()
+		return digest == sha256
+	}
+}
+
+/// OTAFIX bootloader self-update images and the rules for resolving one safely.
+///
+/// The safety contract, unchanged from Android: the **Board-ID the device itself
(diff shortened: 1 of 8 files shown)


bruschill:
Forgot to open the PR last night so you beat me to it, but I have the second half in my PR: https://github.com/meshtastic/Meshtastic-Apple/pull/2339

garthvh:
Addressed five of the eight review findings:

- The install task is stored and cancelled before drive access is released, with Task.checkCancellation between the download and the write — a dismissed sheet can no longer write against a revoked security scope.
- INFO_UF2.TXT is size-capped (16 KB) before being read into memory.
- The write-failure log marks the error private; it can carry the drive path.
- The tests now assert every Board-ID against an explicit audited fixture (board slug in the file name plus digest prefix) and compare the supported-target set exactly — previously two swapped assets passed every structural check, which is the exact failure the map exists to prevent.
- Added the bootloader upgrade flow to docs/user/firmware.md and rebuilt the bundled docs.

Skipped three with reasons: the layout literals, caption text styles, and button construction all match the shipped UF2MassStorageView sheet this screen is styled after — a shared-token or legibility pass belongs to both sheets together, not this PR.

PR #2339, opened 2026-08-20
Add nRF52 bootloader upgrade and flash-level factory erase

### LINKED ISSUE #2336: 🚀 [Feature Request]: nRF52 bootloader upgrade + flash-level factory erase, at parity with Android

bruschill:
## What changed?

- Add nRF52 bootloader upgrade and flash-level factory erase actions to Firmware Updates.
- Select OTAFIX images from the mounted volume's `Board-ID` and verify their pinned SHA-256 before writing.
- Select the factory-erase image from the reported S140 version. The bundled images erase without waiting for a USB serial connection.
- Require a second UF2 pass to reinstall the prepared application after either maintenance action.
- Persist recovery state before each write. Interrupted or indeterminate writes resume forward and are not retried.
- Validate the application image, UF2 family, target range, board identity, SoftDevice, and second-pass volume before writing.

Ordinary UF2 installation and Nordic BLE DFU are unchanged.

## Why did it change?

Closes #2336.

The Apple app had no equivalent to Android's OTAFIX bootloader upgrade or bootloader-level flash erase. These operations are needed when an nRF52 radio cannot recover through the running firmware.

Factory erase intentionally removes the radio's owner, channels, identity keys, settings, and node database. The app does not restore them automatically. Both maintenance actions require the application reinstall pass before the recovery record can be completed.

## How is this tested?

Physical T1000-E validation from an iPhone:

- Read `Board-ID: nRF52840-T1000-E-v1` and `SoftDevice: S140 7.3.0` from `INFO_UF2.TXT`.
- Installed OTAFIX once, then completed the mandatory application reinstall.
- Confirmed bootloader `0.9.2-OTAFIX2.2-BP1.4` and firmware `2.8.0.597f676` afterward.
- Ran factory erase once, then completed the mandatory application reinstall.
- Confirmed the owner, region, channels, identity keys, settings, and node database were erased while OTAFIX remained installed.
- Restored the secured pre-test radio profile afterward and confirmed an empty configuration diff.
- Used a separately signed validation app. Removed it after testing and confirmed both existing Meshtastic apps remained installed.

Automated and build validation:

- Full iOS test run: 2,946 tests passed, including all 16 `UF2MaintenanceTests`.
- iOS `build-for-testing` passed.
- Mac Catalyst build passed.
- SwiftLint passed on the changed Swift files.
- XcodeGen output is stable and `git diff --check` passes.
- Both bundled erase images match their documented SHA-256 and contain valid UF2 blocks.

The other supported boards and the S140 6.1.1 erase path are covered by catalog, hash, address, and recovery tests but were not tested on physical radios.

Not included. The complete flow was validated on the physical radio and iPhone described above.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added UF2 firmware maintenance for supported nRF52840 devices, including bootloader upgrades and factory resets.
  * Added guided firmware recovery with progress tracking, volume validation, reconnection checks, cancellation protection, and resumable interrupted operations.
  * Added support for bundled factory-erase firmware images and firmware maintenance actions in Settings.

* **Bug Fixes**
  * Added comprehensive validation for firmware files, device compatibility, integrity, and write destinations.

* **Documentation**
  * Documented supported factory-erase images, build inputs, versions, addresses, and checksums.


garthvh:
Rebuilt this branch down to the erase flow on top of the merged #2338 infrastructure — the bootloader half is on main now, so this PR carries only the part of #2336 still missing.

What changed from the original approach:

- The erase uses the commit-pinned images from meshtastic/web-flasher — the same nrf_erase2.uf2 / nrf_erase_sd7_3.uf2 Android ships for its own write-to-drive flow, which is in production. The self-built binaries, build script, and patch are dropped: shipping self-patched flash erasers in the repo is a trust model we don't want, and Android's flow demonstrates the canonical images work without a serial connection. Both digests and both first-block flash addresses re-verified against the live pinned files.
- The safety contract is yours and Android's, kept intact: image selected only by the SoftDevice line the drive reports, digest verification, plus a first-UF2-target-address cross-check against the reported SoftDevice before writing — the swapped-row mistake a digest alone cannot catch.
- The recovery-record state machine and mandatory-reinstall gating are gone. The sheet mirrors the shipped bootloader flow: pick the drive, confirm the destructive action, write, done — the drive reappearing is the state, and the done screen points at the normal firmware flow for reinstall.

The T1000-E validation from the original description remains the right hardware test for this: erase, confirm the wipe with OTAFIX intact, reinstall via Firmware Updates. Full suite passes: 2,970 tests in 518 suites, including the new NRF52FactoryEraseTests pinning the images, the SoftDevice parser refusals, and the UF2 header check.

PR #2343, opened 2026-08-20
Read the OTAFIX board map from api.meshtastic.org

jamesarich:
## What changed?

`MaintenanceUf2ManifestPayload` now fetches the OTAFIX board-ID map from
meshtastic/api's new `resource/maintenanceUf2` endpoint (meshtastic/api#121)
instead of the hardcoded dictionary this branch had. A bundled copy still
ships as a fallback for the first launch or when the network is
unreachable, and a fetch is only trusted if its bytes match a compile-time
SHA-256 pin — a mismatch is ignored and the app keeps whatever it already
had, never partially-trusted bytes.

## Why did it change?

This branch's board map was mirrored verbatim from Android's
`MaintenanceUf2.kt`, and Android was about to do the same thing this PR
does. Centralizing it in api.meshtastic.org means a new OTAFIX release is
one JSON edit instead of two separate hand-copies (three once web-flasher
gets its own OTAFIX flow).

Rebased/redirected before merge rather than after, so this never ships as a
third hand-copy of the same table.

## How is this tested?

`xcodebuild build-for-testing` and the full MeshtasticTests suite pass
(2917 tests). `MaintenanceUf2ManifestSeedTests` specifically checks: the
bundled seed's digest matches the pin, a fetch with a mismatched digest is
rejected and the current manifest is kept, and a fetch with a matching
digest is applied.

## Checklist

- [x] I have verified whether these changes require updates to the in-app documentation under `docs/user/` or `docs/developer/`, and updated accordingly. No doc update needed — internal data-source change, no user-visible behavior change.

## Summary by CodeRabbit

* **New Features**
  * Added automatic maintenance firmware metadata updates during app startup.
  * Maintenance firmware information now refreshes periodically, with validation to prevent invalid updates.
  * Added support for updated board mappings, release details, and firmware asset locations.

* **Bug Fixes**
  * Preserves existing firmware metadata when updates fail or contain invalid data.
  * Prevents excessive refresh attempts through update timing controls.


garthvh:
Reworked this to match the app's other API processes and merged current main (including #2338).

- The base64-encoded Swift seed blob is gone. The store seeds from the audited Swift map #2338 shipped — human-readable rows, pinned by the fixture tests, no decode step and no fatalError at startup.
- The compile-time manifest hash was already dropped, but "trust the fetch directly" went too far the other way: a fetched manifest is now validated structurally before it replaces the live map — 64-hex digests, safe board slugs and release tag, https base — and anything that fails is a no-op that keeps the current map. That's the same trust model as the firmware downloads: trust the API over HTTPS, refuse payloads that could produce an unsafe write path or an unverifiable image.
- Downloaded images are still verified against their per-image digest before anything is written to the drive, from whichever map is live.
- The refresh wiring (bundled seed, throttled fetch alongside the event firmware refresh, keep-on-failure) is unchanged — that part already matched the house pattern.

New store tests cover: a valid payload replaces the live map, the seed equals the audited bundled map, and garbage / bad digest / traversal slug / http base / traversal tag are all refused. Full suite passes: 2,961 tests in 517 suites.

PR #2345, opened 2026-08-21
fix(ble): pause scanning during pairing without dropping peripheral

### LINKED ISSUE #2212: 🐞 [Bug]:  iPad (multiple devices) fails to complete BLE pairing with nRF52-based nodes; iPhone connects fine with same firmware/app
### Before submitting

### Firmware Version

2.7.26

### What did you do?

Hardware: ThinkNode M6 (nRF52840) and LilyGo T1000-E (nRF52840)

Firmware Version: 2.7.26 (M6 originally 2.7.15, updated via UF2 through 2.7.25, now 2.7.26 — issue persists across all post-2.7.15 versions tested)

Meshtastic App Version: 2.7.16 (2109)

iPadOS Version: 26.5.2

iPad Models tested: iPad Pro 11-inch (M4) and iPad Pro 11-inch (3rd Generation)

Description:

Both nodes previously paired and connected via BLE without issue. After updating the M6’s firmware from 2.7.15 to 2.7.25 (and later 2.7.26), BLE pairing from iPad to both nodes stopped working. Symptom: node shows in scan list with full signal strength, tapping to connect shows “Connecting,” briefly flashes the system pairing/passkey confirmation screen, then reverts to “Connecting failed,” retries once more, then gives up.

### Expected Behavior

Connect iPads to Nodes 

### Current Behavior

Troubleshooting already completed, no effect on outcome:

	•	Forgot device in iOS Bluetooth settings + app, power-cycled node, re-paired fresh
	•	Reset Network Settings on iPad (full)
	•	Deleted and reinstalled Meshtastic app
	•	Verified Local Network permission granted to app
	•	Verified no Screen Time / Content & Privacy restrictions blocking Bluetooth
	•	Confirmed in-app Bluetooth toggle enabled
	•	Tested on a second, completely separate iPad Pro (3rd Gen, different unit) with a fresh app install — identical failure

### Participation

### Additional comments

 Key diagnostic finding: An iPhone (same Apple ID/network environment) connects to both nodes via BLE with zero issues, using the same app and same node firmware. The failure appears specific to iPad as a platform, not to app installation state, node bonding history, or firmware version — since two independent iPads both fail while iPhone succeeds against the same radios.

### Code of Conduct

bruschill:
## What changed?

- Pause CoreBluetooth scanning before connection and pairing without ending the discovery stream or clearing its peripheral cache.
- Keep deferred discovery and powered-on scan restarts paused for every connection attempt.
- Resolve a selected peripheral from the discovery cache or CoreBluetooth's identifier registry.
- Resume scanning after a failed connection when discovery still has a subscriber.
- Add transport-level regression tests for cache-miss resolution and the retry scan race.

## Why did it change?

BLE discovery remained active while CoreBluetooth subscribed to the encrypted Meshtastic characteristic. On the affected iPad, the PIN sheet appeared and disappeared before entry could complete.

PR #2183 stopped discovery before pairing, but that also cleared `BLETransport.discoveredPeripherals`. PR #2200 reverted it after normal reconnects failed with `Peripheral not found`. This change pauses the scan without dropping the selected peripheral.

Fixes #2212.

## How is this tested?

Automated:

- 15 focused BLE tests passed across 4 suites.
- The retry-race test was mutation-checked. It fails if the logical pause is made conditional on `isScanning` or deferred discovery ignores the pause.
- iOS simulator build passed.
- SourceKit-LSP reported no diagnostics in changed files.
- `git diff --check` passed.

Physical Release validation:

| Device | OS | Fresh pairing | Bonded reconnect |
| --- | --- | --- | --- |
| iPhone 16 Pro Max | iOS 27 beta (`24A5418b`) | Passed | Passed without PIN |
| iPad Pro 13-inch (M4), fresh install 1 | iPadOS 26.6.1 (`23G83`) | Passed | Passed without PIN |
| Same iPad, fresh install 2 | iPadOS 26.6.1 (`23G83`) | Passed | Not repeated |
| Same iPad, fresh install 3 | iPadOS 26.6.1 (`23G83`) | Passed | Not repeated |

Each iPad run used a new app container, no system bond, and no concurrent Meshtastic CoreBluetooth session. The PIN sheet stayed visible, accepted the fixed PIN, and reached `Connected Radio / Subscribed` in every fresh run.

The same iPad, radio, firmware, app sources, and PIN had failed three isolated runs before this change. Exact reporter parity on unsigned iPadOS 26.5.2 was not possible.

N/A. No UI changes.

## Checklist

### DIFF
--- a/Meshtastic/Accessory/Accessory Manager/AccessoryManager+Connect.swift
+++ b/Meshtastic/Accessory/Accessory Manager/AccessoryManager+Connect.swift
@@ -312,6 +312,12 @@ extension AccessoryManager {
 		do {
 			try await connectionStepper?.run()
 			Logger.transport.debug("🔗 [Connect] ConnectionStepper completed.")
+			// The scan pause covers the whole handshake — pairing happens during the
+			// notify subscription, after the link comes up — so resume only now that
+			// every step finished. Failed attempts resume via connectionDidDisconnect.
+			if let bleTransport = transportForType(.ble) as? BLETransport {
+				await bleTransport.resumeScanningAfterConnectionEstablished()
+			}
 		} catch AccessoryError.tooManyRetries {
 			self.lastConnectionError = AccessoryError.tooManyRetries
 			try await self.closeConnection()
--- a/Meshtastic/Accessory/Transports/Bluetooth Low Energy/BLETransport.swift
+++ b/Meshtastic/Accessory/Transports/Bluetooth Low Energy/BLETransport.swift
@@ -333,7 +333,19 @@ actor BLETransport: Transport {
 		centralManager.stopScan()
 	}
 
+	/// The pause exists to keep duplicate-advertisement traffic away from the pairing
+	/// window, which runs through the notify subscription — after the link itself comes
+	/// up. Once the handshake is fully established, resume so the Available Radios list
+	/// stays live for device switching while connected.
+	func resumeScanningAfterConnectionEstablished() {
+		resumeScanningIfPaused()
+	}
+
 	private func resumeScanningAfterFailedConnection() {
+		resumeScanningIfPaused()
+	}
+
+	private func resumeScanningIfPaused() {
 		guard scanningPausedForConnection else { return }
 		scanningPausedForConnection = false
 		guard discoveredDeviceContinuation != nil,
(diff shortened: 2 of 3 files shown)


garthvh:
Merged current main into the branch and added one commit: the scan now resumes when the connection stepper completes, not on disconnect only. Pairing happens during the notify subscription — after the link comes up — so the pause still covers the whole pairing-sensitive window; resuming after the handshake keeps the Available Radios list live for device switching while connected, which matches today's behavior. Added a test in your suite for the resume path. Full suite passes: 2,945 tests in 515 suites. The fix itself is good work — the cache-preserving pause plus registry fallback is what the #2183/#2200 history was missing.

PR #2346, opened 2026-08-21
fix(ble): stop stale saved-radio reconnect loops

bruschill:
## What changed?

- Treat confirmed CoreBluetooth pairing failures as terminal for the current connection process.
- Remove the failed UUID from the pairing hint and clear the preferred UUID only when it still matches.
- Preserve the preferred radio for ambiguous timeouts and plain disconnects.
- Stop another same-session automatic cycle after the existing retry budget without changing the persisted autoconnect setting.
- Add **Forget Saved Radio** in App Settings, including when the radio is unavailable, and disable it while connecting or connected.
- Preserve messages, nodes, SwiftData, and other app settings.
- Add localization, user documentation, and focused state/retry tests.

## Why did it change?

This separate failure was found while diagnosing #2212.

If the app retained a preferred BLE UUID and the user forgot the system bond while the app was closed, relaunch rediscovered that UUID and repeatedly retried before the user could clear the saved state. An unavailable preferred BLE radio also had no persistent forget action.

## How is this tested?

Automated:

- The full iOS simulator suite passed: 2,913 tests across 508 suites.
- `insufficientAuthentication` and `peerRemovedPairingInformation` stop after one attempt and clear only matching saved state.
- Encryption timeout, insufficient encryption/authorization, connection timeout, and plain disconnect keep saved state, use the existing retry budget, then block another same-session automatic cycle.
- A delayed teardown error could not reopen the closed gate.
- Valid saved-radio and later-session reconnect eligibility remained intact.
- iOS simulator build, localization validation, docs generation/drift check, JSON validation, and `git diff --check` passed.

Physical iPhone validation on iOS 27 beta (`24A5418b`):

1. Valid bonded baseline connected without PIN and reached `Connected Radio / Subscribed`.
2. The app was closed while retaining its preferred radio.
3. The radio was forgotten in iOS Settings.
4. Relaunch produced an immediate `Bluetooth Authentication is insufficient` error.
5. No PIN sheet, second attempt, or later automatic cycle appeared.
6. App Settings confirmed the matching saved UUID was gone.
7. Manual re-pair and automatic bonded reconnect passed.
8. With the radio unavailable, **Forget Saved Radio** remained visible.
9. Using it preserved messages, nodes, and other settings.
10. When the radio returned, it appeared in discovery but did not automatically connect.

One observation remains: after the classified error, the radio did not reappear until Bluetooth was disabled on a nearby iPad and the T1000-E was power-cycled. Those actions were concurrent, so the cause was not isolated.

N/A.

## Checklist


garthvh (changes requested):
Thanks for the deep investigation here — the physical repro (iOS reporting `insufficientAuthentication` for an externally removed bond) is genuinely useful knowledge, and the tests around the pairing-hint helpers are good. But stepping back to the product goal, I don't think the direction is right.

The goal is to keep a device paired for messaging. The app should fight to hold onto its saved radio, and main already has a self-heal designed for exactly the stale-bond case: `BLEConnection.isPairingFailure` recognizes `insufficientAuthentication` (and the rest of the stale-bond GATT errors), and on that error the handshake path calls `forgetPairedPeripheral` so the next attempt uses the long pairing window instead of the fast-reconnect timeout — which raises the iOS pairing dialog, the user re-pairs (screenless nodes use the fixed 123456 PIN, so this is always self-serviceable), and messaging resumes. A removed bond is treated as "re-pair and carry on," not "abandon the radio."

Measured against that, the three main changes point the wrong way:

- Clearing the saved preferred radio on a stale bond makes the app forget the radio at exactly the moment the right answer is the pairing dialog. Messaging is down until the user notices and re-selects.
- Stopping same-session automatic retries after one exhausted budget turns going out of range into a failure. It isn't one — a radio at the fringe or mid-reboot burns the two attempts, and then walking back into range does nothing for the rest of the session. Out of range is intended functionality: discovery sees the radio again and reconnects.
- The Forget Saved Radio button is an escape hatch for a trapped state the self-heal is designed to prevent.

If the repro showed the self-heal itself failing — the pairing dialog never appearing after the hint is forgotten, or the loop continuing after a successful re-pair — that evidence would justify a narrow fix, and I'd want to see it. The description shows the dialog re-appearing, which is the recovery working as designed.

Happy to take a much narrower version: the teardown-event guard and the pairing-hint tests stand on their own. The saved-radio invalidation, the retry gate, and the Forget button I'd drop.

garthvh:
Closing after the review discussion. The premise doesn't match how the BLE stack works here.

iOS owns bonding, not the app. When a bond is removed externally, the next protected GATT operation fails with insufficientAuthentication or peerRemovedPairingInformation, and re-initiating the connection is what makes iOS run pairing again — that dialog is the platform's designed recovery, not a failure loop. The app's existing handling already leans on it: isPairingFailure recognizes those errors and forgets the pairing hint so the next attempt uses the long pairing window and the dialog can appear. Every Meshtastic device can complete that re-pair — screen devices display the PIN, screenless devices use the fixed default — so there is no unrecoverable stale-bond state that justifies forgetting the saved radio or stopping retries.

The saved radio and retry-on-discovery behavior is the product's core loop: the goal is to keep the device paired for messaging. Range flapping, reboots after config saves, and bond churn are normal operating conditions the app is supposed to ride through.

The stale-bond repro knowledge is appreciated, and the teardown-event guard plus the pairing-hint tests would be welcome as a small standalone PR. If there's evidence of the existing self-heal actually failing — the pairing dialog never appearing after the hint is forgotten, or the loop continuing after a successful re-pair — that's worth a fresh issue with the log attached.

PR #2352, opened 2026-08-23
Gate MQTT proxy receives before decoding packets

RCGV1:
## Summary

- Acquire the existing MQTT-to-BLE forwarding gate before decoding a service envelope or building a `ToRadio` message.
- Reject MQTT proxy work while a BLE write is in flight.
- Keep the gate held through send completion and release it after successful, empty, or failed work.
- Add deterministic coverage for admission rejection and gate lifecycle behavior.

## Validation

- iOS Simulator: `xcodebuild test -workspace Meshtastic.xcworkspace -scheme Meshtastic -destination "platform=iOS Simulator,id=D8C8D4AA-2592-4FA5-81B2-EA6208020C93" -only-testing:MeshtasticTests/MqttForwardFilterTests -resultBundlePath /private/tmp/mqtt-admission-clean.xcresult`
- Result: **TEST SUCCEEDED** — 14 tests in 1 suite.
- `git diff --check upstream/main...HEAD` passed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MQTT-to-radio message forwarding to prevent duplicate or overlapping transmissions.
  * Dropped MQTT packets without usable payloads before forwarding.
  * Ensured forwarding resumes correctly after successful sends, rejected messages, and transmission errors.

* **Tests**
  * Added coverage for busy, successful, empty, and failed forwarding scenarios.

### DIFF
--- a/Meshtastic/Accessory/Accessory Manager/AccessoryManager+MQTT.swift
+++ b/Meshtastic/Accessory/Accessory Manager/AccessoryManager+MQTT.swift
@@ -33,6 +33,45 @@ actor MqttForwardGate {
 	}
 }
 
+enum MqttForwardAdmissionResult<Work> {
+	case rejected
+	case admitted(Work)
+}
+
+enum MqttForwardAdmission {
+	@MainActor
+	static func admit<Work>(
+		gate: MqttForwardGate,
+		build: () -> Work
+	) async -> MqttForwardAdmissionResult<Work> {
+		guard await gate.tryAcquire() else {
+			return .rejected
+		}
+		return .admitted(build())
+	}
+
+	static func complete<Work>(
+		_ admission: MqttForwardAdmissionResult<Work?>,
+		gate: MqttForwardGate,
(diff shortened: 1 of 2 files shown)


garthvh:
Closing this one. The gate's drop-don't-queue behavior is by design — the public broker outruns BLE, and dropping excess is the intended outcome, so there's no bug here to fix. Moving the decode inside the gate makes things slightly worse, not better: the gate is then held during decoding, which lengthens the busy window and drops more packets, and the @MainActor hop adds main-thread work per packet at public-broker rates. The admission machinery and most of the tests exist to manage a release-on-failed-decode problem that only appears once the decode moves inside the gate.

The two small genuine improvements — releasing the gate synchronously instead of through a deferred task, and logging send failures instead of swallowing them — are too small to carry a PR; the first closes a microsecond window whose only effect is a packet dropped that the gate exists to drop anyway. If either matters later it can ride along in a real change to this file.

PR #2354, opened 2026-08-24
Take the Watch node snapshot on the MeshPackets actor

garthvh:
## Summary

The most frequent crash in the app is a SIGTRAP in `WatchNode.make` under `WatchSessionManager.fetchNodesForWatch`, with SwiftData's `_SD_get_faulting_backingdata_tsd` on the stack — SwiftData asserting because a model's backing row no longer exists in the store. A symbolicated TestFlight report from 2.7.19 (2306) pinned it.

The Watch update walks every node with a user on the main context. Meanwhile the `MeshPackets` actor — its own executor, its own `ModelContext` — deletes rows: node cap eviction and near-duplicate position pruning. A main-context instance whose row was deleted by that other context keeps passing the `modelContext != nil` / `isDeleted == false` liveness guards until the merge is processed; the next persisted-property read traps fatally. The guards added in 2.7.17 narrowed the window, which is why the crash shrank but never disappeared. They cannot close it: those checks only see deletions made through the entity's own context.

## What changed

The snapshot now runs on the `MeshPackets` actor itself (`watchNodeSnapshot`), so it is serialized against every delete that actor performs — the race is removed rather than narrowed. Only `Sendable` `WatchNode` value structs cross back to the main actor, which then encodes and sends them to the Watch as before.

- `WatchSessionManager.performSendNodesToWatch()` is async; the user location is read on the main actor and passed in as plain coordinates. `fetchNodesForWatch()` is gone.
- `WatchNode.make` and its liveness guards stay, as a backstop against main-context writers (admin node removal, database clear) — the comment now explains what they can and cannot catch.
- A retired actor instance (after `recreateShared()`) returns an empty snapshot instead of reading through a possibly torn-down container.

## Testing

- New `WatchNodeSnapshotTests`: mapping and range filtering (in range, out of range, no position, no user), the 0/0 no-fix sentinel, the retired-instance refusal, and a snapshot taken after actor-side cap eviction seeing only survivors.
- Full unit suite passes; iOS builds. The crash itself is a cross-actor timing race, so the serialization property is by construction rather than something a unit test can reproduce — worth watching the crash reporter after the next TestFlight build to confirm the family goes quiet.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Apple Watch node updates to avoid crashes while nodes are removed or cleaned up.
  * Watch data now consistently reflects only active, valid nodes.
  * Excludes nodes without location fixes and filters results by distance.

* **Tests**
  * Added coverage for node filtering, invalidated data sources, location validation, and node eviction scenarios.

### DIFF
--- a/Meshtastic/Helpers/MeshPackets.swift
+++ b/Meshtastic/Helpers/MeshPackets.swift
@@ -5,6 +5,7 @@
 //  Created by Garth Vander Houwen on 5/27/22.
 //
 
+import CoreLocation
 import Foundation
 @preconcurrency import SwiftData
 import MeshtasticProtobufs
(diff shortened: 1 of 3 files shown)


garthvh:
Both review findings addressed: the snapshot fetch now logs failures instead of discarding them with try?, and the eviction test seeds distinct lastHeard timestamps and asserts which nodes survive (the count-only assertion was also nondeterministic — good catch).

PR #2355, opened 2026-08-24
Never hand the message lists duplicate ForEach ids

garthvh:
## Summary

The most frequent 2.7.19 TestFlight crash after the Watch one is a SIGABRT in a SwiftUI List batch update: `NSInternalInconsistencyException — attempt to delete item 1 from section 0 which only contains 1 items before the update`, thrown inside `UICollectionView` under SwiftUI's list coordinator, with no app frames. That exception is what UIKit throws when a List's diff disagrees with the data it was handed — classically, duplicate `ForEach` ids.

Both message lists key their `ForEach` on `messageId`. `messageId` is `@Attribute(.unique)`, but uniqueness is enforced per save, and there is a real window where the store view contains two rows with the same id:

1. `sendMessage` inserts the outgoing row on the main context, then starts the radio transmit in an unstructured `Task` **before** `context.save()` runs. The save can also queue behind the ingest actor's writes.
2. The mesh echoes the packet back within seconds. The ingest actor's duplicate guard fetches by `messageId` — on its own context, which cannot see main's uncommitted row — so it inserts a second row with the same id.
3. The unique constraint merges the rows at save, but a list fetch landing between the commits hands the `ForEach` both. The next batch update throws.

## What changed

- `sendMessage` saves the row before creating the transmit task, so the echo cannot come back before the row is committed. Side benefit: a failed save no longer transmits a packet with no local record.
- Both `UserMessageList` and `ChannelMessageList` deduplicate by `messageId` at load (`MessageEntity.deduplicatedByMessageId`, first occurrence wins), so the `ForEach` can never see a duplicate id regardless of how one reaches the store — including packet-id collisions between different senders, which the per-sender mesh packet id space allows.

## Testing

- New `MessageDeduplicationTests` pin the helper: first-occurrence-wins with order preserved, unique lists pass through unchanged, empty input.
- Full unit suite passes; iOS builds. The crash itself needs a sub-second cross-context race on a live radio, so it is not directly reproducible in a unit test — the dedupe layer is the guarantee, the save reorder removes the common trigger. Worth confirming the family goes quiet after the next TestFlight build, alongside the Watch fix in #2354.

## Summary by CodeRabbit

* **New Features**
  * Outgoing messages are saved locally before transmission, improving message continuity.
* **Bug Fixes**
  * Prevented duplicate messages from appearing in channel and direct-message lists.
  * Improved list stability when sent messages and mesh echoes briefly overlap.
* **Tests**
  * Added coverage for duplicate removal, message ordering, unique messages, and empty lists.

### DIFF
--- a/Meshtastic/Accessory/Accessory Manager/AccessoryManager+ToRadio.swift
+++ b/Meshtastic/Accessory/Accessory Manager/AccessoryManager+ToRadio.swift
@@ -460,14 +460,20 @@ extension AccessoryManager {
 					var toRadio: ToRadio!
 					toRadio = ToRadio()
 					toRadio.packet = meshPacket
-					Task {
-						let logString = String.localizedStringWithFormat("Sent message %@ from %@ to %@".localized, String(newMessage.messageId), fromUserNum.toHex(), toUserNum.toHex())
-						try await send(toRadio, debugDescription: logString)
-						Logger.mesh.info("💬 \(logString, privacy: .public)")
-					}
 					do {
+						// Save BEFORE the transmit task can start: the mesh echoes this packet
+						// back within seconds, and the ingest actor's duplicate guard fetches
+						// the store — it cannot see this context's unsaved row. A transmit that
+						// beats the save re-inserts the echo under the same messageId, and the
+						// message list briefly renders duplicate ForEach ids, which corrupts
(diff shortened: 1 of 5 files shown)


garthvh:
Addressed: the outer catch now rethrows after logging, so a failed save reaches the caller instead of reporting silent success. The function was already declared throws and every caller uses try await.

PR #2357, opened 2026-08-26
Track GeoJSON provenance: planner runs replace, uploads stay user-controlled

garthvh:
## Summary

Map data files did not record where they came from, and per-file visibility did not stick. The toggles in map settings wrote to a transient set that was rebuilt from stored state on the next launch, so hiding a file never survived a relaunch. Successive Site Planner coverage runs piled up on the map instead of replacing each other.

## What changed

- `MapDataMetadata` records a `source`: `manualUpload` or `sitePlanner`. Manifests written before the field decode as manual uploads.
- A Site Planner run deactivates older planner files on import, so only the newest run draws. The map resyncs its enabled overlays from the store after an import, which removes the previous run's overlay in the same render. Manual uploads are never touched by this rule.
- Manually uploaded files arrive visible, and the per-file toggles in map settings now persist through the store — visibility survives relaunch.
- Both file lists (map settings and the map data screen) show a Site Planner badge on coverage files.
- The overlay-file row in MapSettingsForm is extracted into helpers; the added pill pushed the old single expression past the compiler's type-check budget.

## Testing

- New tests: manifests without the source field decode as manual uploads; a second planner run deactivates the first while leaving a manual upload alone; manual uploads arrive active and user-set visibility survives later uploads.
- Full unit suite passes (2,941 tests in 514 suites); iOS builds.
- Worth a hands-on pass: run two coverage estimates back to back (second should replace the first on the map), upload a GeoJSON manually, hide it in map settings, relaunch — it should stay hidden.

## Summary by CodeRabbit

* **New Features**
  * Added support for identifying map overlays from manual uploads and Site Planner imports.
  * New Site Planner imports automatically deactivate older Site Planner overlays.
  * Added controls to activate or deactivate individual overlay files.
  * Site Planner overlays now display a source badge.

* **Bug Fixes**
  * Manual upload visibility is preserved across Site Planner imports.
  * Legacy map metadata remains compatible and defaults to manual uploads.
  * Map displays now rebuild overlays using only active files.

### DIFF
--- a/Meshtastic/Helpers/GeoJSONOverlayManager.swift
+++ b/Meshtastic/Helpers/GeoJSONOverlayManager.swift
@@ -149,4 +149,10 @@ class GeoJSONOverlayManager {
 		// Clear cache to force reload with new file states
 		clearCache()
 	}
+
+	/// Set the active state of an uploaded file explicitly (persisted).
+	func setFileActive(_ fileId: UUID, _ isActive: Bool) {
+		MapDataManager.shared.setFileActive(fileId, isActive)
+		clearCache()
+	}
 }
--- a/Meshtastic/Helpers/MapDataManager.swift
+++ b/Meshtastic/Helpers/MapDataManager.swift
@@ -57,7 +57,7 @@ class MapDataManager: ObservableObject {
 	// MARK: - File Upload & Processing
 
 	/// Process and store an uploaded file
-	func processUploadedFile(from sourceURL: URL) async throws -> MapDataMetadata {
+	func processUploadedFile(from sourceURL: URL, source: MapDataSource = .manualUpload) async throws -> MapDataMetadata {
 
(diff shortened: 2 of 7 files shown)


garthvh:
Addressed four of the five findings:

- setFileActive reverts the in-memory state when the manifest save fails, so the session never shows a visibility that a relaunch would flip back. Kept it non-throwing — the caller is a toggle Binding setter; the revert is what restores consistency.
- A manual upload from map settings now inserts its id into the render set, so the row and overlay show immediately instead of waiting for a toggle.
- The policy tests await cleanup at the end of the body instead of a defer-launched Task, and the visibility test asserts through a fresh manager + loadMetadata() so it fails if persistence stops working.

Skipped the badge-metric finding: the literals match the pre-existing format-pill styling in the same rows, and this codebase has no shared metric constants to point at — inventing a token system is out of scope for this change.

PR #2360, opened 2026-08-26
Keep statuses from unknown nodes; pin the status receive chain

garthvh:
## Summary

A report came in that status messages stopped working, with the merged channel changes as the suspect. The channel merge contains no status-related lines, and the receive chain passes end-to-end tests on current main. The tests did surface one real gap: a status broadcast from a node the app has no row for was dropped, so after a node DB clear statuses stayed missing until the next broadcast — which can be a long wait.

## What changed

- `upsertNodeStatusPacket` creates a stub node when none exists, the way the position and telemetry handlers already do. It fills in when the NodeInfo packet arrives, and it is stamped lastHeard so cap eviction ranks it correctly.
- New `NodeStatusIngestTests` pin the chain: a broadcast updates the node's status and the value the node list renders, an empty broadcast clears it, a status from an unknown node mints a stub, and the module config packet creates the entity the editor form requires.

## Testing

Full unit suite passes (2,942 tests in 514 suites).


garthvh:
Snipe hunt

PR #2383, opened 2026-08-31
Make legacy data migration responsive and replay-safe

bruschill:
## What changed?

This moves the one-time legacy Core Data import behind a process-wide persistence bootstrap. The app now presents a responsive, persistence-free migration screen before it creates normal SwiftData-backed UI or starts database-dependent services.

### Startup and publication

- Add `PersistenceBootstrap` as the single owner of persistence startup.
- Check protected-data availability before touching any store.
- Move the legacy store and its sidecars to an isolated legacy path before creating `ModelContainer`.
- Run the legacy copy off the main actor.
- Publish the normal SwiftUI tree, `MeshtasticAPI`, discovery, map, Watch, TAK, and performance-seed work only after the bootstrap reaches `.ready`.
- Make BLE restoration, CarPlay, SiriKit, App Intents, notification actions, and memory-warning saves wait for the same process-wide readiness gate.
- Show a failure state with Retry without mounting a provisional SwiftData view tree.
- Disable Auto-Lock while migration is active, then restore the value the app inherited.

### Replay-safe migration

- Copy large tables in bounded batches using forward-only Core Data object-ID snapshots.
- Split message migration into scalar insertion and relationship repair so each phase can save in bounded chunks.
- Preserve every legacy V58 message field and sender/recipient relationship.
- Preserve existing non-nil SwiftData message relationships during Retry.
- Reconcile keyless position and telemetry history by full-field fingerprint and multiplicity. This preserves duplicate history rows instead of collapsing them.
- Use fresh destination contexts for bounded scans so SwiftData does not retain the full destination graph in memory.
- Add deterministic checkpoints around parent saves, history batches, message phases, store preparation, and source retirement.

### Store safety

- Prepare and retire the legacy main store and sidecars as explicit, replayable phases.
- Use preparation and retirement markers to distinguish an interrupted move owned by this migration from an unrelated filesystem collision.
- Fail closed when protected data is unavailable or a destination/backup collision cannot be proven safe.
- Close the Core Data source before retirement.
- Retire the source only after the full migration succeeds. The legacy main store becomes `Meshtastic-coredata-backup.sqlite`; no raw SQLite writes, schema changes, or destructive fallback are used.

### Migration UI and coverage

- Add light, dark, and large-text migration/failure states using appearance-specific thumbnails for the selected app icon.
- Add VoiceOver headings, state announcements, identifiers, Retry semantics, and a hidden decorative progress indicator so the status is not spoken twice.
- Add deterministic Debug-only launch previews for the migrating and failed states.
- Add an iOS UI-test target that runs Apple's accessibility audit against both states.
- Add five opt-in pixel-comparison snapshots. Existing snapshots remain dimension-only.
- Add current developer documentation for persistence bootstrap ownership, the legacy import boundary, and cold-launch gating. Regenerate the bundled Markdown, HTML, and search-index entries.
- Remove the old constructor-owned migration path and unused test-only flexibility so production migration has one owner.

## Why did it change?

Older installs can contain a large Core Data history that must be imported into the current SwiftData store. Performing that work before the first usable frame makes an update look frozen even when the process is still copying data.

Moving the copy off the main actor is only part of the fix. The destination is saved in batches, and SwiftData does not provide a migration-wide snapshot that hides those intermediate saves. Mounting normal UI or starting a writer early could expose incomplete parents, messages, relationships, positions, or telemetry. The bootstrap therefore treats `.ready` as the publication point for the whole persistence-dependent app, including alternate cold-launch paths such as BLE restoration, CarPlay, Siri, notifications, and App Intents.

Retries also cannot assume an empty destination. A process may stop after any committed batch or during a multi-file store move. Deterministic IDs cover keyed records, while full-field fingerprint multiplicity covers keyless history. Preparation and retirement markers make interrupted filesystem transitions distinguishable from unrelated collisions. This allows Retry to continue without duplicating data or deleting a store it does not own.

The migration intentionally uses supported Core Data and SwiftData APIs and preserves the audited message, position, and telemetry fields, including message sender and recipient relationships. It does not publish a partial store, omit audited history, write SQLite directly, change the schema, or fall back to deleting unreadable data.

### Focused automated tests

The final focused run passed 41 tests on the iOS simulator:

- 2 Core Data migration tests: one bundled-model guard and one populated-destination integration test.
- 9 disk-backed interruption and Retry tests.
- 10 persistence-bootstrap tests, including concurrent waiters, protected-data recovery, recovery failure, Retry, and idle-timer restoration.
- 9 snapshot-reference tests.
- 4 migration-view behavior tests.
- 5 migration-view pixel snapshots.
- 2 Apple accessibility-audit UI tests.

The interruption matrix covers:

- parent-save interruption;
- every position and telemetry history batch;
- message scalar save and sender/recipient repair;
- committed data before retirement;
- every preparation and retirement store move;
- unrelated retirement collisions;
- protected-data wait and recovery;
- concurrent cold-launch waiters.

The snapshot guard was also mutation-tested by replacing the failed-dark reference with a same-size light image. Verification failed with `Snapshot content changed`, then passed after the exact reference was restored.

### Large mixed-store qualification

The final source was installed on an iOS 26.5 simulator and run three consecutive times against the same reset fixture. The fixture contains 100,000 legacy messages, 100,000 legacy positions, and 100,000 legacy telemetry rows plus an already populated SwiftData destination.

| Trial | First migration UI | Peak RSS | In-process migration | Launch to retirement |
|---|---:|---:|---:|---:|
| 7 | 1.548 s | 665.3 MB | 110.513 s | 111.934 s |
| 8 | 1.839 s | 705.8 MB | 107.804 s | 109.443 s |
| 9 | 1.340 s | 665.2 MB | 106.797 s | 108.011 s |

Every final trial produced:

| Check | Result |
|---|---:|
| Final messages | 152,498 |
| Final positions | 125,000 |
| Final telemetry rows | 175,000 |
| Legacy message scalar or relationship deficits | 0 |
| Position fingerprint multiplicity deficits | 0 |
| Telemetry fingerprint multiplicity deficits | 0 |
| Legacy main store after completion | absent |
| Backup main store after completion | present |

The preservation comparison covers 25 common message scalar fields plus sender/recipient relationships, 12 position fields, and 43 telemetry fields.

<details>
<summary>Qualification anomaly retained for review</summary>

Before the final clean set, one pre-reboot simulator run did not complete before the 300-second harness timeout; message scalar insertion was the last incomplete phase. The next run completed with a 247.248-second in-process duration. Those logs were retained rather than discarded. The cause remains unknown, and the successful post-reboot runs do not establish that it was simulator-only. After rebooting the simulator without erasing it, trials 4-6 completed consecutively near the established baseline. Trials 7-9 above then repeated the qualification on the exact final source.

This PR does not claim a fixed completion time. The release invariants are a responsive first frame, bounded memory, exact preservation, replay safety, and safe source retirement. The large-store results are simulator measurements, not physical-device jetsam or watchdog proof.

</details>

### Build and project checks

- `xcodebuild` for the final iOS simulator app: passed.
- Focused unit, snapshot, and UI-audit run: passed.
- `bash scripts/build-docs.sh --output Meshtastic/Resources/docs`: passed with an 8.1 MB bundle, below the 10 MB limit.
- `xcodegen generate`: stable on a second generation.
- `git diff --check`: passed.
- Deliberate same-size snapshot mutation: failed as expected; restored baseline passed.

Mac Catalyst test execution was not run locally because this machine has no matching Mac Development certificate or provisioning profile. Physical pre-first-unlock BLE restoration, CarPlay, Siri, and notification lifecycle behavior remains device-level validation; those paths compile, share the tested bootstrap gate, and passed static review.

meshtastic-migration-loading-screen-dark
meshtastic-migration-loading-screen-light

Checked-in visual references also cover:

- migrating, light;
- migrating, dark;
- migrating at Accessibility 3 text size on a narrow screen;
- failed, light;
- failed, dark.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a startup screen showing migration progress, failures, and retry options.
  * Added VoiceOver announcements for migration status changes.
  * Legacy data migration now supports resumable progress while preserving relationships and message history.

* **Bug Fixes**
  * Improved reliability when launching through Siri, notifications, CarPlay, or Bluetooth by waiting for local data availability.
  * Actions now provide a clear error when local data cannot be opened.

* **Tests**
  * Expanded coverage for migration recovery, accessibility, UI states, and snapshot accuracy.


garthvh:
Thanks for this — the problem is real. The legacy import runs synchronously inside `PersistenceController.shared`'s init today, so a large Core Data store does block the first frame, and the qualification work here is thorough.

The scope is the blocker. This is 2,598 lines across 49 files to improve a one-time migration from 2.7.12, which is now nine releases back and a shrinking population. I don't want to touch 49 files for that.

Most of the fan-out looks avoidable. 22 of the changed files are production Swift, and 11 of those are App Intents each getting the same one-line readiness gate, plus CarPlay, the Siri handlers, AccessoryManager and BLETransport. The actual persistence change is about four files. If the container accessor itself waited or threw until the bootstrap is ready, none of those entry points would need editing and the diff would be a quarter of the size, with one place to reason about instead of a dozen.

Three other things I'd want settled first:

The unexplained hang. One qualification run passed the 300-second timeout during message scalar insertion, cause unknown and not established as simulator-only. This code gates app launch, so on a device that is a watchdog kill rather than a slow migration. That needs an explanation before anything in this area merges.

The untested paths. The PR notes Catalyst wasn't run and that device BLE restoration, CarPlay, Siri and notification lifecycle were statically reviewed only. Those are exactly the paths the new gate sits in front of, and they're the ones we can't easily catch in review.

The test target. Adding a UI test target and five pixel snapshots brings a CI decision with it that's still open in #2180, and pixel comparisons are a standing flake risk. I'd rather that arrived separately on its own merits.

Suggested split: the bootstrap gate and the off-main-actor copy as one focused PR, with the gate owned by the container accessor; the replay-safety and batching work as a second; the UI test target and snapshots as a third. Happy to look at the first one quickly if you'd like to take that route.

bruschill:
Superseded by #2391 and two more branches that split this work into 3 parts.

PR #2390, opened 2026-09-01
Throttle the launch-time catalog and firmware-list refreshes

garthvh:
## Summary

Railway traced most of api.meshtastic.org's egress to `/resource/deviceHardware` — 1.16 million requests in 24 hours from tens of thousands of clients, every one a full 200 with no caching. We were contributing: the app pulled the full device catalog and the firmware release list on every process start, unthrottled.

Neither endpoint sends `Cache-Control` or an `ETag`, so no client or CDN can cache or revalidate them. Until that changes on the API side, the only lever we have is asking less often, for data that changes a few times a week.

## What changed

Both launch-time refreshes are skipped when the last successful fetch is under 72 hours old. Background wakes count as process starts, so the interval is deliberately long. The device catalog and the firmware screen's own refresh share a single timestamp, stamped where the fetch actually succeeds, so they don't each pull the catalog. The screen keeps a shorter six-hour interval: the user opened it deliberately, and it's the one place a board newer than the build has to resolve by name.

When the catalog refresh is skipped, the bundle-only image and link pass still runs — that one is local and is what puts device images on screen offline.

Nothing user-initiated is throttled: pull to refresh, a user-initiated connect, and the update notifier's existing 24-hour staleness check all behave as before.

## Testing

Built for device. The throttle is a timestamp comparison with the first launch treated as stale, so a fresh install still fetches immediately. Payload sizes for reference: the catalog is 38,536 bytes raw and 4,059 gzipped; the firmware list is 157,357 raw and about 36,000 gzipped.

### DIFF
--- a/Meshtastic/API/MeshtasticAPI.swift
+++ b/Meshtastic/API/MeshtasticAPI.swift
@@ -184,12 +184,17 @@ class MeshtasticAPI: ObservableObject, @unchecked Sendable {
 	static let firmwareGitHubURLEndpoint = URL(string: "https://api.github.com/repos/meshtastic/firmware/releases?per_page=100")!
 	static let nightlyIndexEndpoint = URL(string: "https://raw.githubusercontent.com/meshtastic/meshtastic.github.io/master/firmware-nightly/index.json")!
 
-	/// How long a launch-time API refresh stays fresh. Every cold launch used to pull the full
+	/// How long a launch-time API refresh stays fresh. Every process start used to pull the full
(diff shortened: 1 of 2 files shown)


garthvh:
Superseded by #2393. apiv2 sends Cache-Control and an ETag and returns a 304 with an empty body, so URLSession serves a repeat call from its cache or revalidates it for nothing. A hand-rolled timestamp throttle on top of that would only make the data staler than it needs to be, with no traffic saved.

PR #2391, opened 2026-09-01
Make legacy data migration responsive

bruschill:
## Summary

Moves the one-time legacy Core Data copy off the main actor. The app shows a simple progress view until the copy finishes, then mounts the SwiftData view tree and starts app services.

This is the first split from #2383. Replay batching, checkpoint recovery, retry UI, cold-launch adapter changes, UI tests, and snapshots are not included.

## What changed

- Run the legacy copy on a dedicated actor with its own `ModelContext`.
- Delay the SwiftData view tree and app-service startup until the copy finishes.
- Keep the existing container API and migration failure behavior.
- Test that the legacy copy runs off the main thread and still preserves existing data.

## Testing

- `xcodebuild -project Meshtastic.xcodeproj -scheme Meshtastic -destination 'platform=iOS Simulator,id=D4C8718E-8D3A-4FC5-9132-1DF3BE7D3372' -derivedDataPath /tmp/Meshtastic-PR2391-Narrow-DD build`
- `xcodebuild -project Meshtastic.xcodeproj -scheme Meshtastic -destination 'platform=iOS Simulator,id=D4C8718E-8D3A-4FC5-9132-1DF3BE7D3372' -derivedDataPath /tmp/Meshtastic-PR2391-Narrow-DD test -only-testing:MeshtasticTests/CoreDataMigrationServiceTests`
- `git diff --check $(git merge-base HEAD upstream/main)`


bruschill:
It didn't make the notes, but that long initial runtime from the last PR was because my laptop went to sleep. I was unable to reproduce when the computer stayed awake.

ancientastronauttheorist:
I can reproduce the launch failure on a physical iPhone 15 Pro Max.

Environment

Meshtastic 2.7.21 (build 2341)

iOS 26.6.1 (23G83)

Device model identifier: iPhone16,2

User-observed regression began around 2.7.19

Symptom: black screen for about 20 seconds, then the app exits; every relaunch repeats it

Crash-log confirmation

The attached .ips confirms this is a launch watchdog termination, not a normal Swift exception:

FRONTBOARD 0x8BADF00D
WatchdogEvent: process-launch
exhausted real (wall clock) time allowance of 20.00 seconds
Elapsed application CPU time: 19.934 seconds
Thermal State: nominal

procLaunch was 11:03:58.0231 and the watchdog capture was
11:04:18.0587, a difference of 20.0356 seconds.

At the watchdog sample, Thread 0 / com.apple.main-thread is:

libswiftCore: Sequence.contains(where:)
SwiftData: 18 consecutive frames
Meshtastic: +0x25f3f8
Meshtastic: +0x2625a4
Meshtastic: +0x283874
Meshtastic: +0x2817d8
Meshtastic: +0x2816ac
libdispatch: _dispatch_once_callout
Meshtastic: +0xaca9e4
SwiftUI: static App.main()

That is consistent with SwiftData work occurring beneath a one-time static
initializer before normal app startup completes. The report also says:

wasUnlockedSinceBoot: 1
isLocked: 0

so this instance was not the pre-first-unlock protected-data case.

The Meshtastic app frames are not symbolicated in the user-generated report.
The matching app-slice UUID is:

90F6D0FF-AF49-3481-8DA7-EA3F3C0D1592

Could a maintainer run those offsets through the 2.7.21 build-2341 dSYM to
confirm the precise app functions? With the load address recorded in the report,
the command should be approximately:

dwarfdump --uuid Meshtastic.app.dSYM
atos -arch arm64 \
  -o Meshtastic.app.dSYM/Contents/Resources/DWARF/Meshtastic \
  -l 0x100470000 \
  0x1006cf3f8 0x1006d25a4 0x1006f3874 0x1006f17d8 \
  0x1006f16ac 0x100f3a9e4 0x100f40534

This physical-device report strongly supports the startup/migration problem
that #2391 is addressing: render a usable frame first and run the legacy copy
off the main actor. I can validate a TestFlight build of this PR against the
affected installation and its existing app data. I am preserving the current
installation and have not deleted/reinstalled it.

garthvh (commented):
The core of this is right and worth landing. Splitting it out of #2383 was the correct call, and the BLE part is not the scope creep it looks like at first glance: deferring the copy means CoreBluetooth can cold-launch the app from a background event and drive restoration before the store is migrated, so awaiting the bootstrap there is a necessary consequence, not stray work.

Two changes carry the weight and both look correct. Moving off `swiftDataContainer.mainContext` to a fresh `ModelContext` is what actually makes this possible — `mainContext` cannot be touched from a non-main actor — and `@LegacyMigrationActor` gives it a serial executor. Good.

Some things to trim, since the aim was a narrow PR.

**`protectedStoreIsUnavailable()` is never called.** It is defined at `CoreDataMigrationService.swift:71` and referenced nowhere else in the diff. In #2383 the bootstrap consulted it before touching either store; here it arrived without its caller. Either wire it up — a pre-first-unlock launch really can see the files and fail to read them, and BLE restoration makes that reachable — or drop it and bring it back with the code that uses it.

**`migrateOffMain` is a pass-through.** Its whole body is `try await migrate(into:executionProbe:)`, and `migrate` is already `@LegacyMigrationActor`, so any caller is already hopping off the main actor. That is two public entry points for one behaviour. Calling `migrate` directly would read better and delete a name.

**The execution probe tests a compile-time guarantee.** `MigrationExecutionProbe`, the `executionProbe` parameter on both entry points, and `migrationExecutorIsMainThread()` exist so a test can assert the copy did not run on the main thread. But `@LegacyMigrationActor` is a custom global actor with the default cooperative executor — it cannot run on the main thread, and the compiler enforces the isolation. So three symbols are threaded through the production signature to assert something Swift already guarantees. I would drop the probe and let the annotation be the contract.

**One `save()` for the whole migration.** With `autosaveEnabled = false` and a single `sdContext.save()` at line 145, the entire legacy copy is one unbounded transaction. That is a reasonable scope boundary for this PR — batching was explicitly left to a follow-up — but it means the peak-memory behaviour is unchanged, and it is the largest stores that both take longest and are most likely to be interrupted. Worth saying plainly in the description so it is not mistaken for fixed.

**`handleWillRestoreState(dict:central:)` widened to `CBCentralManager?`.** Production never passes nil; only the test does. A loosened production signature to make something injectable is worth a second look.

**Not this PR, but it changes how you would prioritise the follow-ups.** The failure path still logs to OSLog and swallows the error, and the comment is explicit that behaviour is preserved. Nothing reports a failed migration anywhere off-device, so the silent outcome is that a user loses their message history and no one learns it happened. I went looking for evidence of how often this fails before forming an opinion on the batching and checkpoint work, and there is none to find — not because it does not happen, but because it is not instrumented. One error report in that `catch` would turn the whole replay-safety question from a judgement call into a number.

Happy to re-run the drift check when you want; the one red check is the duplicate `Shared` group flake, and #2501 fixes it at the root.

Issue #2057, opened 2026-07-07
🐞 [Bug]: not able to enter bluetooth pin (dialogue automatically closes)


sronline:
### Before submitting

- [x] I have searched existing issues to make sure this bug hasn't already been reported
- [x] I have updated to the latest version of the software to verify the issue still exists

### Firmware Version

2.7.15

### What did you do?

IOS App 2.7.15 (2109)

Want to pair Meshtastic devices with custom pin, but the bluetooth pin dialogue opens but automatically closes, so I'm not able to enter the pin

### Expected Behavior

able to enter pin

### Current Behavior

bluetooth pin dialogue opens but automatically closes, so I'm not able to enter the pin

### Participation

- [ ] I am willing to submit a pull request for this issue.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #2401, opened 2026-09-03
Revert the BLE pairing PIN handshake (#2057)

garthvh:
## Summary

Reverts #2057 (`786355d2`) and its follow-up `76310a28`.

#2057 introduced a connect path that resumed the suspended connect with a raw
notify-subscription error instead of sending it through
`BLEConnection.handlePeripheralError`, the classifier that decides whether an
error is worth reconnecting for. Connect steps retry by default, so a radio
that had erased its pairing was retried until the budget ran out and reported
"Too Many Retries" instead of the message telling the user to forget the radio
under Settings > Bluetooth.

## What changed

Everything #2057 and its follow-up added is gone:

- Step 1 goes back to a flat 5 second timeout instead of 90 seconds for a
  first-time BLE bond.
- Connect resolves once the characteristics are discovered again, rather than
  holding the step open until the `FROMNUM` subscription confirms.
- The notify-state error handler and `isPairingFailure` classification are
  gone. Pairing failures reach the app only as peripheral disconnects, which
  the existing classifier already treats as final — so they stop the connect
  and report themselves.
- `UserDefaults.pairedPeripheralIds` and its one-time migration go away with
  the timeout that used them, along with their tests and the transport doc
  section.

## What comes back with it

The bug #2057 fixed. On a first connection to an encrypted radio, iOS presents
the pairing PIN sheet during characteristic subscription; the 5 second Step 1
timeout can call `cancelPeripheralConnection` while the sheet is still up,
dismissing it before a 6-digit PIN can be read and entered.

## Alternative

#2400 fixes the retry problem without giving that up: it routes the notify
failure through the same classifier, which is what #2057 should have done. The
two are mutually exclusive — #2400 changes code this PR deletes.

## Testing

Builds for iOS. Full suite passes, 2990 tests (down from 3004: the paired-hint
and pairing-classifier tests are part of what is reverted).

Not tested on hardware. Worth checking both cases on a radio with a PIN: a
first-ever pair, and reconnecting after a factory erase.

## Summary by CodeRabbit

- **Bug Fixes**
  - BLE connections now proceed without waiting for notification-state confirmation during pairing.
  - Connection setup uses a consistent 5-second timeout across transport types, including first-time Bluetooth connections.

- **Behavior Changes**
  - Previously saved Bluetooth pairing hints and related migration behavior are no longer retained.
  - Pairing-failure confirmation handling has been removed from the connection flow.


garthvh:
Closing in favor of #2400.

This fixed the retry bug by giving up the fix that caused it. #2400 fixes the
same bug by routing the notify failure through `handlePeripheralError`, the
classifier that already treated pairing errors as final, and keeps the
pairing-sheet behavior.

The branch stays if anyone wants it later.

PR #2411, opened 2026-09-05
Add remote channels with the shared editor for Android parity

RCGV1:
## What changed?

- Add a native Remote Channels screen for PKI-administered nodes, with sequential reads for all eight channel slots and progress, timeout, retry, and cancellation handling.
- Reuse the regular Settings channel list row (`ChannelRow`) and unchanged channel editor (`ChannelForm`) for remote channels, including name, key, role, MQTT flags, and position precision. Preserve other channel fields when saving enabled channels; disabling a slot clears its settings.
- Validate channel roles, UTF-8 name length, and key length before sending changes, then read each changed slot back and compare the editable fields before reporting success.
- Keep remote channel responses in a target-scoped in-memory snapshot so they cannot overwrite the connected radio's locally persisted channels.
- Abort reads and writes if the connection changes, including reconnecting to the same radio through a different connection.
- Update the Remote Administration documentation for remote channel editing.

This is PR 4 of the native Remote Admin parity stack tracked by #2410 and depends on #2413 (`codex/remote-admin-feedback`). Its base should remain that branch until the preceding layer lands.

## Why did it change?

The Apple app only exposed channel editing for the directly connected radio, leaving PKI-administered nodes without the channel workflow available on Android.

Android reads remote channel slots in sequence and advances the flow from each matching response. See the pinned [remote channel response chain](https://github.com/meshtastic/Meshtastic-Android/blob/5f1a4e58307d30a0a23c3e0b206562448b4ba00a/feature/settings/src/commonMain/kotlin/org/meshtastic/feature/settings/radio/RadioConfigViewModel.kt#L1270-L1297). Its native editor exposes the channel name, PSK, uplink/downlink flags, and position precision in [EditChannelDialog](https://github.com/meshtastic/Meshtastic-Android/blob/5f1a4e58307d30a0a23c3e0b206562448b4ba00a/feature/settings/src/commonMain/kotlin/org/meshtastic/feature/settings/radio/channel/component/EditChannelDialog.kt#L54-L145). This change brings that capability to Apple platforms while adding explicit readback confirmation and SwiftData isolation for the Apple app's persistence model.

## How is this tested?

- Coverage includes response correlation, retry and timeout behavior, readback mismatch rejection, disabled-channel clearing, primary-slot validation, local database isolation, cancellation, and same-radio connection replacement.
- Real bench: iOS Simulator through a serial/TCP bridge to MUZI (2.8.0), administering TBEAM S3 over RF (2.8.1). Read all eight remote slots and opened the shared primary-channel editor. Values matched an independent serial snapshot, including name, role, position precision, and MQTT flags. Verified Back through Remote Settings to the originating node.
- The hardware run performed reads only; channel writes/readback, mismatch, and partial failures were verified by automated transport tests. The UI automation interface did not expose the editor’s Save button, so an end-to-end UI save was not exercised.

The before screenshot shows the original remote Settings entry using a seeded node. The after screenshots use the real TBEAM bench target and the same editor used by regular Settings.

| Before: remote Settings entry | After: remote channel list |
| --- | --- |
| Original remote Settings entry | All eight real remote channel slots |

The unchanged regular channel editor displaying the remote primary channel

## Checklist

### DIFF
--- a/Meshtastic/Accessory/Accessory Manager/AccessoryManager+ToRadio.swift
+++ b/Meshtastic/Accessory/Accessory Manager/AccessoryManager+ToRadio.swift
@@ -798,6 +798,9 @@ extension AccessoryManager {
 	public func saveChannel(channel: Channel, fromUser: UserEntity, toUser: UserEntity) async throws -> Int64 {
 		var adminPacket = AdminMessage()
 		adminPacket.setChannel = channel
+		if fromUser != toUser {
+			adminPacket.sessionPasskey = toUser.userNode?.sessionPasskey ?? Data()
+		}
 		var meshPacket: MeshPacket = MeshPacket()
 		meshPacket.to = UInt32(toUser.num)
 		meshPacket.from	= UInt32(fromUser.num)
@@ -815,6 +818,23 @@ extension AccessoryManager {
 		let messageDescription = "🛟 Saved Channel \(channel.index) for \(toUser.longName ?? "Unknown".localized)"
 		try await sendAdminMessageToRadio(meshPacket: meshPacket, adminDescription: messageDescription)
 		return Int64(meshPacket.id)
+	}
+
+	/// Requests one channel from a remote admin target. The wire request is one based while
+	/// the response's Channel.index remains zero based.
+	public func requestRemoteChannel(index: Int32, fromUser: UserEntity, toUser: UserEntity) async throws -> UInt32 {
+		guard (0...7).contains(index) else {
+			throw AccessoryError.appError("Channel index must be between 0 and 7")
+		}
+		let sessionPasskey = toUser.userNode?.sessionPasskey ?? Data()
+		let packet = try RemoteChannelsPacketBuilder.getRequest(
+			index: index, from: UInt32(fromUser.num), to: UInt32(toUser.num), sessionPasskey: sessionPasskey
+		)
+		try await sendAdminMessageToRadio(
+			meshPacket: packet,
+			adminDescription: "🛟 Requested remote Channel \(index) for \(toUser.longName ?? "Unknown".localized)"
+		)
+		return packet.id
 	}
 
 	/// Join a mesh advertised by a beacon: set the primary channel to the offered channel (name +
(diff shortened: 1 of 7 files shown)


garthvh:

Specific to this one: the remote channel design here is the interesting part — sequential reads of all eight slots, held in a target-scoped in-memory snapshot so they cannot overwrite the connected radio's channels. That is worth discussing on its own, because it sidesteps a schema change we would otherwise need. It also appears to be the first code in the app that sends get_channel_request, which is a question we have not settled: whether current firmware answers one from a remote node over the mesh, and whether the response carries the PSK. That needs an answer before the UI, not after.

garthvh:
Reopened. I closed this earlier on its description and its file count without reading the diff, which was wrong — the code is careful and it answers a question we have open.

What holds up:

- Responses are correlated on both `packet.from == target` and `decoded.requestID`, not just one of them.
- `getRequest`, `matchesResponse`, `WritePlan.execute` and `retry` are pure and testable, with the transport behind a protocol so the view model can be exercised without a radio.
- The read loop checks `Task.isCancelled` per slot rather than only at the top.
- Saves write, read the slot back, and compare the editable fields before reporting success. That is the right shape and it is more than the rest of the app does.
- It fixes a real bug on the way past: `saveChannel` never set `sessionPasskey`, so a remote channel write would have been rejected by a PKI-administered node. That is worth landing on its own regardless of what happens to the rest of this.

Two things to settle before it goes in:

1. **Does firmware answer a `get_channel_request` from a remote node over the mesh, and does the response carry the PSK?** Everything here rests on that and I could not confirm it from this side. If the answer is no, the screen shows only what the app itself wrote, which is a different feature and needs to be presented differently. A one-off probe against a node you administer settles it.
2. **The in-memory target-scoped snapshot is a real design decision, not an implementation detail.** It avoids giving `ChannelEntity` a node key and the schema migration that would come with it, at the cost of the remote table being lost on dismissal. That trade is defensible, but it should be a decision we take deliberately, because the alternative is a fair amount of work in the other direction.

The blocker is structural: this sits fourth in a five PR chain. Rebase it onto `main` on its own and it can be reviewed on its merits. See #2414.

PR #2412, opened 2026-09-05
Add remote Set Time and reset actions for Android parity

RCGV1:
## What changed?

- Add native remote Set Time, Factory Reset, and NodeDB Reset actions to the node administration section.
- Preserve both reset choices: configuration-only or full factory reset, and NodeDB reset with or without favorites.
- Bind confirmation dialogs to an immutable target snapshot and verify the connected radio, concrete connection, target, and live PKI session immediately before sending.
- Use the existing production transport and authenticated admin packet path, including the remote session passkey.
- Keep local database clearing and reconnect side effects limited to resets of the directly connected radio.
- Show send progress and errors for local transport failures, expired sessions, disconnects, and changed targets. Completion means the command was sent; these actions do not yet wait for a remote acknowledgment.
- Update the Remote Administration documentation for the new actions.

This is PR 5 of the native Remote Admin parity stack tracked by #2410 and depends on #2411 (`codex/remote-admin-channels`). Its base should remain that branch until the preceding layer lands.

## Why did it change?

Android exposes time synchronization and reset operations for administered nodes through the same request-tracked admin flow used by other remote settings. The pinned Android implementation routes the actions in [RadioConfigViewModel](https://github.com/meshtastic/Meshtastic-Android/blob/5f1a4e58307d30a0a23c3e0b206562448b4ba00a/feature/settings/src/commonMain/kotlin/org/meshtastic/feature/settings/radio/RadioConfigViewModel.kt#L646-L699), and constructs their packet IDs and local-only reset side effects in [AdminActionsUseCase](https://github.com/meshtastic/Meshtastic-Android/blob/5f1a4e58307d30a0a23c3e0b206562448b4ba00a/core/domain/src/commonMain/kotlin/org/meshtastic/core/domain/usecase/settings/AdminActionsUseCase.kt#L39-L122). This change provides the same remote administration capability in the Apple app and guards destructive confirmations against navigation or connection changes.

## How is this tested?

- Coverage verifies the real production connection receives correctly addressed and authenticated Set Time, Factory Reset, and NodeDB Reset packets, including both reset variants. Guard tests cover disconnects, target changes, session expiry, and same-radio connection replacement.
- Real bench: iOS Simulator connected through a serial/TCP bridge to MUZI (firmware 2.8.0), administering TBEAM S3 over RF (2.8.1). An independent serial observer on TBEAM recorded the Set Time command from MUZI. An expired-session attempt was rejected in the app; refreshing the session allowed the command.
- Both reset confirmations named the remote TBEAM and were canceled. Destructive resets were verified with the capture transport, not executed on hardware. Channel and LoRa hashes were unchanged afterward.

The after screenshots use the real bench target. Both confirmations were canceled.

| Before | After |
| --- | --- |
| Original remote administration actions | Remote administration actions on the real TBEAM |

| Factory Reset target | NodeDB Reset target |
| --- | --- |
| Factory Reset confirmation naming the remote target | NodeDB Reset confirmation naming the remote target |

## Checklist

### DIFF
--- a/Meshtastic/Accessory/Accessory Manager/AccessoryManager+ToRadio.swift
+++ b/Meshtastic/Accessory/Accessory Manager/AccessoryManager+ToRadio.swift
@@ -327,6 +327,30 @@ extension AccessoryManager {
 		let messageDescription = "🕛 Sent Set Time Admin Message to the connected node."
 		try await sendAdminMessageToRadio(meshPacket: meshPacket, adminDescription: messageDescription)
 	}
+
+	public func sendTime(fromUser: UserEntity, toUser: UserEntity) async throws {
+		var adminPacket = AdminMessage()
+		adminPacket.setTimeOnly = UInt32(Date().timeIntervalSince1970)
+		if fromUser != toUser {
+			adminPacket.sessionPasskey = toUser.userNode?.sessionPasskey ?? Data()
+		}
+		var meshPacket: MeshPacket = MeshPacket()
+		meshPacket.to = UInt32(toUser.num)
+		meshPacket.from = UInt32(fromUser.num)
+		meshPacket.id = UInt32.random(in: UInt32(UInt8.max)..<UInt32.max)
+		meshPacket.priority = MeshPacket.Priority.reliable
+		meshPacket.wantAck = true
+		var dataMessage = DataMessage()
+		if let serializedData: Data = try? adminPacket.serializedData() {
+			dataMessage.payload = serializedData
+			dataMessage.portnum = PortNum.adminApp
+			meshPacket.decoded = dataMessage
+		} else {
+			throw AccessoryError.ioFailed("sendTime(fromUser:toUser:): Unable to serialize admin packet")
+		}
+		let messageDescription = "🕛 Sent Set Time Admin Message to: \(toUser.longName ?? "Unknown".localized) from: \(fromUser.longName ?? "Unknown".localized)"
+		try await sendAdminMessageToRadio(meshPacket: meshPacket, adminDescription: messageDescription)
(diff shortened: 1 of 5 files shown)


garthvh (changes requested):
These need to be individual pr's, not going to be merging these stacks of connected pr's that contain each other

garthvh:

Specific to this one: the description says "Completion means the command was sent; these actions do not yet wait for a remote acknowledgment." Factory Reset and NodeDB Reset on a node reachable only over the mesh, reported as complete without confirmation, is the wrong default. If these come back they need to wait for the routing ACK before saying anything succeeded.

garthvh:
Reopened. My close cited the one caveat in your own description and ignored the rest of the code, which was not a fair reading.

`RemoteAdminActionGuard` is the best thing here. Capturing the target as an immutable snapshot — node, name, radio number and `ObjectIdentifier(connection)` — then re-verifying connected state, radio, that exact connection and a live session immediately before sending is protection against a confirmation dialog outliving the context it was opened in. We have fixed several bugs of exactly that shape recently. It is thirty lines, fully injectable, and tested.

`sendTime` follows the conventions of the file, including the session passkey when `fromUser != toUser`.

The gap is the one you name: completion means the packet was sent. For Set Time that is tolerable. For Factory Reset and NodeDB Reset against a node reachable only over the mesh it is not — the user is told a destructive operation succeeded when all we know is that it was queued. These should wait for the routing ACK and report the error path, the way #2413 does for config saves. That is the change I would want before this lands.

Structural blocker as with the rest of the chain: it is fifth of five. Against `main` on its own it is a reviewable change. See #2414.

PR #2413, opened 2026-09-05
Track remote configuration progress and retries for Android parity

RCGV1:
## What changed?

- Track each remote configuration request and save by target node, settings section, and packet ID.
- Keep request progress visible until the matching admin response arrives, and keep save progress visible until the matching routing ACK or error arrives.
- Show timeouts and routing failures in the settings screen with an in-place retry action instead of dismissing the editor as soon as the packet is queued.
- Fail outstanding operations when the radio disconnects, while ignoring unrelated ACKs and admin responses.
- Preserve the admin session when module responses arrive or cached configuration is updated, so a successful retry loads the editor without erasing its session.
- Update the Remote Administration documentation for the confirmed-save behavior.

This is PR 3 of the native Remote Admin parity stack tracked by #2410 and depends on #2409.

## Why did it change?

Remote configuration previously treated a successful local write to the transport as completion. That could dismiss an editor even when the remote node rejected the request, never received it, or failed to respond.

Android already keeps request IDs in a loading state, applies a 30-second timeout, distinguishes read responses from routing ACKs, and defers completion until the response appropriate to the operation arrives. See the pinned Android implementation in [request registration and timeout handling](https://github.com/meshtastic/Meshtastic-Android/blob/5f1a4e58307d30a0a23c3e0b206562448b4ba00a/feature/settings/src/commonMain/kotlin/org/meshtastic/feature/settings/radio/RadioConfigViewModel.kt#L1048-L1095) and [response completion handling](https://github.com/meshtastic/Meshtastic-Android/blob/5f1a4e58307d30a0a23c3e0b206562448b4ba00a/feature/settings/src/commonMain/kotlin/org/meshtastic/feature/settings/radio/RadioConfigViewModel.kt#L1376-L1403). This change gives the Apple app the same user-visible completion semantics while retaining its native SwiftUI and transport architecture.

## How is this tested?

- Tests cover packet/target correlation, pending saves until remote ACK, timeout, disconnect, unrelated responses, forced retry, and preservation of session keys during module responses and cache updates.
- Simulator with a localhost protocol fixture: suppressed a fresh MQTT response until the 30-second timeout, then restored responses and clicked Retry. Verified a new request ID, the matching response, initialized editor values, and the remote administration header without another session refresh. This controlled failure test uses a simulated radio.
- Real bench: MUZI (2.8.0) → TBEAM S3 (2.8.1) over RF, with the simulator connected through a serial/TCP bridge. A fresh MQTT module response retained the session and initialized values matching an independent serial read, without manually refreshing the session. The stack also passed eight channel reads and node/Settings Back navigation. Hardware destructive resets and channel writes were not performed.

The before image shows the previous Ambient Lighting feedback. The after sequence exercises MQTT timeout and retry using a controlled protocol fixture.

| Before: no-response message | After: timeout with Retry |
| --- | --- |
| Previous configuration no-response feedback | MQTT timeout with actionable Retry |

| Request pending | Retry succeeded and loaded remote values |
| --- | --- |
| Configuration request remains pending | Successful Retry retains the remote session and initializes the editor |

Real RF request and response:

| Requesting TBEAM MQTT configuration | Remote values loaded with session intact |
| --- | --- |
| Real RF module request pending | Real TBEAM response with preserved administration session |

## Checklist

### DIFF
--- a/Meshtastic/Accessory/Accessory Manager/AccessoryManager+ToRadio.swift
+++ b/Meshtastic/Accessory/Accessory Manager/AccessoryManager+ToRadio.swift
@@ -200,7 +200,23 @@ extension AccessoryManager {
 		toRadio = ToRadio()
 		toRadio.packet = meshPacket
 
+		let operationID = RemoteAdminConfigTracker.currentOperationID
+		let enrolled = remoteAdminConfigTracker.registerPacket(
+			packetID: meshPacket.id,
+			targetNodeNum: Int64(meshPacket.to),
+			operationID: operationID
+		)
 		try await send(toRadio)
+		if enrolled, let operationID {
+			switch await remoteAdminConfigTracker.waitForPacket(packetID: meshPacket.id, operationID: operationID) {
+			case .succeeded:
+				break
+			case .failed(let message):
+				throw AccessoryError.ioFailed(message)
+			case .timedOut:
+				throw AccessoryError.timeout
+			}
+		}
 		if let adminDescription {
 			Logger.admin.debug("\(adminDescription, privacy: .public)")
 		}
--- a/Meshtastic/Accessory/Accessory Manager/AccessoryManager.swift
+++ b/Meshtastic/Accessory/Accessory Manager/AccessoryManager.swift
@@ -162,6 +162,8 @@ class AccessoryManager: ObservableObject, MqttClientProxyManagerDelegate {
 	/// back off repeats (see `shouldSurfaceFirmwareNotice`). In memory on purpose: after a
 	/// relaunch a standing notice alerts once more, which is the useful behaviour.
 	var firmwareNoticeHistory: [String: (count: Int, lastShown: Date)] = [:]
+	let remoteAdminConfigTracker = RemoteAdminConfigTracker()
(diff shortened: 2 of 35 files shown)


garthvh:

Specific to this one: 35 files for request and save progress tracking is a lot of surface for what it does. If confirmed saves are worth having, that is its own change against main.

garthvh:
Reopened. My close said 35 files was a lot of surface for what this does. That was a count, not a review, and it was wrong about the shape.

The 35 files are one new type — `RemoteAdminConfigTracker`, 144 lines — plus 194 lines of tests for it, and then roughly a dozen lines each across the config screens adopting it. That is the right shape for making every config screen report progress: one place with the logic, uniform small edits at the call sites.

Keeping progress up until the matching admin response or routing ACK arrives, and offering retry in place rather than dismissing the editor when the packet is queued, is the behaviour we want. #2412 in this same chain should use it for the destructive actions.

Two things I would look at:

- The 50 lines in `UpdateSwiftData.swift` are the part that deserves the most attention and got the least description. That file is the persistence hot path.
- What happens to an operation whose view goes away while it is outstanding — does the tracker drop it, and does anything still hold the entry?

Structural blocker as with the rest: third of five in a chain. See #2414.

PR #2428, opened 2026-09-05
Start location updates only when needed

bruschill:
## Summary

Core Location now starts when an app feature needs location and stops when that demand ends. Opening the app on the Connect tab without a connected radio no longer starts location updates.

## Background

[PR #441](https://github.com/meshtastic/Meshtastic-Apple/pull/441) changed `MeshtasticAppDelegate` in December 2023 from restarting previously active location updates to starting them at app launch. This made the latest phone location available to features throughout the app.

[Commit d477de80](https://github.com/meshtastic/Meshtastic-Apple/commit/d477de80c2865593c80bbd7f0a7ed0201d88aa0d) last changed that launch call in October 2024 when it removed the iOS 16 availability condition. The unconditional startup behavior otherwise remained the same.

The app now has several location consumers with different lifetimes, including maps, node distances, route recording, Watch sync, and connected-radio position sharing. This change keeps those uses while tying Core Location delivery to current demand.

## What changed

- Track location demand from visible map and distance UI, route recording, connected-radio position sharing, Watch sync, and continuous background updates.
- Pause foreground-only demand while the app is inactive and stop delivery after the last consumer releases it.
- Release connected-radio demand on disconnect and reconcile the sharing toggle immediately.
- Wait for a bounded one-shot location when Watch sync has no cached fix.
- Refresh node distances and nearby MQTT topics after the first phone location arrives.
- Add lifecycle tests for overlapping consumers, app activity, authorization changes, recording, Watch timeouts, and permission retries.

## Testing

- `xcodebuild test -project Meshtastic.xcodeproj -scheme Meshtastic -destination 'platform=iOS Simulator,id=59BB9160-4ED9-4606-8EF3-305BF07AAA88'`: 3,083 tests in 533 suites passed.
- Focused location and Watch suites: 15 tests in 3 suites passed.
- iOS Simulator with a Meshtastic TCP replay radio: no location updates before connection; updates started after the connection reached `Subscribed`; a simulated location arrived; updates stopped when the TCP radio disconnected.
- `git diff --check`

Background route recording and continuous background updates were not exercised on a physical device.

### DIFF
--- a/Meshtastic/Accessory/Accessory Manager/AccessoryManager+Position.swift
+++ b/Meshtastic/Accessory/Accessory Manager/AccessoryManager+Position.swift
@@ -16,10 +16,12 @@ extension AccessoryManager {
 	}
 
 	func initializeLocationProvider() {
+		updateLocationProviderDemand()
 		self.locationTask = Task {
 			repeat {
 				let sleepSeconds = Self.locationProviderSleepSeconds(configuredInterval: UserDefaults.provideLocationInterval)
-				try? await Task.sleep(for: .seconds(sleepSeconds)) // Throws if task is cancelled
+				try? await Task.sleep(for: .seconds(sleepSeconds))
+				updateLocationProviderDemand()
 
 				guard let fromNodeNum = activeConnection?.device.num else {
 					return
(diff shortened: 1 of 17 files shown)


garthvh:
Closing this.

The app uses location for plenty of features that have nothing to do with a connected radio — distance in the node list and contacts, the compass, the map showing your own position, route recording, nearby MQTT topics. Demand modelled around connection state does not match how location is actually used here, and the added machinery is not warranted for what it buys.

Not a comment on the code quality. If the always-on start is worth addressing later, it should be a much smaller change than 429 lines across 17 files — and worth noting that three of the four non-UI purposes here are not really consumers: `routeRecording` and `continuousBackground` are booleans `LocationsHandler` already owns, and `watchSync` is a one-shot that acquires and releases inside its own `defer`.

Thanks for the work on it.

bruschill:
@garthvh I'd be interested in simpler suggestions if you have any. Seeing an app constantly holding my location is annoying at the very least.

PR #2437, opened 2026-09-07
Update protobufs for T-Connect Pro and Codec2

bruschill:
## What changed?

Updated the protobuf submodule to `f008c45` and regenerated the Swift sources with the pinned SwiftProtobuf plugin.

This adds:

- The T-Connect Pro hardware model.
- Codec2 700C and 450 bitrate values.
- Audio settings for the new bitrate values while retaining deprecated 700 and 700B values for existing configurations.
- String-catalog entries for the updated Audio settings text.

## Why did it change?

The upstream protobuf definitions added T-Connect Pro and replaced the removed Codec2 700 modes with 700C and 450. The Apple app needs the generated definitions to decode those values and needs matching picker options to load and save them.

Existing raw values 7 and 8 remain available and are labeled deprecated. New values use raw values 9 and 10 from the protobuf definition.

## How is this tested?

- Regenerated twice with the pinned `protoc-gen-swift` 1.36.1 plugin and confirmed byte-for-byte identical output.
- Built `MeshtasticProtobufs` successfully.
- Built the Meshtastic iOS app for the iPhone 17 Pro simulator.
- Ran strict SwiftLint on `AudioConfig.swift`: 0 violations.
- Checked the changed Swift files with SourceKit-LSP: no diagnostics.
- Validated `Localizable.xcstrings` as JSON.
- Ran `git diff --check`.

Not captured. The change adds 700C and 450 to the existing Audio bitrate picker and labels 700 and 700B as deprecated.

## Checklist

- [x] I have verified whether these changes require updates to the in-app documentation under `docs/user/` or `docs/developer/`, and updated accordingly. The existing Audio settings documentation already describes bitrate configuration, so no documentation update is needed. The `skip-docs-check` label is applied.

## Summary by CodeRabbit

* **New Features**
  * Expanded the Codec2 bitrate picker with 700C, 450, 700, and 700B options.
* **Documentation**
  * Updated the bitrate description to clarify that 700C replaces 700 and 700B.
  * Added guidance that lower bitrates use less bandwidth but may reduce audio quality.


garthvh:
Closing this as superseded — thanks for it, the substance did land, just through other pull requests.

- The submodule bump and regeneration landed in f1d304dc (#2487) and b7708f93 (#2506). `T_CONNECT_PRO`, `CODEC2_700C` and `CODEC2_450` are all on main today.
- Main's submodule is now `84d5b899`, which is a descendant of this branch's `f008c459`, so merging as-is would move the submodule backwards and drop `AXIOMETA_GENESIS_MINI` from `HardwareModel`.
- The `project.pbxproj` change targets a file exception set that no longer exists now that the project uses synchronized folders.

One real gap remains, and it is worth stating so it is not lost: the Audio screen's bitrate picker still stops at 700B, so 700C and 450 are not selectable in the app. The field-metadata registry already carries the new description and marks 700/700B deprecated, so when Audio moves to the schema-driven form the two new rates appear on their own and the deprecated ones hide unless the radio is on them. If that migration is not close, adding the two picker lines to the existing screen is a small change on main.

PR #2465, opened 2026-09-09
Use a mesh radio icon for the signed node indicator

garthvh:
## Summary

The signed node indicator borrowed `person.badge.shield.checkmark.fill`, so a row about a radio was drawn with a person. This swaps in a custom symbol built from the app's existing mesh radio glyph.

## What changed

- New custom symbol `custom.mesh.radio.badge.shield.checkmark` in the asset catalog, next to the other `custom.*` symbols. It is the `custom.mesh.radio` glyph with a filled shield and knocked-out checkmark badged on the lower right, following the `person.badge.shield.checkmark.fill` layout. The radio's outline is broken where the badge overlaps it, with the gap the same width as the body stroke, so the two shapes read as separate objects.
- `SignedNodeIcon` now names that symbol and vends it as an `Image`. Asset catalog symbols only load through `Image(_:)`; `Image(systemName:)` finds nothing and SwiftUI silently draws nothing, so the view is vended rather than the name to keep callers from reaching for the wrong one out of habit.
- `IconAndText` gained an `init(icon:)` for asset catalog symbols. The existing `init(systemName:)` is unchanged, so its other 20-odd call sites are untouched.
- The node list rows and node detail use the new symbol.
- The node list help had no entry for this indicator, so it gets one, in a new Security section. It is deliberately not in `LockLegend`: that legend is shared with the Direct Messages help, where a per-node trust signal has nothing to explain.
- Updated the sentence in the nodes doc that described the old glyph.

## Testing

- `xcodebuild test -only-testing:MeshtasticTests/SignedNodeIconTests` — 2 tests pass. The resolution test now uses `UIImage(named:in:with:)`, since `UIImage(systemName:)` returns nil for a custom symbol and the old test would have passed for the wrong reason.
- `actool` compiles the symbolset with no errors and `assetutil` shows it in the compiled `Assets.car` under the expected name.
- Rendered the glyph at icon sizes to check the badge and checkmark stay legible.

## Summary by CodeRabbit

* **New Features**
  * Added a security explanation to the Node List help screen for signed nodes and automatic identity verification.
  * Updated signed-node indicators across node lists and detail views with a radio-and-shield badge.
* **Documentation**
  * Clarified that the Signed Node row appears in both node lists and detail views.
  * Updated icon descriptions to reflect the radio-and-shield design.
* **Bug Fixes**
  * Improved rendering of the signed-node badge using the correct custom visual asset.


garthvh:
Superseded by the artwork-only change. This branch was cut from a stale main and predates #2449 through #2459, which already added the custom signed-node symbol, moved signing into the icon beside the name and removed the Signed node row — so merging this would have regressed all of that.

Issue #2444, opened 2026-09-08
🐞 [Bug]: Node list stutters when scrolling


iamamemario:
### Before submitting

- [x] I have searched existing issues to make sure this bug hasn't already been reported
- [x] I have updated to the latest version of the software to verify the issue still exists

### Firmware Version

2.7.26

### What did you do?

scroll with a node list of at least 300 nodes.

### Expected Behavior

the node list should have smooth scrolling

### Current Behavior

scrolling stutters

### Participation

- [ ] I am willing to submit a pull request for this issue.

### Additional comments

I am using a iPhone 17 pro with the latest iOS 26.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #2472, opened 2026-09-10
Defer node list refreshes while scrolling

bruschill:
## What changed?

- Refresh the node snapshot only after node data, filters, the active radio, or the location used by distance filtering changes.
- Coalesce pending refreshes on the existing 350 ms cadence.
- Defer full-list fetch, filter, and sort work while the list is scrolling or decelerating on iOS 18 and later.
- Preserve the current snapshot and retry after fetch or filter lookup failures.
- Add tests for refresh-state transitions, failure retries, and scrolling behavior.

## Why did it change?

The Nodes screen rebuilt its complete filtered snapshot every 350 milliseconds while the tab was active, even when its inputs had not changed. At 300 nodes, profiling found refresh bursts with a 33 ms median and 45 ms p95. Each exceeded one 60 Hz frame.

Addresses #2444.

## How is this tested?

- Full test suite: 3,148 tests passed.
- Focused `NodeListRefreshStateTests`: 7 tests passed.
- SwiftLint passed.
- iOS and Mac Catalyst builds passed.
- SourceKit-LSP diagnostics are clean.
- A 300-node simulator trace reduced recurring full-list refresh bursts from 69 to 0.
- On an iPhone 16 Pro Max running iOS 27.0, five measured up/down scroll iterations with static seeded data averaged 0.2 hitches per iteration, a 0.665 ms/s hitch ratio, and 85.013 fps.
- A standalone 30-second physical-device Instruments trace recorded three manual scroll/deceleration intervals with no full-list refresh overlapping scrolling. It still recorded 123 rendering hitch events during scrolling: 12.5 ms median, 50.0 ms p95, and 62.5 ms maximum. This PR does not claim to eliminate all node-row rendering hitches.

[issue-2444-physical-manual-scroll.trace.zip](https://github.com/user-attachments/files/32066752/issue-2444-physical-manual-scroll.trace.zip)

## Checklist

## Summary by CodeRabbit

* **Improvements**
  * Node lists now update automatically when filters, device status, saved data, or location information changes.
  * Refreshes are coordinated to reduce unnecessary work and are deferred while scrolling for smoother navigation.
  * Refresh failures are handled more reliably, keeping updates pending so they can be retried.
  * Location updates are now distributed consistently to supported parts of the app.


RCGV1 (commented):
Independent profiling with a 5,000-node simulator database confirms the refresh loop is the dominant idle cost. The current 350 ms full-refresh baseline averaged 60.8% CPU; an invalidation-driven prototype averaged 1.94% under the same idle sample, a 96.8% reduction. That is strong evidence for this PR’s overall direction.

I found one remaining cadence issue inline: enabling the online filter still schedules the expensive path every 350 ms. No hardware was involved in these measurements.

Meshtastic/Views/Nodes/NodeList.swift
```diff
@@ -405,37 +434,71 @@ private struct FilteredNodeList: View {
...
-			// ingestion): the stale snapshot otherwise holds dead nodes for as long as the
-			// user stays away, and any row re-evaluation in that window hits them.
-			refreshDisplayedNodes()
+			// Refresh immediately when entering the tab. While the tab stays visible, saves and
+			// filter changes are coalesced on the existing cadence. Expensive work waits until
+			// scrolling ends so it cannot interrupt interaction or deceleration.
+			runPendingRefresh()
 			while !Task.isCancelled {
 				try? await Task.sleep(for: .milliseconds(350))
+				if filters.isOnline {
```
RCGV1: When `isOnline` is enabled, this marks the full node snapshot dirty every 350 ms, so the fetch/filter/sort path runs about 171 times per minute even with no model changes. Our 5,000-node baseline using that full-refresh cadence averaged 60.8% idle CPU. Because the online cutoff is two hours, a separate 60-second aging task would bound staleness to one minute while reducing time-driven refreshes to one per minute; model/filter saves can keep the existing 350 ms coalescing path. Please add coverage that online aging refreshes are bounded rather than tied to the fast coalescing cadence.

RCGV1: Verified in `48984376`. Online aging is now decoupled from the 350 ms coalescing loop, runs at a 60-second cadence only while the filter is enabled, and has focused cadence/disabled-state coverage. This addresses the remaining issue from my 5,000-node profile.

RCGV1 (approved):
The 5,000-node profiling concern is resolved by `48984376`: online aging now runs once per minute instead of invalidating the full list every 350 ms, with focused tests for cadence and disabled behavior. The change-driven refresh and scroll deferral remain aligned with the measured hotspot. Approved pending the repository's normal CI.

garthvh:
Closing this. Thanks for the physical-device trace — publishing the post-fix numbers is what made the call clear, and I'd rather have that than a PR that only reports the "before".

The issue is #2444, "node list stutters when scrolling". The trace here still records 123 rendering hitch events during scrolling, 12.5 ms median and 50.0 ms p95, at 85 fps on a 120 Hz device. That p95 is higher than the 45 ms refresh burst this removes. So the refresh bursts are genuinely gone, 69 to 0, but they weren't what the reporter was feeling — the stutter is still there.

Against that, this adds a second refresh-gating axis (`needsRefresh && !isScrolling`) to a view that already carries a throttled snapshot, a purge cadence and a container-generation guard. That machinery is where the NodeListItem and NodeDetail crashes come from — stale snapshots holding deleted entities — so more state in that file has to buy more than one contributor out of several.

The "refresh only when inputs changed" half also buys least on the meshes that need it most: when node data is changing constantly, the gate is true nearly every tick anyway.

What would actually close #2444 is working out what those 123 row-rendering hitches are. Your trace is the best evidence anyone has on that — worth attaching to the issue so it isn't lost with this PR.

PR #2489, opened 2026-09-14
Localize the app's own enum strings

garthvh:
## Summary

195 strings in the app's enums were never reaching translators, and nor were the 19 interval picker labels. Both render in English today regardless of device language.

First of three changes from [spec 019](https://github.com/meshtastic/Meshtastic-Apple/pull/2487), but it stands alone — these strings are untranslated whether or not settings search ever ships.

## What was wrong

The enums went through a helper whose receiver is a variable:

```swift
var localized: String { NSLocalizedString(self, comment: self) }
```

Xcode's extractor only sees string *literals*, so none of these were in `Localizable.xcstrings`. Converted to `String(localized:comment:)`, which still returns `String`, so no call site changed. The comment names the enum and property — the context a translator needs for strings like "Back", "Select" and "Right".

`UpdateIntervalPicker` had the same problem by a different route: `pickerLabel` was a `String`, which binds `Picker`'s non-localizing `StringProtocol` overload. Changing the type to `LocalizedStringKey` fixes all 19 labels without touching a single call site, since they already pass literals.

## Four smaller bugs found along the way

- `ScreenUnits` returned bare `"Metric"`/`"Imperial"` — not localized at all, not even the broken kind.
- `ScreenCarouselIntervals` returned `"off".localized.capitalized`. The catalog key is `"Off"`, and capitalizing a translation is wrong in locales whose casing rules differ.
- `MapLayer` and `OverlayType` returned `self.rawValue.localized`, so the map settings UI displayed raw case names — `tileServer`, `geoJson` — with a `.capitalized` call papering over it.
- `DetectionSensorRole.description` returned bare literals, and its `.localized` property was dead code.

## One key split

`"Standard"` was a single key shared by two map-style enums and the offline-map download detail level, under a comment describing the wrong case ("Label for a high-detail offline map download" — on the *standard* case). The download levels now have explicit keys; the map styles keep the shared one, since they do mean the same thing.

## Not in scope

123 `.localized` sites remain, all in the 13 proto-backed enums — `RegionCodes`, `ModemPresets`, `DeviceRoles` and the rest. Those take their strings from the protobuf schema in a later change, so localizing them here would be thrown away.

## Testing

- Builds clean; SwiftLint reports only three warnings that pre-exist on `main`.
- 3,143 tests in 541 suites pass.
- Verified via the compiler's `.stringsdata` that extraction now works: 36 strings from `RoutingError`, 37 from `IntervalEnums`, and all 19 interval labels from their own call sites.

`Localizable.xcstrings` is deliberately **not** updated here. `xcodebuild` emits `.stringsdata` but only the Xcode GUI writes back to the catalog, and that diff is worth making on its own so it can be checked for being purely additive — the file is 3.1 MB and 1,842 keys.

## Summary by CodeRabbit

* **Localization**
  * Improved localization support across map settings, device configuration, status labels, intervals, routing, messaging, telemetry, and other app text.
  * Added clearer translation keys and context for offline map detail labels.
  * Map layer names now preserve their localized capitalization.
  * Picker labels are now properly recognized as localized text.


garthvh:
The `Committed project matches project.yml` failure is pre-existing and unrelated to this change — this PR touches no `project.yml` or `.xcodeproj` file.

Reproduced on a clean detached checkout of `origin/main` with no edits at all:

```
$ git worktree add --detach /tmp/wt-main origin/main
$ cd /tmp/wt-main && xcodegen generate
$ git diff Meshtastic.xcodeproj
-				456E28D41A9B7B1788FEE0CD /* Shared */,
 				5A09511A8EFEB6219741EDDB /* Shared */,
+				456E28D41A9B7B1788FEE0CD /* Shared */,
```

`project.yml` declares `- path: "Shared"` twice, at lines 210 and 406, for two different targets. That yields two groups with the same name in the root group list, and xcodegen's sort is not stable for equal names, so the pair swaps roughly half the time. Recent PRs have been passing on the winning side of the coin flip.

Happy to open a separate issue, or fix it by disambiguating one of the two group names — but not in this PR, which is a localization change.

garthvh:

PR #2490, opened 2026-09-14
Fix the PAX Counter gate, drop a dead TAK section

garthvh:
Two small settings bugs found while surveying every control for [spec 019](https://github.com/meshtastic/Meshtastic-Apple/pull/2487). Unrelated to that work, so they are here on their own.

## PAX Counter screen disabled by the wrong config

```swift
.disabled(!accessoryManager.isConnected || node?.powerConfig == nil)
```

A node reporting a PAX Counter config but no power config gets the entire form greyed out, with nothing saying why. Every sibling module screen gates on its own config — `audioConfig`, `externalNotificationConfig`, `detectionSensorConfig`, `cannedMessageConfig` — so this was a copy-paste. Now gates on `paxCounterConfig`.

## Dead `takSection`

`Settings.swift` still declared `var takSection`, unreferenced since 2.7.10 ([#1648](https://github.com/meshtastic/Meshtastic-Apple/pull/1648)) moved TAK into the module configuration section. Worth removing rather than leaving, because the comment inside it explains why two entry points are deliberately kept — which stopped being true when that same change removed it from the body, so it reads as a live decision rather than leftovers.

## Deliberately not changed

`.takConfig` has no list row and is reachable only by deep link. That looks like an omission, but #1648 added the destination and removed the row in the same commit, so it is intentional.

## Testing

Builds clean. No behaviour change beyond the gate, which only ever disabled a screen that should have been editable.

## Summary by CodeRabbit

- **Bug Fixes**
  - The PAX Counter settings are now correctly disabled when no device is connected or when the device lacks PAX Counter configuration.

- **Settings**
  - Removed the duplicate TAK Server entry from the standalone settings section. The link remains available under Module Configuration.


garthvh:

PR #2491, opened 2026-09-14
Settings search

garthvh:
## Summary

Search the Settings screen by control name, keyword or option value. Typing filters the list in place rather than pushing a results screen, so clearing the field puts the user back where they were, and each result names the screen and section it came from — "Enabled" labels six different controls, so the breadcrumb is what tells them apart.

Implements [spec 019](https://github.com/meshtastic/Meshtastic-Apple/pull/2487) US1. Depends on [protobufs#952](https://github.com/meshtastic/protobufs/pull/952) and [#1081](https://github.com/meshtastic/protobufs/pull/1081); the submodule currently points at #1081's branch and **must be re-pointed at `master` before this merges**.

## The index has two halves

Proto-backed settings come from `FieldMetadataRegistry`, generated from the schema, so their labels cannot drift from what every other client reads. App-level settings have no protobuf behind them and are curated in one file, which keeps "what is indexed" readable in one place instead of assembled from twenty views.

`scripts/gen_protos.sh` gains a phase that generates the registry — 318 entries, 280 labels, 70 descriptions.

**It generates into the app target, not the package, and that is the point.** The registry emits labels as `String(localized:)`, and Xcode only extracts those from a target that owns a string catalog and sets `SWIFT_EMIT_LOC_STRINGS`. A SwiftPM package has neither, so a registry there would compile, run, and resolve to English forever with no error anywhere. Verified the right way round: the build emits **370 strings** from that file, including all 36 region names, which have been English for every non-English user until now.

## Two things the generator cannot do for us

**It emits no imports**, because it cannot know which module holds the protobuf types. The script adds `import MeshtasticProtobufs`.

**Its per-message static accessors break ordinary code across modules:**

```swift
loraConfig.modemPreset = .shortFast
// error: static member 'modemPreset' cannot be used on instance of type 'Config.LoRaConfig'
```

Reads are fine; writes are not. Same file it compiles, across modules it does not — and our layout is necessarily cross-module for the reason above. Reduced to a two-module repro and [reported upstream](https://github.com/meshtastic/protobufs/pull/952#issuecomment-5658336177). Until it is fixed, `scripts/strip-fieldmeta-message-accessors.py` removes those blocks and we use `FieldMetadataRegistry.get(type:tag:)`, which is what the index needs anyway. If upstream nests the accessors, that script becomes a no-op and should be deleted.

## Ranking and visibility

Field-weighted: exact label, prefix, substring, keyword, then description. Ties break by section then label, so the same query always produces the same order — scoring alone is unstable and would reshuffle results between keystrokes. Matching is `localizedStandardContains`, so "resume" finds "résumé".

Visibility is computed per query rather than baked into the index, which stays static. One mechanism covers every reason a result is not plainly available — needs a radio, DIY-only hardware, admin-only, deprecated — so nothing disappears without a reason the row can state.

## Three guards, because a catalogue this size rots quietly

- **Completeness** parses the config protos and fails when a field with a settings screen is neither indexed nor exempted for a stated reason. It immediately found two real regressions: `rx_gpio` and `tx_gpio` had lost their labels in an upstream rebuild. Fixed in #1081.
- **Structure** asserts every destination is real, no label is blank or padded, no two entries collide on one screen, and no label leaks a raw catalog key.
- **Counts** are pinned per screen, so a control added without an entry fails rather than going quietly unsearchable.

The exemption list is 37 fields and each names why. Most are firmware settings this client does not offer yet — `buzzer_mode`, `ipv6_enabled`, the health telemetry fields — verified to have no control at all. They should leave the list when the app grows a control, which is the prompt the list is meant to give.

SC-002 is a query corpus — "hops", "psk", "transmit power", "region" — each asserted to reach its control, rather than requiring every entry to carry a keyword it does not need.

## Also here

- SwiftLint excludes the generated registry (lines past the 400-character limit) and the `protobufs` submodule, where building the plugin leaves a SwiftPM checkout of swift-protobuf's own tests.
- `unit-tests.yml` checks out submodules. Without it the completeness test parses an empty directory and passes for the wrong reason.
- `xcodegen-drift.yml` watches `protobufs`, since a submodule bump regenerates sources.

## Scope note

The submodule bump brings unrelated upstream protos: `SoilWaterMetrics` split out of `EnvironmentMetrics`, `storeforward.original_id`, a `HostMetrics` cap. Those account for most of the `.pb.swift` diff. Nothing in the app referenced the moved telemetry fields.

## Testing

3,156 tests pass, 13 of them new. Builds clean.


garthvh:
The `Committed project matches project.yml` failure is pre-existing and unrelated to this branch, which changes no `project.yml` or `.xcodeproj` file.

Regenerating on a clean detached checkout of `origin/main` produces the identical one-line diff — the same two blob hashes CI reports:

```
-				456E28D41A9B7B1788FEE0CD /* Shared */,
 				5A09511A8EFEB6219741EDDB /* Shared */,
+				456E28D41A9B7B1788FEE0CD /* Shared */,
```

`project.yml` declares `- path: "Shared"` twice — line 210 for `Meshtastic`, line 406 for `WidgetsExtension` — so XcodeGen emits two identically-named synced root groups and its sort is not stable between them.

It is genuinely nondeterministic, not one committed ordering being wrong. Six consecutive generates on an unmodified tree:

```
run 1: 0 changed   run 2: 1 changed   run 3: 0 changed
run 4: 1 changed   run 5: 0 changed   run 6: 1 changed
```

So committing either ordering cannot fix it, and this branch's own generate adds nothing beyond that one line — the new `Meshtastic/Model/Search/` files are inside an existing synced folder and need no project change.

I tried giving the second declaration a distinct `name:`; XcodeGen ignores `name` on a `syncedFolder`, and the flake continued. A real fix probably means one shared synced group rather than two, or pinning the sort upstream. That is its own change and I have left it alone rather than bundle it here.

garthvh:

PR #2501, opened 2026-09-16
Fix XcodeGen message sharing group drift

bruschill:
## Summary

Fix XcodeGen drift caused by two root project groups named `Shared`.

## What changed

- Put the Messages shared sources under a distinct `Message Sharing` navigator group.
- Regenerate the Xcode project with XcodeGen 2.46.0.

## Testing

- Repeated XcodeGen generation is byte-identical.
- Generation from a clean Git archive leaves the tree unchanged.
- `build-for-testing` succeeds on an iPhone 17 Pro simulator.

## Summary by CodeRabbit

* **Refactor**
  * Organized the Message Sharing source folder under the dedicated “Message Sharing” group in the project structure.


garthvh (approved):
Verified this rather than taking the determinism claim on faith, because I spent most of yesterday losing to this exact check.

**It works.** Twelve consecutive `xcodegen generate` runs on this branch produce one hash, matching the committed `project.pbxproj`. On `main` the same loop produces **two** competing hashes, roughly 60/40 — which is why the drift check has been a coin flip on every PR regardless of what it touched, and why a re-run clears it about half the time.

**Why it works, which is worth recording.** On `main` there are two root groups both named `Shared`:

```
456E28D41A9B7B1788FEE0CD /* Shared */
5A09511A8EFEB6219741EDDB /* Shared */
```

XcodeGen sorts the main group's children by name and nothing breaks that tie, so it emits either order. The second one is synthesised for the nested `Shared/MessageSharing` path. Naming that group removes the duplicate, and the tie with it. I had assumed the collision was the app target against `WidgetsExtension`, since both list `- path: "Shared"` — that was wrong, and this is the narrower, correct fix.

**No target gains or loses sources.** I diffed `fileSystemSynchronizedGroups` per target across both revisions:

| Target | before | after |
| --- | --- | --- |
| `MeshtasticMessages` | `MeshtasticMessages, MessageSharing` | `MeshtasticMessages, Shared/MessageSharing` |
| `WidgetsExtension` | `Shared, Widgets` | unchanged |
| `Meshtastic` | 20 groups incl. `Shared` | unchanged |

Same folder on disk, different navigator group. Nothing else moves.

One line of `project.yml` for a check that currently fails half the time repo-wide. Worth landing ahead of anything else queued.

